PR #1, opened 2010-09-01
Synapse bugfix/3215 inflector count params

zeelot:
fixes http://dev.kohanaframework.org/issues/3215 and tests pull requests 2.0 =P

### DIFF
--- a/classes/kohana/inflector.php
+++ b/classes/kohana/inflector.php
@@ -69,18 +69,15 @@ class Kohana_Inflector {
 	 */
 	public static function singular($str, $count = NULL)
 	{
+		// $count should always be a float
+		$count = ($count === NULL) ? 1.0 : (float) $count;
+
+		// Do nothing when $count is not 1
+		if ($count != 1)
+			return $str;
+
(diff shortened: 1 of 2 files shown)


kiall:
Merged

shadowhand:
This does more than just change to floats, the logic/flow changes need to be removed.

zeelot:
I thought I simplified it a bit, the reason I moved the 'remove garbage' section down was because the one above it says 'do nothing' so it would have been inaccurate

shadowhand:
I guess it should say "do nothing else" then.  ;) Logic should follow through. Test: `Hamster => hamsters` and you'll see.

zeelot:
well actually, why are we running strtolower on it? how is that part of the process involved in pluralizing a string?

shadowhand:
Pretty hard to use array features when you don't normalize what you are searching for.

shadowhand:
Uh, kiall, I explicitly said that this needs to be fixed before it can be merged. Please revert your merge.

PR #2, opened 2010-09-05
Patch for issue # 3226

joelpittet:
words with ending "us" are singular this patch keeps them that way.


BRMatt:
Just gonna close this as it says it wants to merge into master

PR #6, opened 2010-09-15
Validation Luhn Test and Fix

joelpittet:
For issue #3200


joelpittet:
Consider this not an API change since luhn method never existed before.

shadowhand:
Fixed by e399e41461714b60c0b122d5701dc22de8ea7b57, thanks for the patch!

PR #7, opened 2010-09-16
Issue/3239

cs278:
Pull request as requested for http://dev.kohanaframework.org/issues/3239


cs278:
I've updated the fix and added a unit test, per discussion on the tracker.

shadowhand:
Merged newer pull request, this is invalid.

PR #15, opened 2010-09-22
3.1.x

jerfowler:
New Text static method ordinal. Gives the ordinal value of an integer: 1st, 2nd, 3rd, 11th...
 Text::ordinal($num);


jerfowler:
Whoops, didn't see this is already done in Kohana_Num...

PR #20, opened 2010-10-01
#3300 - Cache Directory must be writable even with Caching disabled

mintbridge:


shadowhand:
See comments at http://dev.kohanaframework.org/issues/3300

PR #21, opened 2010-10-06
Translation fix

chrisgedrim:
Within the Validate->errors() function labels and values were not being correctly translated if a language was passed as $translate. (See http://dev.kohanaframework.org/issues/3311)

The __() function within base.php would never translate to the language passed as $lang. (See http://dev.kohanaframework.org/issues/3312)


shadowhand:
Fixed by 28cb13a1a6b260d121754df7d84aa431798e3e8e

PR #24, opened 2010-10-10
Route test

bluehawk:
Add url testing to routes, via Route::test so you can test what route a url or a set of urls is matching, as well as compare it to what you expect it to match.

(I seem to have somehow gotten an earlier commit dragged into this branch, you might have to cherry pick 0385dfe4.)


BRMatt:
I'd [personally] prefer to add Route testing as a testcase in the unittest module so that they can be tested as part of a build process.

kiall:
Looks like these got left behind! Was there a ticket opened for discussion? If not .. can you open one!

PR #25, opened 2010-10-10
Kohana info

bluehawk:
Added Kohana::info().  It's like php_info, but with 100% more Kohana!


kiall:
Looks like these got left behind! Was there a ticket opened for discussion? If not .. can you open one!

PR #26, opened 2010-10-17
Resumable Downloads

ThePixelDeveloper:
Allows people to pause a download and restart it. Can also download parts of a file in parallel.

Ticket: http://dev.kohanaframework.org/issues/1882


ThePixelDeveloper:
You should use the newer request here: http://github.com/kohana/core/pull/27

I fixed some formatting and removed a debugging call.

PR #27, opened 2010-10-17
Feature/1882 resumable downloads

ThePixelDeveloper:
A better pull request.


kiall:
Works as advertised - shadowhand's fseek() comment was moved lower down the file to cater for determining the starting position

PR #28, opened 2010-10-18
3.0.x

jimktrains:
http://dev.kohanaframework.org/issues/3336

I added support for better hashing (using the crypt function w/ a block cipher), and improved the encryption class to serialize the data so that all pieces of information needed to decrypt the data (save the key itself;)) (i.e.: cipher, mode, iv for metadata, encrypted metadata containing the length, checksum/hash, and the iv for the data). This allows the data to be decrypted even when the defaults in the config changes.

In the Text class I changed the random string code to be a little more robust (choosing letters and numbers to ensure that alnum has one of each from an array instead of using char values (e.g.: A->65).


shadowhand:
Duplicate.

PR #29, opened 2010-10-18
3.0.x

jimktrains:
I added support for better hashing (using the crypt function w/ a block cipher), and improved the encryption class to serialize the data so that all pieces of information needed to decrypt the data (save the key itself;)) (i.e.: cipher, mode, iv for metadata, encrypted metadata containing the length, checksum/hash, and the iv for the data). This allows the data to be decrypted even when the defaults in the config changes.

In the Text class I changed the random string code to be a little more robust (choosing letters and numbers to ensure that alnum has one of each from an array instead of using char values (e.g.: A->65).


jimktrains:
jeremyf76:

> These are all missing DocBlocks....

Yeah, I should have updated them.

> This will build a rather large multidimensional array each time random() is called... Better to store as a static var if you do it this way. Not sure I understand the benefits either as the old way worked just fine...

I didn't like how, when checking if the string was all num or alpha if it was of type alnum, it would just pick ints and then convert them to chars.  You're right, though, that that should be static.

> Not a fan of this... Should keep with standard convention and use the static $default var...

I'd appreciate antecedents:) Do you not like any of my code  or just that I broke convention.  The default can be changed to the $default var if that would make everyone happy.

shadowhand:
Please merge your commits down to 2 or 3 using `git rebase` and [create a ticket](http://dev.kohanaframework.org/projects/kohana3) that explains the benefits of what you are trying to implement here. Per-user configuration is already possible (manually) and I think putting that much information into the output string is bad for security.

PR #31, opened 2010-10-23
3.0.x: fall-through routes

tylerl:
# Fall-Through Routes

**Summary:** _Optionally allows the request engine to detect whether or not a controller and action exist before committing to a specific route choice. This in turn allows users to continue to employ "catch-all" routes while still easily detecting invalid URLs and routing invalid requests to an appropriate error page._

This patch adds a potential solution to the problem people are experiencing getting proper 404-handling working. Because proper 404 handling is so difficult to set up correctly, many sites do not implement it properly and instead end up showing a 500 error when a 404 is appropriate. Such a site is less user-friendly and less search-engine-friendly than it could have otherwise been.

**Case in point:  http://kohanaframework.org/foo**

The primary problem relates to catch-all "default" routing rules. These allow the controller and action to be specified dynamically with no further route customization; but unfortunately such a routing rule greedily accepts many URL patterns that it cannot process. These invalid URLs show up as 500 errors with a stack trace, as seen in the **kahonaframework.org** link above.
## How this problem is solved

The solution is simple: Routes can now optionally be marked as "`fall_through`" in the bootstrap.php file. If a route is a "`fall_through`" route, then extra checking is done during the construction of the `Request` object. For such routes, the framework now verifies that the controller exists, and that it has a method matching the action that will be used to process the request. If that check fails, the framework skips that route and "falls through" to the next route in the list.

This allows for a route listing as follows. Note that this is better for a novice user than the [wingsc](http://github.com/shadowhand/wingsc/blob/master/application/bootstrap.php) solution in that the user is allowed to take advantage of dynamic catch-all routing, and the user isn't expected to write complex error handling logic. Furthermore, the _wingsc_ solution incorrectly treats **ALL** exceptions as 404 errors, not just those triggered by an invalid URL.

```
Route::set('default', '(<controller>(/<action>(/<id>)))')
    ->defaults(array(
        'controller' => 'welcome',
        'action'     => 'index',
    ))->fall_through();

Route::set('404', '<path>')
    ->defaults(array(
        'controller' => 'error',
        'action'     => '404',
    ));
```

A route is marked as being a "fall-through" route using the `fall_through()` member function as follows:

```
$route->fall_through(TRUE);
```

or simply

```
$route->fall_through(); // the default is TRUE 
```

I implemented this feature as a core modification rather than as a module because it changes the core behavior in such a way that would make maintaining separate module needlessly complex and error-prone. The feature causes no change in performance or behavior when not explicitly enabled. And the feature is useful in very nearly every Kohana installation and dramatically simplifies and improves the user's experience.

Contact me at devel@tlarson.com if you have any questions. I'm new to working on this project, and while I've tried to follow all the coding standards, if you see any spaces out of place or other minor errors, please just fix the small detail rather than rejecting the pull request.

Thanks!
-Tyler

### DIFF
--- a/classes/kohana/request.php
+++ b/classes/kohana/request.php
@@ -632,31 +632,67 @@ class Kohana_Request {
 		{
 			if ($params = $route->matches($uri))
 			{
+				// Save defaults as temporary (in case we fall-through)
+				$controller = $this->controller;
+				$directory = $this->directory;
+				$action = $this->action;
+
+				if (isset($params['directory']))
+				{
+					// Controllers are in a sub-directory
+					$directory = $params['directory'];
+				}
+
+				// Store the controller
+				$controller = $params['controller'];
+
+				if (isset($params['action']))
+				{
+					// Store the action
+					$action = $params['action'];
+				}
+				else
+				{
+					// Use the default action
+					$action = Route::$default_action;
+				}
+
+				if ($route->fall_through)
+				{
+					try
+					{
+						// Perform same substitution found in execute()
+						$prefix = 'controller_';
+						if ($directory) 
+						{
+							$prefix .= str_replace(array('\\', '/'), '_', trim($directory, '/')).'_';
+						}
+						$class = new ReflectionClass($prefix.$controller);
+
+
+						if ( ! $class->getMethod('action_'.$action)) {
+							// Action not found -- on to next route
+							continue;
+						}
+					} 
+					catch (ReflectionException $e)
+					{
+						// failed to load controller class -- skip to next route
+						continue;
+					}
+					// Controller and method exist: continue with processing
+				}
+				
+				// Commit
+				$this->controller = $controller;
+				$this->directory = $directory;
+				$this->action = $action;
+
 				// Store the URI
 				$this->uri = $uri;
 
 				// Store the matching route
 				$this->route = $route;
-
-				if (isset($params['directory']))
-				{
-					// Controllers are in a sub-directory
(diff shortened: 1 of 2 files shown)


kiall:
Hiya - looks interesting :)

Can you post a feature request over @ http://dev.kohanaframework.org/projects/kohana3/issues ?

We keep discussion/issue tracking etc over there...

Yahasana:
why not class_exists and method_exists.

rename fall-through to route_recursive maybe more clear.

tylerl:
**@kiall**: Issue created. Thanks. http://dev.kohanaframework.org/issues/3346  

**@Yahasana**: I'm intentionally using the exact same logic and mechanism as the `execute()` function to eliminate any chance that one method will fail while the other succeeds. As for "route_recursive", I'm not dead-set on any given name, I just want the functionality to work. However, I don't think there's anything recursive about what's going on here -- the route doesn't call itself. The concept I'm trying to encapsulate with the name is _"if this route doesn't match, use the next one instead"_. If you can come up with a name that communicates that concept better, that's fine with me.

Also note that this property is set on the route itself, not globally enabled; some routes are fall-through while other routes are not.

tylerl:
Per Yahasana' suggestion, I added code to compare the required number of arguments for the action against the provided arguments and fail-through if there aren't enough.

shadowhand:
Please see 87378293df010a23f6cc21e5a4d084baf0752840 for my take on this request. I think blindly passing on to the next route isn't flexible or powerful enough.

PR #35, opened 2010-11-01
Feature/3363 num bytes

jeremeamia:
This fixes issue #3363 by adding a `Num::bytes()` function which converts numbers like `5K` to their byte values (5K = 5120). The first commit creates this functionality by copying code directly from the Upload::size() function. The second commit refactors this code to allow for unitless numbers, which are assumed to be in bytes already. This is important for numbers retreived using `ini_get()` which do use `K`, `M`, and `G` suffixes, but do not use `B`.

### DIFF
--- a/classes/kohana/num.php
+++ b/classes/kohana/num.php
@@ -17,6 +17,46 @@ class Kohana_Num {
 	const ROUND_HALF_ODD	= 4;
 
 	/**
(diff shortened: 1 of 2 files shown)


jeremeamia:
OK, I added support for decimals and a variety of byte units (M, Mi, MB, MiB, etc.). I think `Num::bytes()` is a pretty good addition to the `Num` class team.

PR #37, opened 2010-11-04
Tests for Issues/3361

ThePixelDeveloper:
See http://dev.kohanaframework.org/issues/3361


ThePixelDeveloper:
A better patch without all the merge commits:

https://github.com/kohana/core/pull/38

PR #40, opened 2010-11-09
3388 patch

joelpittet:
http://dev.kohanaframework.org/issues/3388

Fix for :
Inflector::plural changing case of input

example:
Inflector::plural('Cat', 1)
returns Cat

Inflector::plural('Cat', 2)
returns cats


shadowhand:
Merged in https://github.com/kohana/core/commit/c8f994c9bb87ffa5feef1e8c29e5b200016c1452

PR #41, opened 2010-11-17
Synapse feature/validation cleanup

zeelot:
This takes care of quite a few things.
1. Filters are no longer in Validate http://dev.kohanaframework.org/issues/2705
2. Rules are now defined using standard PHP callback syntax
   - array('class', 'method')
   - 'class::method'
   - 'method' (taken from validate class)
   - 'function'
   - function(){} (PHP 5.3+)
3. You can bind variables to be replaced in the parameter definition http://dev.kohanaframework.org/issues/2966
4. Callbacks are no longer needed http://dev.kohanaframework.org/issues/3360
5. Removed the $allow_empty parameter from check() http://dev.kohanaframework.org/issues/3366


zeelot:
P.S. I removed some of the old unit tests related to these changes but have not written new ones yet... I am having some difficulties with unit tests on the 3.1.x branch but I will give this a shot soon!

zeelot:
This pull request _should_ also close the following issue: http://dev.kohanaframework.org/issues/3355

jonathangeiger:
This looks even better than my stuff. I'm looking forward to seeing this in 3.1.

zeelot:
Few more issues done (hopefully)
- http://dev.kohanaframework.org/issues/3402 / http://dev.kohanaframework.org/issues/3158
- http://dev.kohanaframework.org/issues/3359

zeelot:
I think I am done with this pull request... tests seem to pass and I added one for bind(). Let me know if I am missing something :)

Quick note: I added a default for rules like so:

```
// These two are equivalent
$v->rule($field, $callback)
$v->rule($field, $callback, array(':value'))
// This is how you add a rule with NO parameters
$v->rule($field, $callback, array())
```

Isaiah suggested this because it avoids having to type array(':value') so many times.

PR #44, opened 2010-11-20
3.0.x

LegendSzymon-P:
redirect without index.php/

### DIFF
--- a/classes/kohana/url.php
+++ b/classes/kohana/url.php
@@ -101,7 +101,7 @@ class Kohana_URL {
 		}
 
 		// Concat the URL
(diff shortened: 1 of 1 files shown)


shadowhand:
Mm, why?

szymon-przybyl:
hm, becouse for example www.service.com/home/contact looks better than www.service.com/index.php/home/contact

shadowhand:
Yes, but that is a configuration issue. If you set `index_file` to `FALSE` in bootstrap, then "index.php" will not be in the URL.

szymon-przybyl:
oh, i didn't know, i was trying to set there '', but it doesn't worked, thanks :)

PR #45, opened 2010-11-24
3.0.x

ellisio:
Modified documentation for core to better support IDEs with support for auto-complete. Also altered some files that were doing assignments in conditions to be something like if(($var = something) != FALSE){}; basically just to make IDEs not bitch and moan.

### DIFF
--- a/classes/kohana/arr.php
+++ b/classes/kohana/arr.php
@@ -161,7 +161,7 @@ class Kohana_Arr {
 				$values = array();
 				foreach ($array as $arr)
 				{
-					if ($value = Arr::path($arr, implode('.', $keys)))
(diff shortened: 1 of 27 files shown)


zombor:
Make sure you attach this to the proper issue on redmine so it's properly tracked.

ellisio:
Pardon my ignorance, but how do I do that?

zombor:
Link to this pull request here: http://dev.kohanaframework.org/issues/2893

Also, this won't be merged in, there's more than just phpdoc refactoring.

I'm going to close this, please send another one when there's only phpdoc changes.

ellisio:
Roger that, will send another request shortly once I've rolled out the non-phpdoc changes.

PR #46, opened 2010-11-24
3.0.x

ellisio:
Modified documentation for core to better support IDEs with support for auto-complete.

### DIFF
--- a/classes/kohana/config.php
+++ b/classes/kohana/config.php
@@ -11,7 +11,10 @@
  */
 class Kohana_Config {
 
-	// Singleton static instance
+	/**
+	 * @access  protected
+	 * @var     Kohana_Config  Singleton static instance
+	 */
 	protected static $_instance;
 
 	/**
@@ -32,7 +35,10 @@ class Kohana_Config {
 		return self::$_instance;
 	}
 
-	// Configuration readers
+	/**
+	 * @access  protected
+	 * @var     array      Configuration readers
+	 */
 	protected $_readers = array();
 
 	/**
--- a/classes/kohana/config.php
+++ b/classes/kohana/config.php
@@ -12,8 +12,7 @@
 class Kohana_Config {
 
 	/**
-	 * @access  protected
-	 * @var     Kohana_Config  Singleton static instance
+	 * @var  Kohana_Config  Singleton static instance
 	 */
 	protected static $_instance;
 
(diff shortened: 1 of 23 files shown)


ellisio:
I viewed the ticket on Redmine, and there isn't an option for me to associate a revision to the ticket #. Can you please do this for me?

Regards,
Andrew

ThePixelDeveloper:
Just add a comment with a link to this pull request, that's all they're asking for.

ellisio:
Done and done. :)

rafi:
Oh man, this is great! thanks for doing it

pocesar:
awesome, I was about to do the same lol

seeYouInHell:
@access property is useless for PHP5.

ellisio:
Out curiosity why is that useless when it's meant purely for documentation purposes?

seeYouInHell:
This property was introduced because of PHP4 hasn't public, protected and private keywords. Now we can see visibility modifier without @access property. So this property is inconsistent with PHP5 code.

ellisio:
Do PHPDOC generators read the code's access key though? If not, I would assume @access is still useful for doc generation?

pocesar:
yes, it's still useful for documentation (aside from the code completion on IDEs)

shadowhand:
Please remove `@access`, it is completely useless in PHP5 code.

shadowhand:
And please follow our conventions for `@var` comments:

```
@var  type  comment
```

I see several places where you are using:

```
comment
@var  type
```

Which does not fit our coding style.

mikelbring:
`@access` is not pointless. As mentioned before its for PHPDOC and IDEs.

ellisio:
I'll revise the comments, but I disagree about the `@access` key. This is purely documentation, and that key is used in PHPDoc generators to document what kind of access the method has as they can not read the actual access keys on the code.

But if those need to be removed for this change to go in, I'll make that change, but I completely disagree with this.

shadowhand:
It doesn't matter if you disagree, phpdoc only defines `@access` before of PHP4. If your IDE is too stupid to understand the PHP, too bad.

ellisio:
This is a community driven framework, there is no reason you should be taking feedback from the community and rebuttal with attitude. If you don't want community feedback on general day-to-day user needs, then make this a closed-source framework and charge for it.

Anyways, I'm making the changes now, I'll close this and send another pull request shortly.

shadowhand:
We will still accept your patch, we just don't want to encourage the use of legacy features.

shadowhand:
Merged in c7432cdaeccd47b58d55bb61e89c281738bc8096

PR #51, opened 2010-12-02
Feature/3364 post max size exceeded

jeremeamia:
This adds the `post_max_size_exceeded` method to the `Request` class as requested in http://dev.kohanaframework.org/issues/3364


BRMatt:
Used in 3a478454af74366d9238a27997f62423955caba2

PR #58, opened 2010-12-21
Assert type

le6o:
assertType() has been deprecated, use assertInternalType() or assertInstanceOf() instead.

See: https://github.com/sebastianbergmann/phpunit/issues/issue/77


isaiahdw:
I'm not sure if we want to apply this change because it would break compatibility with phpunit 3.4.x. Can you please create an issue on the Kohana issue tracker (http://dev.kohanaframework.org/projects/kohana3/issues) so we can keep track of this request and add it when we decide to stop supporting phpunit 3.4? Thanks!

PR #63, opened 2011-01-07
Bugfix/validation empty rules

zeelot:
http://dev.kohanaframework.org/issues/3552 - Implemented the $run_if_empty parameter to rule()
http://dev.kohanaframework.org/issues/3553 - Implemented Validation::not_empty()

I still need to write unit tests but I have to get some sleep so I will write them tomorrow night if no one else has already.

I really quite like this implementation after trying it because you can do this for example:

```
$validation
    // This field cannot be empty!
    ->rule('field1', 'min_length' array(':value', 5), TRUE)
    // This field can be empty!
    ->rule('field2', 'min_length' array(':value', 5));
```

This feels like another step towards a very powerful, but simple, validation library.

### DIFF
--- a/classes/kohana/validation.php
+++ b/classes/kohana/validation.php
@@ -192,6 +192,25 @@ class Kohana_Validation extends ArrayObject {
 	}
 
 	/**
+	 * Helper method to quickly add 'not_empty' rules to an array of fields
+	 *
+	 *     // This provides a much shorter syntax for not_empty rules
+	 *     $validation->not_empty(array('first_name', 'last_name'));
+	 *
+	 * @param   array  array of fields to apply the 'not_empty' rule to
+	 * @return  $this
+	 */
+	public function not_empty(array $fields)
+	{
+		foreach ($fields as $field)
+		{
+			$this->rule($field, 'not_empty', NULL, TRUE);
+		}
+
+		return $this;
+	}
+
+	/**
 	 * Executes all validation rules. This should
 	 * typically be called within an if/else block.
 	 *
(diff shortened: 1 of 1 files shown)


kiall:
Zeelot - Whats the status of this pull?

zeelot:
This is still my preferred fix, but zombor feels we can find a better way to do this (without adding a third parameter to `rule()`.

We should leave this pull request open until the related issues are resolved (now for 3.3.0)

kiall:
@Zeelot ping.. Its 3.3 time!

zeelot:
@kiall I still believe this is the best solution, granted we have not discussed this further... I feel like this is much better than our current implementation and it should be pulled in for 3.3.0. If you guys agree (mainly @zombor as he was the one disagreeing 6 months ago) I can easily fix this pull request and merge it in.

If we are still disagreeing, we should bump to 3.4.x and continue discussing ideas :( I don't think we should leave the empty rules functionality as is.

shadowhand:
Until someone suggests a better syntax, this seems like a good solution. In this particular case I don't see a more obvious solution than another parameter.

kiall:
@Zeelot I actually don't have an opinion either way really.. I was just going through the pull requests and wanted to remind you to get this (or something better?) merged or closed :)

The only thing I remember from 6 months back re this was my suggestion that no rules at all run on empty values bar the "not_empty" rule.. That way you just have to be explicit about if you want an empty value or not...

But I seem to remember this suggestion being shot down for some reason months ago :)

zeelot:
No one seems to have suggested anything better than this but it's too late for 3.3.x so I will simply be making this change in 3.4.x. Still leaving this open until I get to it.

PR #64, opened 2011-01-11
Core debug

misa:
Debug methods were moved to separate class refs #2682


isaiahdw:
Fixed by f6a1dea45e8783351c95019a6fba1e2c18a55fd1

Thanks!

PR #67, opened 2011-01-14
Feature/3536 args

tylerl:
Assign arguments by name for controller actions. This works by using reflection to determine the appropriate order of the arguments, and then creating an array that is ordered appropriate for the function call.

This way we don't have to resort to silliness like `$this->request->param('foo')` when simply `$foo` will suffice.

### DIFF
--- a/classes/kohana/core.php
+++ b/classes/kohana/core.php
@@ -1061,17 +1061,25 @@ class Kohana_Core {
 			} 
 			else
 			{
-				if ($null_pad) 
+				if ($func_param->isDefaultValueAvailable())
 				{
-					// Note that null-padding the array isn't, strictly speaking,
-					// "correct". However, if you *don't* null pad, then parameters
-					// after the missing one can't be assigned. Either way, we're
-					// dealing with undefined behavior, so we have to pick a compromise.
-					$args[] = NULL;
-				} 
+					// use the default value provided in the function signature if it exists
+					$args[] = $func_param->getDefaultValue();
+				}
 				else
 				{
-					break;
+					if ($null_pad)
+					{
+						// Note that null-padding the array isn't, strictly speaking,
+						// "correct". However, if you *don't* null pad, then parameters
+						// after the missing one can't be assigned. Either way, we're
+						// dealing with undefined behavior, so we have to pick a compromise.
+						$args[] = NULL;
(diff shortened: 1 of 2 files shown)


tylerl:
There's already plenty of reflection going on for every request. 

This code simply adds a single call to `$method->getParameters()`, and then iterates once through the list. It's hardly a performance-killer. Maybe a few fractions of a millisecond per request at the worst. Certainly nothing detectable on any scale.

It's worth pointing out that the overhead of the alternative -- `$this->request->param('name')` is not significantly lower, and with a lot of parameters, may even be higher as you're doing a lot of unnecessary lookups. You'd have to benchmark it to find out which is faster, but it's not a slam dunk either way.

Point being: arguing over microseconds when the readability and ease-of-use of the code is at stake is unproductive.

shadowhand:
I really like this idea, but I think the implementation could use some work. It would likely be better to have a private method (maybe `Request::_remap` or so) that would take the action (`ReflectionMethod`) and args array and reorder the args based on the action parameters. Having a `Kohana` method for this really doesn't make any sense and I still want to end up calling `invokeArgs`.

tylerl:
@shadowhand: I wanted the mechanism to be reusable, but I wasn't quite sure where to put it. It seemed to be general-purpose, which is why it ended up in `Kohana::`, but it could go anywhere. It could just as easily be a function in `Request::`.

Also worth considering is using the same mechanism on `before()` and `after()`. It only passes the variables that the function signature requests, so unless you ask for a parameter, it won't be used (i.e. backwards compatible). So, for example, you could say `before($action)` or `after($id)` or something. It seems like it could prove useful if you actually have a use for the before and after functions.

I left those two as there are, though, to avoid rocking the boat too much.

kiall:
Anyway - lets move this conversation back over to the ticket .. Redmine wont let me add comments from my phone hence I commented here! http://dev.kohanaframework.org/issues/3536 ..

tylerl:
This patch is now bugfix-only. All original behavior is preserved except in the case where kohana was incorrectly ordering the parameters, in which case it will now behave correctly. No other behavior is changed.

kiall:
Closed per http://dev.kohanaframework.org/issues/3536

PR #69, opened 2011-01-25
3.1/release/3.1.0

nanodocumet:
First time doing this. My first 2 cents back to Kohana.

Added the changes in URL class when upgrading from 3.0 to 3.1, although not sure if I grab the correct branch.

Thanks,

Nano.


isaiahdw:
It looks like something went wrong with this pull request. Can you please resubmit it? I think you might have merged branches wrong so git is showing a lot more changes or something. I can't tell what you actually changed!

Also please open an issue on the issue tracker for all changes.

PR #70, opened 2011-01-26
3.1/develop Just for consistency, updated text to match Kohana::FILE_SECURITY

nickolasgregory:
A few files had "No direct access allowed."
Changed to "No direct script access." as per Kohana::FILE_SECURITY


shadowhand:
Applied to 3.0 in e051609a8144709d3a48c2b4cd1adbc0723c8595, and 3.1 in 3fd4ca861977081fc39678f4eb0e047628e75c8a

PR #71, opened 2011-01-26
Updated version number

simshaun:
This is the 3.1.0 (RC2) release, correct? If so, shouldnt the core version constant reflect that?

**If so,**
The stable 3.0.9 release available from http://kohanaframework.org/ also has this constant wrong. It's set to 3.0.8 in that package.

### DIFF
--- a/classes/kohana/core.php
+++ b/classes/kohana/core.php
@@ -16,7 +16,7 @@
 class Kohana_Core {
 
 	// Release version and codename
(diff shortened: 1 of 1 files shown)


misa:
No, this is "3.0/develop" branch, so it is correct as it is.

simshaun:
Wooops. Guess I should pay attention what branch I'm pulling. ha

PR #73, opened 2011-01-29
3.1/bugfix/3658 validation class method as error name

zeelot:
Sorry for the long delay =( here is the fix with a unittest. The only rules that cannot simply return a boolean value to add errors now are lambda functions because they have no 'name'. A lambda function will have to add the errors manually to the validation object.

http://dev.kohanaframework.org/issues/3658


isaiahdw:
Thanks!

PR #77, opened 2011-02-10
Upgrade guide: Adding Log const changes

sfgeorge:
3 constants for log message types have moved from Kohana_Core to Kohana_Log. I've updated upgrading.md in the userguide to mention these changes.

Also, I filed the bug report at http://dev.kohanaframework.org/issues/3723

Thanks,

Stephen

### DIFF
--- a/classes/kohana/http/header/value.php
+++ b/classes/kohana/http/header/value.php
@@ -129,7 +129,7 @@ class Kohana_Http_Header_Value {
 		// Unrecognised value type
 		else
 		{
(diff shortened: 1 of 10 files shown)


zombor:
The destination for this should be 3.1/develop. That will remove all those extra commits.

sfgeorge:
I botched this by tracking the base branch as 3.1/master instead of 3.1/develop.  I've things right in pull request 78: https://github.com/kohana/core/pull/78.

Closing request 77.

PR #82, opened 2011-02-24
3.1/develop

acoulton:
Clarifying in the migration guide that the new Request::controller(), action(), etc methods are instance methods not statics per [this forum post](http://forum.kohanaframework.org/discussion/comment/55215/#Comment_55215)

### DIFF
--- a/guide/kohana/upgrading.md
+++ b/guide/kohana/upgrading.md
@@ -12,10 +12,10 @@ This has changed to:
 
 Some properties that existed in the request class have been converted into methods:
 
-	- Request::$controller -> Request::controller()
-	- Request::$action -> Request::action()
-	- Request::$directory -> Request::directory()
-	- Request::$uri -> Request::uri()
+	- $this->request->controller -> $this->request->controller() //Or Request::current()->controller();
+	- $this->request->action -> $this->request->action()
+	- $this->request->directory -> $this->request->directory()
+	- $this->request->uri -> $this->request->uri()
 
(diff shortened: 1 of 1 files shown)


kiall:
This isn't right .. sorry :)

`$this->request` is one place you may find the request class .. not all of them. 

`Request::instance()->controller` would be the only 100% correct way to reference it.

(edited because of a mistake..)

isaiahdw:
The original examples are correct. This doesn't need changing.

acoulton:
@isaiahdw the examples are correct but not as clear as they could be.

The samples appear in code formatting - Class::method convention is generally only used in text, with formatted code differentiating between static and instance methods.

@kiall - agreed $this->request is only one possibility, but in context on that page it comes immediately following use of $this->request to illustrate request/response change. Felt that the comment on line one of my change to show an alternative was sufficient, and that should presume reader knows how to get a Request instance.

Think this change would aid clarity with no downside and worth a second pitch. YMMV. If closed again will leave it so.

isaiahdw:
@Andrew, Ah I didn't notice it was inside a code block. I'm fine with this change. I agree that changing it to $this->request would be more consistent with the examples above.

isaiahdw:
Can you please open an issue for this (if you haven't so we don't forget about this)?

acoulton:
@isaiahdw thanks - [ticket 3806](http://dev.kohanaframework.org/issues/3806)

PR #83, opened 2011-02-24
3.1/docs/validation

zeelot:
This should be a good start for the Validation documentation in 3.1.x. There seems to be some redundant information at the top of the page but all the information should be accurate now. I added some explanation for variable binding, the two ways of running rules, and the ways errors are added. I also added explanations for the Valid class and how validation checks for rules there first.

There are a few minor things missing, like some explanation of how the errors() method works and how the messages are looked up/translated. Should also include some better details of how to add errors (passing of error parameters)

CHEERS!


zombor:
Merged.

PR #85, opened 2011-03-03
Doc: Updates for the coding guideline

goyote:
http://dev.kohanaframework.org/issues/3816

### DIFF
--- a/guide/kohana/conventions.md
+++ b/guide/kohana/conventions.md
@@ -29,20 +29,133 @@ Form                  | classes/form.php
 In order to produce highly consistent source code, we ask that everyone follow the coding standards as closely as possible.
 
 ### Brackets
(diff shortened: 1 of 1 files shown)


shadowhand:
Actually `class Foo {  }` is the correct option; 2 spaces inside empty brackets.

goyote:
Can we change that?

1) Empty brackets are currently being used everywhere in the source code, e.g `class Arr extends Kohana_Arr {}` (with the exception of a couple of places.)

2) Empty brakets are more intuitive, since we already remove whitespace from similar structures, e.g. 
`$arr[] = $item;`, `$empty = array();`

shadowhand:
It used to be that empty curlies would have two spaces between them, but I am fine with being more consistent.

goyote:
yay :)

shadowhand:
b705e0b7b7f1836c6bcd3b464adedcc1f8ccb942

PR #90, opened 2011-03-16
3.1/master

tonyhb:
Fixing http://dev.kohanaframework.org/issues/3857 (currently invalid, but I'm not sure I explained it well enough)

### DIFF
--- a/classes/kohana/validation.php
+++ b/classes/kohana/validation.php
@@ -31,7 +31,7 @@ class Kohana_Validation extends ArrayObject {
 	protected $_labels = array();
 
 	// Rules that are executed even when the value is empty
-	protected $_empty_rules = array('not_empty', 'matches');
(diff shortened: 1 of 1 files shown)


kiall:
'min_length' cannot be included in $_empty_rules .. 

If it is, then optional fields that have a min length are impossible.

The addition of 'valid::not_empty' and 'valid::matches' to $_empty_rules is probably valid though, even though there not intended to be used that way..

tonyhb:
Didn't think of that, sorry.

kiall:
One other nitpick ;) Send pull requests to 3.1/develop rather than 3.1/master .. the master branches == the latest release .. so no changes go there directly..

tonyhb:
Yezzir =)

Will take out the `min_length` and sort all of that out mate.

> ---
> 
>   kiall 
> mailto:reply@reply.github.com
> 16 March 2011 15:46
> 
> One other nitpick ;) Send pull requests to 3.1/develop rather than 
> 3.1/master .. the master branches == the latest release .. so no 
> changes go there directly..

PR #92, opened 2011-03-16
3.1/develop

tonyhb:
Teeny. http://dev.kohanaframework.org/issues/3857

### DIFF
--- a/classes/kohana/validation.php
+++ b/classes/kohana/validation.php
@@ -31,7 +31,7 @@ class Kohana_Validation extends ArrayObject {
 	protected $_labels = array();
 
 	// Rules that are executed even when the value is empty
-	protected $_empty_rules = array('not_empty', 'matches');
+	protected $_empty_rules = array('not_empty', 'matches', 'valid::not_empty', 'valid::matches');
 
 	// Error list, field => rule
 	protected $_errors = array();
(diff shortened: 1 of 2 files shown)


shadowhand:
Needs a unit test.

tonyhb:
Reinstalled mac a couple of days ago, phpunit is being temperamental. Will sort it out this weekend.

tonyhb:
Got around to adding the unittest for you. In retrospect, Valid classes aren't meant to be called this way and adding the if statement feels horrible. Don't know if it's worth it?

kiall:

!build

kiall:
Looks like this is an actual build failure, rather than a result of the builder bug.

<pre>
06:12:02  1) Kohana_ValidationTest::test_check with data set #1 (array(''), array(array(array('Valid::not_empty'))), array(), false, array('foo must not be empty'))
06:12:02  Undefined offset: 1
</pre>

tonyhb:
Yeah no problem Kiall. Haven't looked at this code in a while, I'll check it out and see what's going on this evening mate.

tonyhb:
Kiall, I deleted my branch a couple of weeks back, just going through the old code from the diff though. The build failure's from this pull request, but I'm not sure why. The diff shows that this should be in the provider:

```
array(
    array('foo' => ''),
    // Test static Valid calls
    array('foo' => array(array('Valid::not_empty'))),
    array(),
    FALSE,
    array(
        'foo' => 'foo must not be empty',
    ),
),
```

But the data set that threw an error was this:

```
(array(''), array(array(array('Valid::not_empty'))), array(), false, array('foo must not be empty'))
```

Seems to be missing some array keys, somehow. 

The original ticket is marked as invalid on the tracker, and it was only to include not_empty checks with Valid:: static classes so it's no big deal.

Is there anything you need me to do for this, or shall we just leave it?

kiall:
We just implemented automated building of Pull Requests, I haven't actually looked at any of the pull's just checking the builder worked correctly.

 IMO, Using the full qualified name for not_empty should be valid. I'll re-open the ticket.

PR #95, opened 2011-03-23
3.1/master

sublimeowl:
Added a small condition to accept also uris like javascript:void(0);


shadowhand:
Inline Javascript is bad, period, end of story. We will not be accepting this.

PR #97, opened 2011-03-27
3.1/master

kemo:
It really doesn't make any sense to promote calling the parent's before() action _after_ the child's.

Take an ACL which initiates in Controller_Application for example:

if it doesn't do the job that's supposed to be done on the application level, child controllers extending the Controller_Application will be potentionally vulnerable, at least the stuff being executed in the before() method.

### DIFF
--- a/classes/kohana/controller/rest.php
+++ b/classes/kohana/controller/rest.php
@@ -53,6 +53,8 @@ abstract class Kohana_Controller_REST extends Controller {
 	 */
 	public function before()
 	{
+		parent::before();
+		
 		$this->_action_requested = $this->request->action();
 
 		$method = Arr::get($_SERVER, 'HTTP_X_HTTP_METHOD_OVERRIDE', $this->request->method());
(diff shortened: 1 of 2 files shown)


shadowhand:
This is entirely subjective and depends hugely on what your application does.

kemo:
@shadowhand maybe does seem subjective but in case of Kohana_Controller_Template it totally makes sense, I've already had 3 'students' confused on this one.

kiall:
I have to agree that for most use cases, it makes perfect sense to call `parent::before()` before your before code, and `parent::after()` after your after code..

Sure there will be some cases where that's not what you want - but when it's not one of those cases, i think @kemo's suggestion should be followed.

In any case - this can't be merged into 3.1/master (or 3.1/develop either ..).. Will have to be done in 3.2/develop if we change it ..

kiall:
Controller_REST has been removed in 3.2. Closing.

PR #98, opened 2011-03-29
Fixed a few bugs in the tutorial custom error pages for Kohana 3.1:

ddrake:
References: Bug Report #3894
1. The original code generates a url with the base_url prepended.
   For apps with non-empty base_url, that produces the error:
   "Unable to find a route to match e.g. 'kohana_test/error/404/message...'
2. The controller name should be consistent in the route.
3. Request::$instance doesn't exist anymore and uri is a method now.

### DIFF
--- a/guide/kohana/tutorials/error-pages.md
+++ b/guide/kohana/tutorials/error-pages.md
@@ -38,7 +38,7 @@ Our custom exception handler is self explanatory.
 					}
 
 					// Error sub-request.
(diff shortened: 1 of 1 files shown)


ddrake:
Redmine url: http://dev.kohanaframework.org/issues/3894

ddrake:
Isaiah's fixes in 3.1.3.1 are basically equivalent.  Closing this...

PR #103, opened 2011-04-07
Added support for setting a recursion limit in Debug::dump parameters

timborden:
When dumping the response from the FedEx API, the Debug::dump didn't display the full set of data.  I've added support for specifying the amount of recursion when dumping.  The default value remains 10.

### DIFF
--- a/classes/kohana/debug.php
+++ b/classes/kohana/debug.php
@@ -45,11 +45,12 @@ class Kohana_Debug {
 	 *
 	 * @param   mixed    variable to dump
 	 * @param   integer  maximum length of strings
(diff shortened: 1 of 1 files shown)


shadowhand:
Did you create an issue on [dev.kohanaframework.org](http://dev.kohanaframework.org/projects/kohana3/issues)? If not, please do so and add a comment here with the issue number.

timborden:
Sorry, didn't know I was suppose to....here it is:
http://dev.kohanaframework.org/issues/3909

shadowhand:
Merged in https://github.com/kohana/core/tree/feature/3909-debug-recursion-level

PR #106, opened 2011-04-26
Added default uri values and automatic retaining of uri values to Kohana_Route

ymtn:
A route can be declared like this, with an optional key of, for instance, 'language'. This key will be automatically injected into all subsequent calls to Route::uri

```
Route::set('default', '(/<language>)(/<controller>(/<action>(/<id>)))',
    array(
        'language' => '(nl|en|fr)'
    ))
    ->defaults(array(
        'controller' => 'foo',
        'action'     => 'bar'
    ))
    ->retain(array(
        'language'
    ));
```

When on the location www.mysite.com/fr, a call to

```
Route::get('default')->uri(array('controller' => 'foo','action' => 'bar'))
```

Will render /fr/foo/bar

The function Route::uri_defaults can be used to set these default uri values explicitly:

```
Route::get('default')->uri_defaults(array('language'=>'fr'));
```

Changes to Kohana_Route:

Added
Kohana_Route::_retain
Kohana_Route::retain()
Kohana_Route::_uri_defaults
Kohana_Route::uri_defaults()

Changed
Kohana_Route::uri()
Kohana_Route::matches()

### DIFF
--- a/classes/kohana/route.php
+++ b/classes/kohana/route.php
@@ -269,6 +269,11 @@ class Kohana_Route {
 	protected $_defaults = array('action' => 'index', 'host' => FALSE);
 
 	/**
+	 * @var  array
+	 */
+	protected $_retain=array();
+	
+	/**
 	 * @var  string
 	 */
 	protected $_route_regex;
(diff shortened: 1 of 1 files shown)


samsoir:
This would be an api change, so it cannot go into 3.1 - possibility for it to go into 3.2 if it has a feature request ticket associated with it.

ymtn:
I allready created an issue at http://dev.kohanaframework.org. for 3.1.3. 
Should I make another one for 3.2 ?

samsoir:
Just update the existing ticket for 3.1.3 to target 3.2 instead. It will not go into 3.1.3 as it is an API change, so this is being rejected.

ymtn:
Ok, 3.2 it is. 
It seems I can't update the ticket though; I can just add a comment.

BRMatt:
Just for reference, ticket is [3943](http://dev.kohanaframework.org/issues/3943)

PR #117, opened 2011-05-11
Validation example correction

ghost:
Changed Validate to Validation in the example in order to comply with 3.1 code


kiall:
This has already been fixed/merged in from somewhere else .. sorry it took so long to get to this pull!

PR #119, opened 2011-05-15
Added 'expense' to the irregular list

edy:


cbandy:
This has bugged me for a long time. Shouldn't words that end in "se" always be handled the same way? Why are they irregular?

edy:
they can't be handled all the same way:
mouse -> mice
goose -> geese
moose -> moose
...

i think i should close this pull request. i can add 'expense' to my own config file

kiall:
I've merged this in.

If there is a better default for the handling of -se words, great .. lets use it.. 

If not .. lets add exceptions as needed.

PR #120, opened 2011-05-16
#3996 - Native Session Library floods log

mabeyj:
This fixes the problem in 3.0.11 where "Error reading session data:" is added to the logs after every page request.

Based on commit 55fceefb to fix #3964.


kiall:
Merged

PR #121, opened 2011-05-22
#3436 - Text::auto_link_urls() breaks between tags

tasuki:
Hello good people, this is my first attempt at a pull request, so please be patient with me.

I'm trying to fix #3436 by stopping the url match on white space or the less-than sign. If there's a less-than sign in the passed url (shouldn't happen, I hope?), this commit will break it.

I'm not quite sure about the branching, but as target is 3.0.12, I'm submitting to 3.0/develop (that okay?).

In case this is accepted, it should probably be merged into both 3.1/develop and 3.2/develop, which suffer from the same.


ghost:
Looks great, thanks :-)

PR #122, opened 2011-05-23
#4013 - Arr::get should use array_key_exists instead of isset

Ikke:
This fixes #4013.


Ikke:
Commited against the wrong version, closing it.

PR #125, opened 2011-05-25
Added stack traces from xdebug to fatal errors.

leth:
I was annoyed that there were no stack traces for fatal errors, so I did some digging and found that xdebug will give you a stack trace after a fatal error

The corresponding issue is [#4020](http://dev.kohanaframework.org/issues/4020)


samsoir:
This cannot go into 3.1 as it represents an API change.

PR #126, opened 2011-05-25
Added stack traces to fatal errors

leth:
I was annoyed that there were no stack traces for fatal errors, so I did some digging and found that xdebug will give you a stack trace after a fatal error

The corresponding issue is [#4020](http://dev.kohanaframework.org/issues/4020)

### DIFF
--- a/classes/kohana/core.php
+++ b/classes/kohana/core.php
@@ -998,12 +998,20 @@ class Kohana_Core {
 				$trace = array_reverse(xdebug_get_function_stack());
 				array_shift($trace);
 
-				// xdebug doesn't currently set the call type key
-				// Open feature request: http://bugs.xdebug.org/view.php?id=695
 				foreach ($trace as & $frame)
 				{
+					// xdebug doesn't currently set the call type key
+					// Open feature request: http://bugs.xdebug.org/view.php?id=695
 					if ( ! isset($frame['type']))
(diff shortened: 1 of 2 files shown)


leth:
I've had another think about this, and [it could be implemented without API changes by subclassing ErrorException](http://dev.kohanaframework.org/issues/4020#note-6).
I'd appreciate your thoughts or advice on this :)

leth:
Oh ignore that. I just double-checked the signature and the getTrace method is final.

leth:
I've updated the commits, will @Kohana-Builds notice, since the rebase left the commit dates in the past?

kiall:
@Kohana-Builds is smarter than that ;) Basically, if commits are added/updated since the last time he commented.. 

Re the pull: the changes look good, but it was overlooked for the 3.2 window :( We'll be opening a 3.3 branch soon, I'll do my best to remember it!

leth:
Awesome, thanks. I'm looking forward to seeing what 3.3 is going to look like :)

kiall:
I've manually merged this into 3.3 as the exception handler has had lots of changes already :)

Thanks :)

https://github.com/kohana/core/commit/f86c9b877895e5d356284d197de94bc1581c670a

leth:
Cool! :)

PR #129, opened 2011-05-26
#4019: Injected routes are not taken into account for initial request

czukowski:
Redmine ticket: [#4019](http://dev.kohanaframework.org/issues/4019).

This should fix it.


kiall:

!build

PR #131, opened 2011-06-02
3.1/develop

ghost:
Trivial change as the validation page really needs a rewrite, but this small diff threw me off while learning validation. I'll see how I can contribute to rewriting this page.


kiall:
This has already been fixed/merged in from somewhere else .. sorry it took so long to get to this pull!

PR #136, opened 2011-06-23
3.2/feature/validation changes

zeelot:
- #3424 http://dev.kohanaframework.org/issues/3424
- #4062 http://dev.kohanaframework.org/issues/4062
- #4034 http://dev.kohanaframework.org/issues/4034
- #3552 http://dev.kohanaframework.org/issues/3552

As many commits as possible so we can still debate things if anyone complains, but I did confirm all the changes with @zombor before applying :)

### DIFF
--- a/classes/kohana/validation.php
+++ b/classes/kohana/validation.php
@@ -126,10 +126,22 @@ class Kohana_Validation implements ArrayAccess {
 
 	/**
 	 * Returns the array representation of the current object.
+	 * Deprecated in favor of [Validation::data]
 	 *
+	 * @deprecated
 	 * @return  array
 	 */
 	public function as_array()
+	{
+		return $this->_data;
+	}
+
+	/**
+	 * Returns the array representation of the current object.
+	 *
+	 * @return  array
+	 */
+	public function data()
 	{
 		return $this->_data;
 	}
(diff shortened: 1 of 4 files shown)


zeelot:
Ah, forgot to make ArrayAccess read-only and unit test the methods... I will take care of this tonight.

classes/kohana/validation.php
```diff
@@ -63,19 +119,31 @@ class Kohana_Validation extends ArrayObject {
...
 	 * @return  array
 	 */
 	public function as_array()
 	{
-		return $this->getArrayCopy();
+		return $this->_data;
+	}
+
+	/**
+	 * Returns the array representation of the current object.
```
cbandy: This is not clear enough for me. It should say something about the data/values.

zeelot: How about: `Returns the array of data to be validated.`

cbandy: I like it.

zeelot:
This should be everything again :)
- reverted back to `_empty_rules`
- reverted all the `Valid` methods to make more sense
- fixed an issue with `Valid::email_domain` running on empty values
- added tests for `Validation` `ArrayAccess` methods
- made `Validation` read-only (`ArrayAccess`)
- added tests for `Valid` methods to run on `NULL` and `FALSE` values

PR #142, opened 2011-07-14
Fixes #4112, fix __() to use default source language

ernestas:


kiall:

!build

shadowhand:
This pull request does not do enough. In every case where Kohana is currently calling `__(...)`, it needs to explicitly state `__(..., 'en-us')` or messages are no longer going to translate properly.

ernestas:
@shadowhand there are tickets for that. I'll change this pull to be against 3.3.

Tickets listed in first comment: http://dev.kohanaframework.org/issues/4145#note-1

ernestas:
https://github.com/kohana/core/pull/195

PR #143, opened 2011-07-17
Cleaning UTF8 class

damianstasik:
Syntax and code cleaning of UTF8 class.


zeelot:
Sorry this slipped so far… looks like you deleted your fork so I'm having someone create a new PR with this change. Thanks for submitting it in the first place.

P.S. will do my best not to let PRs go unanswered in the future.

PR #144, opened 2011-07-20
Feature/4125

le6o:
Changed Security::token() and Security::check() to use an array of tokens rather than just one.

see: http://dev.kohanaframework.org/issues/4125


zeelot:
Would rather keep this method simple. It's easy enough to extend if you decide to do things like this.

PR #148, opened 2011-07-25
Fix Request::redirect() when using an index_page

chrismeller:
Commit 035155ef185671deb13c broke request redirection when using an index_page a few days ago, since URL::base() expects the $index parameter to be a boolean, not a string.

### DIFF
--- a/classes/kohana/request.php
+++ b/classes/kohana/request.php
@@ -946,13 +946,13 @@ class Kohana_Request implements HTTP_Request {
 
 		if (strpos($referrer, '://') === FALSE)
 		{
-			$referrer = URL::site($referrer, TRUE, Kohana::$index_file);
+			$referrer = URL::site($referrer, TRUE, !empty(Kohana::$index_file));
 		}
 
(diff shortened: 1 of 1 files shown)


parshap:
On a related note, I thought `FALSE` should be used for the `index_file` setting when not using the index file as part your URLs. If so, [why does Kohana::$index_file get coerced into a string?](https://github.com/kohana/core/blob/3.2%2Fmaster/classes/kohana/core.php#L324)

samsoir:
@chrismeller your patch is valid (except it doesn't meet Kohana coding conventions, i.e. space between exclamation mark and function), but we also need an issue created on our [issue tracking](http://dev.kohanaframework.org) system before this can be pulled.

chrismeller:
Use it or don't, check out the IRC logs for more context. I'm done with Kohana, thanks.

samsoir:
Thanks, bye.

le6o:
http://dev.kohanaframework.org/issues/4141

PR #149, opened 2011-07-26
Buggy Arr::map()

shesek:
Arr::map() doesn't work well in those two cases:
1. Passing class functions (`array('class_name', 'function_name')`) or object methods (`array($object, 'method_name'`) is considered an array of callbacks, so 'class_name' and 'function_name' are considered two separate callbacks.
   I added an `!is_callbale()` check for those cases to make it work well. However, its not perfect - if you have `foo()`, `bar()`, and a foo class with a bar function (`foo::bar()`) there's no real way of telling if `array('foo', 'bar')` means `foo()` and `bar()` or `foo::bar()`. In this fix, it'll be considered as `foo::bar()`.
2. When the value of an element is an array, its always replaced - regardless of what was passed in the `$keys`  argument. For example, `Arr::map('strtoupper', array('a'=>array('bar'), 'b' => 'qux'), array())` ignores 'qux', but changes 'bar' to 'BAR'.

### DIFF
--- a/classes/kohana/arr.php
+++ b/classes/kohana/arr.php
@@ -363,27 +363,27 @@ class Kohana_Arr {
 	 * @param   array   array of keys to apply to
 	 * @return  array
 	 */
-	public static function map($callback, $array, $keys = NULL)
+	public static function map($callback, $array, array $keys = NULL)
 	{
 		foreach ($array as $key => $val)
 		{
+			if (NULL !== $keys and !in_array($key, $keys))
+				continue;
+
 			if (is_array($val))
 			{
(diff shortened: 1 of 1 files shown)


kiall:

!build

kiall:
Looks good, It would be great if you could:

A) Re-target this pull request to 3.3/develop as its a API change,
B) Add a redmine ticket for the change and amend the commit message to include "Fixes #nnnn"

Thanks :)

zeelot:
Cleaning up old PRs. This is a valid PR but I would rather not complicate the method at this point… I'm going to look into simplifying the functionality further in the next major release.

Sorry this stagnated.

PR #151, opened 2011-08-13
Fixes a problem with redirect when you have Kohana::$index_file set.

tylermorten:
For whatever reason, whenever I use Request::current()->redirect and I
have Kohana::$index_file set to index.php, the framework refuses to
attach 'index.php' onto the base url.  If I cast the third argument to
bool, it fixes the problem without messing you up if you happen to have
Kohana::$index_file set to FALSE.

### DIFF
--- a/classes/kohana/request.php
+++ b/classes/kohana/request.php
@@ -936,13 +936,13 @@ class Kohana_Request implements HTTP_Request {
 
 		if (strpos($referrer, '://') === FALSE)
 		{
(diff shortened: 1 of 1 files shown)


BRMatt:
Thank you for contributing this, unfortunately this bug has already been fixed in b7869b16248fd7ed24ac

tylermorten:
Sorry.  I was not using the 3.2/develop branch as you can probably tell.  Thanks.

PR #152, opened 2011-08-15
Adding in Debug::log_var which will log a variable to the logs

ae0000:
This provides a quick and easy way to debug to logs. It uses the default log writer. Not 100% on the name of it. 'log' is obviously taken.

### DIFF
--- a/classes/kohana/debug.php
+++ b/classes/kohana/debug.php
@@ -237,6 +237,21 @@ class Kohana_Debug {
 	}
 
 	/**
+	 * Saves a variable to the logs. 
+	 *
+	 *     // Logs the initial request object to the default log
+	 *     Debug::log_var('request object', Request::$initial);
+	 *
+	 * @param   string	variable identifier 
+	 * @param   mixed   variable to log 
+	 * @return  void
+	 */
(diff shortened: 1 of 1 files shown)


kiall:
I'm not sure I like this - I personally think Kohana's logs should be like nearly every other log you find in /var/log .. 1 line = 1 log entry. Otherwise, it becomes very difficult to parse in a meaningful way.

eg logging a route object results in this:

```
<?php defined('SYSPATH') or die('No direct script access.'); ?>

2011-08-16 15:18:01 --- DEBUG: test : Route Object
(
    [_callback:protected] => 
    [_uri:protected] => (<controller>(/<action>(/<id>)))
    [_regex:protected] => Array
    (
    )

    [_defaults:protected] => Array
    (
        [controller] => welcome
        [action] => index
    )

    [_route_regex:protected] => #^(?:(?P<controller>[^/.,;?\n]++)(?:/(?P<action>[^/.,;?\n]++)(?:/(?P<id>[^/.,;?\n]++))?)?)?$#uD
)
```

ae0000:
Fair enough - I find it useful to dump data like that, but I understand the reasoning behind keeping it standard.

PR #156, opened 2011-08-25
Patch #4231

daveWid:
Adding the ability to specify the prefix for a controller action. Patch #4231


daveWid:
If you see the most recent push into this patch I moved all of the logic into the Controller and out of the Request class as suggested on the ticket.

https://github.com/daveWid/core/commit/9a2e0be653d3027b9da8dfb7914bbbbd5283cb2b

kiall:
Closed per comment in ticket.

PR #159, opened 2011-09-12
4256 post superglobal

AlexC:
http://dev.kohanaframework.org/issues/4256


kiall:
Merged to 3.2 and 3.3 - Thanks

PR #160, opened 2011-09-14
Clarify syntax for passing array callbacks to Arr::map

acoulton:
Per the forum discussion and [issue #4259](http://dev.kohanaframework.org/issues/4259), a patch that updates the method signature and comments for Arr::map to clarify how to use it with array-form callbacks.

I think clarifying the method signature probably the best/only way to resolve, as I can't see how you could change the method to accept an array callback without entirely reverting the feature introduced by [#3381](http://dev.kohanaframework.org/issues/3381).

### DIFF
--- a/classes/kohana/arr.php
+++ b/classes/kohana/arr.php
@@ -349,21 +349,30 @@ class Kohana_Arr {
 	}
 
 	/**
-	 * Recursive version of [array_map](http://php.net/array_map), applies the
-	 * same callback to all elements in an array, including sub-arrays.
+	 * Recursive version of [array_map](http://php.net/array_map), applies one or more
+	 * callbacks to all elements in an array, including sub-arrays.
 	 *
 	 *     // Apply "strip_tags" to every element in the array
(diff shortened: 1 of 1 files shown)


kiall:

!build

kiall:
Looks like this failure wasn't actually that bug :)

<pre>06:08:46  There was 1 error:
06:08:46  
06:08:46  1) Kohana_ArrTest::test_map with data set #1 ('strip_tags', array(array('<p>foobar</p>'), array('<p>foobar</p>')), NULL, array(array('foobar'), array('foobar')))
06:08:46  Undefined variable: callback
06:08:46  
06:08:46  /var/lib/jenkins/jobs/kohana-3.2-pull/workspace/system/classes/kohana/arr.php:381
06:08:46  
06:08:46  FAILURES!
06:08:46  Tests: 1214, Assertions: 2836, Errors: 1, Incomplete: 1, Skipped: 69.</pre>

acoulton:
Could have sworn I'd run the tests at the time, but obviously not. Need to tighten up my own pre-pull process!

PR #162, opened 2011-09-16
Request Issues 4261

kanema:
http://dev.kohanaframework.org/issues/4261


shadowhand:
As per my comments on the Redmine issue, this is better solved by #192.

PR #163, opened 2011-09-29
Issue/4276

le6o:
Fix to prevent illegal offset error.


samsoir:
Grr, never mind - I'm being silly.

PR #164, opened 2011-10-01
Check for PHP 5.4 before doing get_magic_quotes_gpc

kudos:
Check for PHP 5.4 before doing get_magic_quotes_gpc as it has been deprecated.

### DIFF
--- a/classes/kohana/core.php
+++ b/classes/kohana/core.php
@@ -324,8 +324,8 @@ class Kohana_Core {
 			Kohana::$index_file = trim($settings['index_file'], '/');
 		}
 
-		// Determine if the extremely evil magic quotes are enabled
-		Kohana::$magic_quotes = (bool) get_magic_quotes_gpc();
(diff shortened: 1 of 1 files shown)


classes/kohana/core.php
```diff
@@ -324,8 +324,8 @@ class Kohana_Core {
 			Kohana::$index_file = trim($settings['index_file'], '/');
 		}
 
-		// Determine if the extremely evil magic quotes are enabled
-		Kohana::$magic_quotes = (bool) get_magic_quotes_gpc();
+		// Determine if the extremely evil magic quotes are enabled, it has been deprecated in 5.4
+		Kohana::$magic_quotes = (bool) version_compare(PHP_VERSION, '5.4.0') < 0 && get_magic_quotes_gpc();
```
ThePixelDeveloper, line 328: The boolean cast is unnecessary

PR #165, opened 2011-10-01
Check for PHP 5.4 before doing get_magic_quotes_gpc as it has been deprecated.

kudos:
Pull-request updated to work for pre-release versions of 5.4 (e.g. 5.4.0-beta1), removed bool cast, changed request from master branch to develop.

### DIFF
--- a/classes/kohana/core.php
+++ b/classes/kohana/core.php
@@ -325,7 +325,7 @@ class Kohana_Core {
 		}
 
 		// Determine if the extremely evil magic quotes are enabled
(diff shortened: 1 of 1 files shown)


samsoir:
Does this have a ticket number yet? If so can you add that to the commit reference. It will not be accepted without a ticket id.

kudos:
Commits squashed and updated with ticket number.

kiall:
Merged to 3.2 and 3.3 - Thanks!

PR #166, opened 2011-10-03
#4013 Arr::get should use array_key_exists instead of isset

d4rky-pl:
Currently Arr::get() relies on isset so it can't handle NULL values:

eg.
$test = array('foo' => 'foo' , 'bar' => NULL);
$test = Arr::get($test, 'bar', 'default');

Test is 'default' when it should be NULL. Fix is to use array_key_exists instead of isset.


d4rky-pl:
GitHub's inline edit sucks. Sorry.

PR #167, opened 2011-10-03
#4013 Arr::get should use array_key_exists instead of isset

d4rky-pl:
Currently Arr::get() relies on isset so it can't handle NULL values:

eg.
$test = array('foo' => 'foo' , 'bar' => NULL);
$test = Arr::get($test, 'bar', 'default');

Test is 'default' when it should be NULL. Fix is to use array_key_exists instead of isset.

### DIFF
--- a/classes/kohana/arr.php
+++ b/classes/kohana/arr.php
@@ -274,7 +274,7 @@ class Kohana_Arr {
 	 */
 	public static function get($array, $key, $default = NULL)
 	{
-		return isset($array[$key]) ? $array[$key] : $default;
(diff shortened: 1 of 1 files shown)


kiall:

!build

kiall:
Looks like a real build failure, rather than that build bug I just fixed..

kiall:
Looks like a duplicate of pull #176 / Issue #4012 - which passes all tests and works with ArrayObjects (validation uses these IIRC).

Thanks though :)

d4rky-pl:
Sure, I'm glad someone fixed it right :)

PR #169, opened 2011-10-27
Change $_SERVER['REQUEST_URI'] to $_SERVER['DOCUMENT_URI'] in detect_uri(

Girt:
Change $_SERVER['REQUEST_URI'] to $_SERVER['DOCUMENT_URI'] in detect_uri() function.
It is for SSI.
REQUEST_URI at SSI does not change and DOCUMENT_URI shows the real URI for subrequest

### DIFF
--- a/classes/kohana/request.php
+++ b/classes/kohana/request.php
@@ -259,9 +259,9 @@ class Kohana_Request implements HTTP_Request {
 		}
 		else
 		{
-			// REQUEST_URI and PHP_SELF include the docroot and index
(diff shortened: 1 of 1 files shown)


samsoir:
This will include `index.php` in the URI even if it has been rewritten to exclude it. For why would we want that?

samsoir:
Additionally, `$_SERVER['DOCUMENT_URI']` is a volatile global variable, and there is no guarantee it will be available across all platforms. Please see the PHP manual: http://www.php.net/manual/en/reserved.variables.server.php

PR #172, opened 2011-11-03
3.2/develop

cphrmky:
Calls to Kohana_Request redirect() that execute in a sub-request result in the return value of Request::$initial->execute() ( in the initial non sub-request request ) being and empty string instead of an instance of Response.  I've added one line to redirect() to simply void out if the currently executing request isn't the initial.  With this in place calls to redirect in sub-requests (which are _everywhere_ in apps that want pre-HMVC code) still return a response object so you can $response->status() and whatever else.

### DIFF
--- a/classes/kohana/request.php
+++ b/classes/kohana/request.php
@@ -947,6 +947,9 @@ class Kohana_Request implements HTTP_Request {
 	 */
 	public function redirect($url = '', $code = 302)
 	{
+		if(Request::initial() !== Request::current())
+			return NULL;
+		
 		$referrer = $this->uri();
 		$protocol = ($this->secure()) ? 'https' : TRUE;
 


samsoir:
I appreciate the idea and I am willing to accept this, but I think the implementation requires a little more thought. Rather than returning `NULL`, we should return a Response object with the `Location:` header set correctly. This would allow the caller to handle the sub-request redirect as they wish.

kiall:
This has been fixed in the 3.3/feature/RequestResponseCleanup[1] branch which should be merged into 3.3 soon...

The fix we went with was that Redirects should return the same response object that a external request would, aka a `Response` with a `Location` header.

[1] https://github.com/kohana/core/tree/3.3/feature/RequestResponseCleanup

PR #173, opened 2011-11-03
Paths extract

zbigniewmotyka:
These changes make it possible to extract values ​​from an array using paths instead of keys. This gives you the ability to filter data in multidimensional arrays rather than as it was until now a one-dimensional arrays.


kiall:
Updated to include tests and point at the 3.3 branch @ https://github.com/kohana/core/pull/189

PR #174, opened 2011-11-04
added wrapper for closure to make it serializable, fixed bug #4301

chvanikoff:
Closure wrapper code was generally taken from http://www.htmlist.com/development/extending-php-5-3-closures-with-serialization-and-reflection/ (fixed code to satisfy Kohana cs, added docblocks)

Route::__construct() wraps closure with that class to make it serializable (fixes http://dev.kohanaframework.org/issues/4301)

### DIFF
--- a/classes/kohana/php/closure.php
+++ b/classes/kohana/php/closure.php
@@ -1,0 +1,180 @@
+<?php defined('SYSPATH') or die('No direct script access.');
+
+/**
+ * Wrapper for closures to make it serializable
+ */
+class Kohana_PHP_Closure {
+    
+    /**
+     * @var Closure
+     */
+    protected $closure = NULL;
+    
+    /**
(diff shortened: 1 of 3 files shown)


zeelot:
Just a note, that class was created for fun and was never meant for production use. I haven't checked back with Jeremy but as far as I know, it has still never been used in production anywhere.

isaiahdw:
Have you benchmarked this? I think the best solution is to just not cache closures. Reflections/eval/reading the file from the disk every time doesn't sound very quick to me. So it would kind of defeat the whole point of caching ;)

chvanikoff:
Not yet. Will do a little bit later.
What about skipping caching - what if user will use only cached routes? I mean
-check cache
-if cache not set then write cache
-get from cache
That means lambda-routes will not be written to cache but user will not know about it and its even worse than throw exception and not allow to cache it I think.

PR #175, opened 2011-11-04
added wrapper for closure to make it serializable, fixed bug #4301

chvanikoff:
Closure wrapper code was generally taken from http://www.htmlist.com/development/extending-php-5-3-closures-with-serialization-and-reflection/ (fixed code to satisfy Kohana cs, added docblocks)

Route::__construct() wraps closure with that class to make it serializable (fixes http://dev.kohanaframework.org/issues/4301)

### DIFF
--- a/classes/kohana/php/closure.php
+++ b/classes/kohana/php/closure.php
@@ -1,0 +1,180 @@
+<?php defined('SYSPATH') or die('No direct script access.');
+
+/**
+ * Wrapper for closures to make it serializable
+ */
+class Kohana_PHP_Closure {
+    
+    /**
+     * @var Closure
+     */
+    protected $closure = NULL;
+    
+    /**
+     * @var ReflectionFunction  reflection of the closure
+     */
+    protected $reflection = NULL;
+    
+    /**
+     * @var string  closure code string
+     */
+    protected $code = NULL;
+    
+    /**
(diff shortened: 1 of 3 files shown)


zeelot:
Just a note (again), that class was created for fun and was never meant for production use. I haven't checked back with Jeremy but as far as I know, it has still never been used in production anywhere.

chvanikoff:
For lambda-routes it works - I've tested it. But I can also make some unit-tests later.

zeelot:
I'm mostly worried about performance issues at the moment

chvanikoff:
Well I think performance of such caching would be better than without it at all. Lambda-routing is a great Kohana feature but make people select between possibility to cache routes and use lambda is not good idea I think.

cs278:
Using `eval()` is really not a great idea.

chvanikoff:
I agree but is there another solution to avoid it?

chvanikoff:
Oops... Last commit was not for this pull request - how can I remove it from here?

acoulton:
You'll have to revert the commit in your 3.2/master, and then create a new branch in your repo for use until this pull is merged or closed. Generally it's always a good idea to use a specific topic branch for pull requests, as Github will add all future commits on that branch to the pull.

shadowhand:
No.

PR #176, opened 2011-11-04
3.1/fix 4012

Ikke:
Fixes [#4012](http://dev.kohanaframework.org/issues/4012)

### DIFF
--- a/classes/kohana/arr.php
+++ b/classes/kohana/arr.php
@@ -274,7 +274,11 @@ class Kohana_Arr {
 	 */
 	public static function get($array, $key, $default = NULL)
 	{
-		return array_key_exists($key, $array) ? $array[$key] : $default;
+		if (array_key_exists($key, $array)
+		    OR ($array instanceof ArrayAccess AND $array->offsetExists($key)))
+			return $array[$key];
+		else
+			return $default;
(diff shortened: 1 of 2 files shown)


Ikke:
I fixed the unittests and added some new unittests to handle for all situations.

`arr::get()` should now return the value even if the value == NULL and still work for ArrayAccess objects.

Fixes #4012

kiall:
Apologies, Just restored CI server from backup .. Its not quite fixed yet ;) Ignore those last messages!

!build

kiall:
@Ikke Looks good to me, Considering this changes the behaviour of how NULL values are treated, I think this needs to go into the 3.3/develop branch.

Also - There are a few Coding Style issues:

1) Indentation inside the test provider seems wrong, probably the wrong combination of spaces and tabs..
2) the if / else block needs {'s per the CS guide

Thanks :)

shadowhand:
I don't see a reason to change the current behavior of 3.1, as it is nearing its EOL.

Ikke:
There is a new pull request for 3.3

On Thu, Jan 5, 2012 at 7:25 AM, Woody Gilk
reply@reply.github.com
wrote:

PR #180, opened 2011-11-14
Upload::type and Upload::size bugs fixed

aaronsofaly:
The current code returns true if the uploaded file was not uploaded successfully.  I changed it to return false.

### DIFF
--- a/classes/kohana/upload.php
+++ b/classes/kohana/upload.php
@@ -146,7 +146,7 @@ class Kohana_Upload {
 	public static function type(array $file, array $allowed)
 	{
 		if ($file['error'] !== UPLOAD_ERR_OK)
-			return TRUE;
+			return FALSE;
 
 		$ext = strtolower(pathinfo($file['name'], PATHINFO_EXTENSION));
 
@@ -177,7 +177,7 @@ class Kohana_Upload {
 		if ($file['error'] !== UPLOAD_ERR_OK)
 		{
 			// The upload failed, no size to check
-			return TRUE;
+			return FALSE;
 		}
 
 		// Convert the provided size to bytes for comparison


shadowhand:
Don't you think this "bug" would have been fixed long ago if it was actually a bug?

aaronsofaly:
Are you saying it's not a bug or it's already fixed?  Because it's clearly a bug.  But if it's been fixed already, then I apologize as I didn't see it fixed.

shadowhand:
It is not a bug. This test `$file['error'] !== UPLOAD_ERR_OK` ensures that something was uploaded before testing the type, size, etc. Effectively, it prevents working on `NULL` uploads. If you require something to be uploaded, use `Upload::not_empty` followed by other rules.

aaronsofaly:
Exactly, and right now the code in that if statement returns true if the file was not uploaded successfully.  Returning true means the uploaded file passed the Upload::type check.  It needs to return false so the file fails the Upload::type check.

shadowhand:
No, you do not understand. A `NULL` upload is always valid, because nothing can be checked. If you want to make an upload required, use `Upload::not_empty`.

aaronsofaly:
I'm not trying to make an upload required.  I have a file that I'm uploading that I'm checking the file extension against with the Upload::type method.

If the uploaded file extension matches the allowed file extensions then the file passes the checks.  However, if the uploaded file has errors, then it also passes the checks since it returns true on if ($file['error'] !== UPLOAD_ERR_OK).

UPLOAD_ERR_OK means the file uploaded with success.  So that if statement is saying if the file did not upload with success then return true.

Can you please look at the Upload::type method?  This line: "return in_array($ext, $allowed);" returns true if the the file extension of the uploaded file is in the array of allowed file extensions.  So we've established that returning true is what we want when the file extension is an allowed type.  Okay, now look at the "if ($file['error'] !== UPLOAD_ERR_OK)" line.  Currently it returns true if the uploaded file was not uploaded successfully.  If it was not uploaded successfully, then it should not return true.

shadowhand:
The code in question has been unchanged for years and is being used successfully by thousands of people. Take a walk, come back, and read the code again. There is no bug here.

aaronsofaly:
It may have been like this for years, but that does not mean there isn't a bug here.  The reason it hasn't been caught yet is because the bug occurs if you are checking an uploaded file's extension and the uploaded file has errors.  How many people does this bug affect?  Probably a very small amount of people.  But just arguing the fact that this is the way it is because it's always been this way is invalid.  If you actually take the time to look at the code then you will see a bug.  Otherwise I can put together a demo and give you a file that has errors on it.  And when you upload that file and that file is NOT an allowed file extension, then it will still pass the error check.

shadowhand:
Okay, if this is a bug, what would be the syntax to allow only "jpg" files to be uploaded **without** making it required to upload a photo?

aaronsofaly:
``` php
$form = Validation::factory(array_merge($_FILES, $_POST));
if ( ! empty($form['field']['name']))
{
    $form->rule('field', 'Upload::type', array(':value', array('jpg')));
}
if ($form->check())
{
    // Code goes here if the form checks were passed
}
```

This code works perfectly fine.

But where this code fails is if you upload something like a .txt file (anything other than a .jpg file) AND that file has errors.  Since the file has errors, it returns true on the if ($file['error'] !== UPLOAD_ERR_OK) check and therefore passes the error checking even though it should fail because it's a .txt file and not a .jpg.

acoulton:
I think you need to add an Upload::valid check to your validation too.

aaronsofaly:
@acoulton A file with errors will still pass the Upload::valid check.

```
public static function valid($file)
{
    return (isset($file['error'])
        AND isset($file['name'])
        AND isset($file['type'])
        AND isset($file['tmp_name'])
        AND isset($file['size']));
}
```

isset() will return true on anything except NULL as long as the var exists.  So if $form['field']['error'] equals 0 or 1, it's still valid according to this method because it exists and it does not equal NULL.

The whole point of this bug fix though has to do with the Upload::type method

```
public static function type(array $file, array $allowed)
{
    if ($file['error'] !== UPLOAD_ERR_OK)
        return TRUE;

    $ext = strtolower(pathinfo($file['name'], PATHINFO_EXTENSION));

    return in_array($ext, $allowed);
}
```

The `return in_array($ext, $allowed);` line returns true if the file extension of the uploaded file is in the array of allowed file extensions.  So returning true means the file **PASSED** the checks.  So then why does the `if ($file['error'] !== UPLOAD_ERR_OK)` line return true if the file was not uploaded successfully?  If it didn't upload successfully, then it should return **FALSE**.

shadowhand:
For the last time: **optional file uploads**.

aaronsofaly:
Can you clarify your last comment?

shadowhand:
I feel like this has been pretty clear from the beginning... Unless you return `TRUE` when no file is uploaded, it is impossible to have uploads that are not required but type checked. Returning `FALSE` would make it impossible to have eg: a non-required "avatar" upload.

acoulton:
I'm with @shadowhand on that (for what it's worth!) but I do see your point @anarchtica. I think there are possibly two approaches - either change Upload::valid() to test whether the upload has errors or not (tbh, I thought it did) eg:

``` php
public static function valid($file)
{
    return (isset($file['error'])
        AND ($file['error'] === UPLOAD_ERR_OK)
        AND isset($file['name'])
        AND isset($file['type'])
        AND isset($file['tmp_name'])
        AND isset($file['size']));
}
```

or, change the test in Upload::type to distinguish between whether or not a file has been uploaded - eg:

``` php
public static function type(array $file, array $allowed)
{
    switch($file['error'])
    {
        case UPLOAD_ERR_NO_FILE:
            return TRUE;

        case UPLOAD_ERR_INI_SIZE:
        case UPLOAD_ERR_FORM_SIZE:
        case UPLOAD_ERR_PARTIAL:
        case UPLOAD_ERR_NO_TMP_DIR:
        case UPLOAD_ERR_CANT_WRITE:
        case UPLOAD_ERR_EXTENSION:
             return FALSE;

        case UPLOAD_ERR_OK:
             // Check the details
    }

    $ext = strtolower(pathinfo($file['name'], PATHINFO_EXTENSION));

    return in_array($ext, $allowed);
}
```

The issue with this latter approach is that if you're now testing two things within both Upload::type and Upload::size. Which means if you use the failed validation rules to generate user messages, your user is going to get both these:
- The file extension was not valid
- The file was too large

In situations where the connection was interrupted, for example, and actually you should be telling them "There was a problem, please try again".

Either way, the pull here isn't the right solution and I would suggest that either updating Upload::valid or adding a new Upload::successful or similar is a better way to go.

shadowhand:
> change Upload::valid() to test whether the upload has errors or not

This is what `Upload::not_empty` is for.

aaronsofaly:
Okay, I see your point of allowing the Upload::type method to return true if no file was uploaded.  But it should be done like this:

```
if ($file['error'] === UPLOAD_ERR_NO_FILE)
    return TRUE;
```

This way if no file was uploaded it will return true.  Otherwise, what happens with the old way is it will return true if the file uploaded has errors.

acoulton:
@shadowhand, unless I'm missing something, I do think there's an edge case missing here but @anarchtica's fix is not the right one.

`Upload::not_empty` makes the upload **required** - but is currently the only method AFAIK that checks for errors.

What would be your rule combination to verify that an **optional** upload was of the correct type and size and had uploaded without errors?

aaronsofaly:
@acoulton I agree that this pull request is no longer valid.  I think what's best is to add the `AND ($file['error'] === UPLOAD_ERR_OK)` check to the Upload::valid method and change the Upload::type method to `if ($file['error'] === UPLOAD_ERR_NO_FILE)`.

I can create a new pull request if everybody agrees.

aaronsofaly:
For the record, if you use Upload::type to test an uploaded file is an allowed file extension, you must also add the Upload::not_empty rule if this field is required.  Otherwise if your uploaded file has an error, it will pass the Upload::type check.  Adding the Upload::not_empty rule makes sure the uploaded file is valid.

acoulton:
@anarchtica I think `Upload::type` and `Upload::size` should stay the same for the reasons at the end of my post above - you don't want to tell the user that the upload was the wrong type if actually the file they tried to upload is fine but there was an error in transmission - that's really poor UX.

The only thing that should be changed/added is a test for upload errors as a separate thing from validating size and type.

Probably the best thing is if you create a ticket on http://dev.kohanaframework.org which is actually where this discussion should be happening, and the core devs can pitch in with their thoughts on how best to handle it.

shadowhand:
> Probably the best thing is if you create a ticket on http://dev.kohanaframework.org which is actually where this discussion should be happening, and the core devs can pitch in with their thoughts on how best to handle it.

Yes.

PR #181, opened 2011-11-22
Add default id as $name for label, input and select

14113:
If I create form with label and input, it could by fine, when they have the same id..


shadowhand:
Please do not resubmit your pull request, this functionality has been suggested before and rejected.

PR #182, opened 2011-11-23
Session data serialization

parshap:
I recently needed to override how session data is serialized to store it as JSON-encoded strings (for interoperability). I found this difficult to do as the serialization (and "encoding") calls were tightly coupled with the `Session#read` and `Session#__toString` methods. The changes I've made here make this much easier to do.

I know it's yet another layer of function calls - but it should only be 4 calls per Session instance (internal requests not included ... i.e., 4 additional calls per request).

If this is deemed valuable, I can create the associated Redmine ticket.

### DIFF
--- a/classes/kohana/session.php
+++ b/classes/kohana/session.php
@@ -128,7 +128,7 @@ abstract class Kohana_Session {
 	/**
 	 * Session object is rendered to a serialized string. If encryption is
 	 * enabled, the session will be encrypted. If not, the output string will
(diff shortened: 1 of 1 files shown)


shadowhand:
Issue created: http://dev.kohanaframework.org/issues/4379

shadowhand:
Fixed by https://github.com/kohana/core/commit/43804ad9b42b3e4a1bc9e0d2ddce8cc84373c15e

PR #183, opened 2011-11-26
Fixed bug with incorrect mime type

stillman:
Issue [#4344](http://dev.kohanaframework.org/issues/4344).

### DIFF
--- a/classes/kohana/response.php
+++ b/classes/kohana/response.php
@@ -446,7 +446,7 @@ class Kohana_Response implements HTTP_Response {
 			if ( ! isset($mime))
 			{
 				// Get the mime type
-				$mime = File::mime($filename);
+				$mime = File::mime($download);
 			}
 
 			// Open the file for reading


shadowhand:
Why would using `$filename` in this context cause an issue?

stillman:
Because file name that will be show in the download/open file dialog is in the $download variable and it seems logical to get mime type for it. Original file may not have extension at all (especially when download is generated from the current response [Use TRUE for the filename to send the current response as the file content])

cs278:
If `$filename` is true it uses `$download` for the MIME type.

stillman:
Okay, I want to give the user file named "picture.jpg", but the original file stored as md5_file($file) (without extension). What mime type should be in this case?

shadowhand:
Ah, I see... the sent filename should match the sent mime-type.

Please [create an issue](http://dev.kohanaframework.org/projects/kohana3/issues) and reference the pull request so that this doesn't get lost. Editing the pull request to also edit the ticket gets bonus points.

Thanks!

stillman:
Already did :)
http://dev.kohanaframework.org/issues/4344

PR #185, opened 2011-12-15
Added support for PHP 5.3+ namespaces in routing.

colinbm:
Allows you to create a route like:

``` php
<?php # init.php

namespace My\Discussion;

\Route::set('my-discussions', 'discussions')
    ->defaults(array(
        'namespace'  => 'My\\Discussion',
        'controller' => 'discussion',
        'action'     => 'index'
    ));
```

This allows a module to be separated by namespace. If you don't specify a namespace, then the 5.3 syntax will not be used, so it is still backwards compatible with <5.3.

### DIFF
--- a/classes/kohana/core.php
+++ b/classes/kohana/core.php
@@ -488,8 +488,7 @@ class Kohana_Core {
 		try
 		{
 			// Transform the class name into a path
-			$file = str_replace('_', '/', strtolower($class));
-			$file = str_replace('\\', '/', $file);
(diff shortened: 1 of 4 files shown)


kiall:
This will probably tie in with http://dev.kohanaframework.org/issues/4001

colinbm:
@kiall Looks good - discovered in practice that my changes don't really work, as the core then tries to run the code in the module, and can't find it as it's in a different namespace. So will leave this to your linked issue and close/withdraw this pull request. - Thanks.

PR #192, opened 2012-01-03
#4337 File Uploads in Requests

shadowhand:
Adds new `Request::files()` method, along with support for uploading files via the HTTP, Stream, and Curl clients.

### DIFF
--- a/classes/Kohana/Request.php
+++ b/classes/Kohana/Request.php
@@ -1340,10 +1340,6 @@ class Kohana_Request implements HTTP_Request {
 	/**
 	 * Gets or sets HTTP FILE parameters to the request.
 	 *
-	 * [!!] The initial request will have `$_FILES` content attached using
-	 * this method. In this case, the return value of `files($key)` will be
-	 * an array rather than a string.
-	 *
 	 * @param   mixed   $key   Key or key value pairs to set
 	 * @param   string  $file  full path to file
 	 * @return  mixed
--- a/classes/Kohana/Request.php
+++ b/classes/Kohana/Request.php
@@ -627,7 +627,7 @@ class Kohana_Request implements HTTP_Request {
 	 * @var array    cookies to send with the request
 	 */
 	protected $_cookies = array();
-	
+
 	/**
 	 * @var array   files to upload with the request
 	 */
@@ -1336,39 +1336,43 @@ class Kohana_Request implements HTTP_Request {
 
 		return $this;
 	}
-	
+
 	/**
 	 * Gets or sets HTTP FILE parameters to the request.
 	 *
-	 * @param   mixed  $key   Key or key value pairs to set
-	 * @param   string $file  Value to set to a key
+	 * [!!] The initial request will have `$_FILES` content attached using
+	 * this method. In this case, the return value of `files($key)` will be
+	 * an array rather than a string.
+	 *
+	 * @param   mixed   $key   Key or key value pairs to set
+	 * @param   string  $file  full path to file
 	 * @return  mixed
 	 */
-    public function files($key = NULL, $file = NULL)
-    {
-        if (is_array($key))
-        {
-            // Act as a setter, replace all files
-            $this->_files = $key;
-            
-            return $this;
-        }
-        
-        if ($key === NULL)
-        {
-            // Act as a getter, all files
-            return $this->_files;
-        }
-        elseif ($file === NULL)
-        {
-            // Act as a getter, single file
-            return Arr::get($this->_files, $key);
-        }
-        
-        // Act as a setter, single file
-        $this->_files[$key] = $file;
-        
-        return $this;
-    }
+	public function files($key = NULL, $file = NULL)
+	{
+		if (is_array($key))
+		{
+			// Act as a setter, replace all files
+			$this->_files = $key;
+
+			return $this;
+		}
+
+		if ($key === NULL)
+		{
+			// Act as a getter, all files
+			return $this->_files;
+		}
+		elseif ($file === NULL)
+		{
+			// Act as a getter, single file
+			return Arr::get($this->_files, $key);
+		}
+
+		// Act as a setter, single file
+		$this->_files[$key] = $file;
+
+		return $this;
+	}
 
 } // End Request
--- a/classes/Kohana/Request.php
+++ b/classes/Kohana/Request.php
@@ -627,6 +627,11 @@ class Kohana_Request implements HTTP_Request {
 	 * @var array    cookies to send with the request
 	 */
 	protected $_cookies = array();
+	
+	/**
+	 * @var array   files to upload with the request
+	 */
+	protected $_files = array();
 
 	/**
 	 * @var Kohana_Request_Client
@@ -1331,5 +1336,39 @@ class Kohana_Request implements HTTP_Request {
 
 		return $this;
 	}
+	
+	/**
+	 * Gets or sets HTTP FILE parameters to the request.
+	 *
+	 * @param   mixed  $key   Key or key value pairs to set
+	 * @param   string $file  Value to set to a key
+	 * @return  mixed
+	 */
+    public function files($key = NULL, $file = NULL)
+    {
+        if (is_array($key))
+        {
+            // Act as a setter, replace all files
+            $this->_files = $key;
+            
+            return $this;
+        }
+        
+        if ($key === NULL)
+        {
+            // Act as a getter, all files
+            return $this->_files;
+        }
+        elseif ($file === NULL)
+        {
+            // Act as a getter, single file
+            return Arr::get($this->_files, $key);
+        }
+        
+        // Act as a setter, single file
+        $this->_files[$key] = $file;
+        
+        return $this;
+    }
 
 } // End Request
--- a/classes/Kohana/Request/Client.php
+++ b/classes/Kohana/Request/Client.php
@@ -14,6 +14,11 @@
 abstract class Kohana_Request_Client {
 
 	/**
+	 * @var  string  HTTP end-of-line
+	 */
+	const CRLF = "\r\n";
+
+	/**
 	 * @var    Cache  Caching library for request caching
 	 */
 	protected $_cache;
--- a/classes/Kohana/Request/Client/Curl.php
+++ b/classes/Kohana/Request/Client/Curl.php
@@ -28,12 +28,6 @@ class Kohana_Request_Client_Curl extends Request_Client_External {
 		// Set the request method
 		$options[CURLOPT_CUSTOMREQUEST] = $request->method();
 
-		// Set the request body. This is perfectly legal in CURL even
-		// if using a request other than POST. PUT does support this method
-		// and DOES NOT require writing data to disk before putting it, if
-		// reading the PHP docs you may have got that impression. SdF
-		$options[CURLOPT_POSTFIELDS] = $request->body();
-
 		// Process headers
 		if ($headers = $request->headers())
 		{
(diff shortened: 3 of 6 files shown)


shadowhand:
#162 is most likely related to this as well.

shadowhand:
I'm pretty sure the test that is failing has incorrect expectations.

classes/kohana/request/client/curl.php
```diff
@@ -52,6 +46,35 @@ public function _send_message(Request $request)
...
+				foreach ($files as $name => $file)
+				{
+					// Set the files to upload
+					$post[$name] = '@'.$file;
+				}
+			}
+
+			// We really don't need to set the request body here, but it is
+			// good for debugging and the unit tests require it.
+			$request->body(http_build_query($post, NULL, '&'));
```
kiall: It looks like we have lost the ability to POST, say, some JSON again?

This should work correctly:

```
$request = Request::factory('/home/api');
$request->method(Request::POST);
$request->headers('Content-Type', application/json');
$request->body(json_encode(array(bla bla bla)));
$request->execute();
```

a POST is not always application/x-www-form-urlencoded :)

classes/kohana/request/client/stream.php
```diff
@@ -37,6 +37,50 @@ public function _send_message(Request $request)
...
+					));
+				}
+
+				$request->body(implode("\n", $body));
+				$request->headers('content-type', 'multipart/form-data; boundary='.$boundary);
+			}
+			else
+			{
+				$request->body(http_build_query($request->post(), NULL, '&'));
+				$request->headers('content-type', 'application/x-www-form-urlencoded');
```
kiall: Looks like the same issue I mentioned above?

> a POST is not always application/x-www-form-urlencoded :)

To me - this is a **vitally** important feature that we just cant break (again!)

kiall:
@shadowhand - re that failing test, the expectations are indeed correct in my opinion.

If you do the following:

```
$request = Request::factory('/home/api');
$request->method(Request::POST);
$request->body(json_encode(array(bla bla bla))); // Set the body explicitly
$request->post('foo', 'bar')'; // Set the body again implicitly
$request->execute();
```

We had decided the explicitly defined body would be used instead.

kiall:
Also, Initial vs Internal vs External request parity should be considered before merging.

For example, if the browser sends a file as part of the initial request, $reqiuest->files(); should contain the files in the same way that a subrequest would.

classes/kohana/request/client/curl.php
samsoir: Agreed, we need to support the full array of HTTP post methods and content types

samsoir:
My issue with this implementation is that is loses much of the functionality of HTTP. If we are going to handle files, which was always the intention, then we should handle them as specified in RFC 2387 (http://www.ietf.org/rfc/rfc2387.txt) rather than the PHP hacky way. Files should be handled as multipart/mixed type messages within the HTTP message body.

kiall:
@samsoir I was thinking about that too, I think its okay to introduce something kinda hacky for the moment.

But - I think its important the API is clean and consistent for Initial vs Internal vs External requests so the hack can be removed later without API changes..

samsoir:
@kiall the issue with hacky is that it _should_ be legal to have a `Request::body()` and `Request::files()` - so far, we will loose this with stream/native at the moment whatever we do.

shadowhand:
Slow down guys, we just need to decide whether `post()` or `body()` takes precedence here. My assumption is that if any POST data is present, it will override the body.

kiall:
The current assumption is that if you explicitly set a body, that body will win. Only if no body is set will the post() array be applied to the body.

shadowhand:
@kiall Actually, Kohana is currently assuming that POST data always wins, even when explicitly setting the body.

My latest commit fixes the following issues:
1. Sending form data using POST now works properly.
2. All clients ignore body content when POST data is present.
3. Custom Content-Types are now replaced when POST data is present.

However, the test that has been failing is **still failing because it is broken**. It uses a mock of `Request_Client_External`, which is no longer handling POST data, refs 7a7b4a19cd73ac640e444a8b02a0c0ed08c332d5. I am not sure how to properly fix the test.

kiall:
I was sure body() won out over post() it both were set, But looking over the 3.2 code is proving me wrong!

I'll have another proper look over all this in a few mins...

shadowhand:
The only remaining task on this pull request is to add `Request::files()`, which would contain the `$_FILES` content for initial requests.

acoulton:
Won't this also need work on the Upload class if Request::files is going to be usable instead of $_FILES for initial and internal requests? Currently upload makes heavy use of is_uploaded_file which will fail for internal requests...

classes/kohana/request/client/stream.php
```diff
@@ -37,6 +37,53 @@ public function _send_message(Request $request)
...
+				// Set multipart form boundary
+				$boundary = '----------------------------'.Text::random('hexdec', 12);
+
+				if ($post = $request->post())
+				{
+					foreach($post as $key => $val)
+					{
+						$body[] = implode("\n", array(
+							'--'.$boundary,
+							'Content-Disposition: form-data; name="'.HTML::chars($name).'"',
```
cs278: HTML escaping is incorrect here. See http://tools.ietf.org/html/rfc2388#section-3

shadowhand: Thanks @cs278, would using [iconv_mime_encode](http://php.net/iconv_mime_encode) be the correct solution here? There is also [quoted_printable_encode](http://php.net/quoted_printable_encode), but that seems to be a PHP 5.3+ function...

shadowhand: @cs278 Does 80d625533b5e78b55e71be15edaf1294bd46799d fix this correctly?

shadowhand:
@kiall and @samsoir: Now that 5fc99466c952403e3995ca74c3ecc31fae80f09a, do you think this is ready to be merged?

I have one lingering concern: https://gist.github.com/1573764 Any thoughts?

acoulton:
I think if you want initial - internal - external parity then Upload also needs to be modified to handle files in an internal request. It probably means adding something like:

``` php
class Kohana_Upload
{
    public static function is_uploaded_file($file)
    {
        if (Request::$current === Request::$initial)
        {
            return is_uploaded_file($file);
        }
        else
        {
            return TRUE;
        }
    }

    public static function move_uploaded_file($from, $to)
    {
        if (Request::$current === Request::$initial)
        {
            return move_uploaded_file($from, $to);
        }
        else
        {
            return rename($from, $to);
        }
    }
}
```

And then using these instead of the native functions in the relevant Upload:: methods? I can submit as a separate pull if you like.

I realise that there's a theoretical security issue here, in that if an attacker can convince KO that the initial request is a subrequest then they could "upload" an arbitrary file from anywhere on the server - however I don't think it's likely and anyway there's probably no way round that if you want to support files on internal requests.

shadowhand:
@acoulton, I really don't think that is an issue. Users can check if the value returned by `files($name)` is an array and handle it accordingly.

acoulton:
I'd obviously also misunderstood the intention here. From the talk of initial vs internal request parity, I was expecting that Request::files() would always return a suitable array in $_FILES format. Hence my comments about changes to the Upload class, as I'd thought the whole point was to be able to send an internal request that was able to be handled identically to an initial, as with other Request:: values.

EG, that I would be able to write a method (obv without validation etc):

```
   public function action_photo()
   {
       if ($file = $this->request->files('photo'))
       {
           Upload::save($file, 'photo.jpg');
       }
   }
```

This would be equivalent to the use of `$this->request->query` and `$this->request->post` which are specifically provided to make HMVC transparent to controller actions. I already had a couple of use cases in mind where this would make life easier.

What is the reason for **not** using the $_FILES format for internal requests? Aren't there more use cases where it would be helpful to be able to make an internal request to an existing action without changing the action code than cases where initial/internal matters?

shadowhand:
@acoulton It is not possible to fake `is_uploaded_file()`, among other things.

I don't see the point of all this extra processing that is being suggested. It simply is not (reasonably) possible to emulate the entire HTTP server. The goal here is to fix one issue: allow external requests to upload files. This pull request makes that possible. There is a reasonable level of parity between the two methods, all that is required is a simple `is_array()` call:

``` php
<?php

$file = $this->request->files('upload');

if (is_array($file))
{
    $file = Upload::save($file);
}

// Continue as normal
```

acoulton:
Except you're missing the else, if you want to accept files from internal requests...

```
$file = $this->request->files('upload');

if (is_array($file))
{
    // Bear in mind this will be more complex, because you're going to want to validate it
    $file = Upload::save($file);
}
elseif($file)
{
    // Implement your whole own identical functionality to Upload::save, apart from calling rename instead of move_uploaded_file
    // And you'll need to duplicate the validation logic etc.
    // Or first check if you have a file and from where, then validate it, then again check where the file came from to save it
}
```

I'm aware it's not possible to fake is_uploaded_file, hence my suggestion above that the Upload class be tweaked to use an alternative method that always returns TRUE for a subrequest (potentially only if the file exists).

If the intention is only to allow external requests to **send** files, rather than to make any difference to **receiving** files in HMVC requests, then I think Request::files() should only be relevant for outbound external requests and should simply be NULL for incoming (internal or initial) requests.

If the end developer has to do all the work around receiving files differently for an initial or internal request, providing a Request::files() interface just confuses things and doesn't really add any value, IMHO. It's clearer to know that you can use $_FILES for an initial request and you have to roll your own for internal.

kemo:
How about having the File class to provide similar capabilities like `SplFileInfo` (with "lazy loading"), having instances of it return the full path on `__toString()`? (yeah, this would include adding more non-static stuff to File, it'd be worth it)

Request::files() method always returning array of File objects seems way more consistent, extendable and foolproof than any of the current solutions.

And it could handle its own validation methods, leaving the Upload class intact for $_FILES validation only. Or fix it completely to check if current Request is initial to decide whether or not it should be using `is_uploaded_file`, as @acoulton suggested.

samsoir:
IMO consistency is key and I have to agree with @kiall that having differing formats for `Request::files()` between the `Request::initial()` instance and any other sub-requests is not a good idea. A lot of the issues we've had up until now are due to inconstancies through our interfaces, so we should be reducing those not adding to them.

Processing the `$_FILES` superglobal into `Request::initial()` is inefficient in most cases, so we shouldn't do it unless the code specifically invokes `Request::files()` on the initial request object. But it is critical that the response is always the same; doing a check for `Request::is_initial()` each time `Request::files()` is called is no better for efficiency.

FYI: for the record, 3.3 was scheduled to only support PHP 5.3.3 and above (http://dev.kohanaframework.org/projects/kohana3/repository/revisions/c8eb139db74b88fed2c9bccd6490f43a3328f1cc)

shadowhand:
Removed the initial setting of `$_FILES` to `Request::files()`, removed PHP 5.2 code.

What else needs to be done?

classes/kohana/request/client/stream.php
cs278: Looks good to me

shadowhand:
I brought this in sync with the `3.3/develop` branch. Any objections to merging this now?

classes/Kohana/Request.php
```diff
@@ -1329,4 +1334,42 @@ public function post($key = NULL, $value = NULL)
 		return $this;
 	}
 
+	/**
+	 * Gets or sets HTTP FILE parameters to the request.
+	 *
+	 * [!!] The initial request will have `$_FILES` content attached using
```
acoulton: Looks like this is no longer the case and - per the pull discussion - the files() method is only relevant for external requests in 3.3? Other than updating this comment, all this looks OK to me...

zeelot:
Closing for now… ticket got moved out of `3.3` and we're going to rethink the request/response classes down the road.

If anyone wants to clean this up for `3.4.0`, we can probably get it in.

PR #195, opened 2012-01-05
Fixes #4147, Translated messages should not rely on the default language being English

ernestas:


shadowhand:
What if two different languages have the same word or phrase but mean different things? For instance, if English has "No" meaning "No", and another language has "No" meaning "Two". In this situation, it is impossible to translate both correctly, as they are saying the same space.

TL;DR: To implement this change, the format of I18n files would have to be changed dramatically.

shadowhand:
One solution would be (`i18n/en.php`):

``` php
<?php defined('SYSPATH') or die('No direct script access.');

return array(
    'es' => array(
        'Español' => 'Spanish',
        '¡Hola, mundo!' => 'Hello, world!',
    ),
    'fr' => array(
        'Français' => 'French',
        'Bonjour, monde!' => 'Hello, world!',
    ),
);
```

Any thoughts on this format?

shadowhand:
Related: #199.

ernestas:
Reverted #4120 since it brings more complexity than the intent of the change. Will use gettext or https://github.com/czukowski/I18n_Plural

zeelot:
Closing since you seem to have gone with other options

PR #194, opened 2012-01-03
Issue #'s 4377, 4375, 4374, 4373

kiall:
Issue #'s 4377, 4375, 4374, 4373

PR #196, opened 2012-01-05
Route Callbacks/Filters

shadowhand:
http://dev.kohanaframework.org/issues/4388

### DIFF
--- a/classes/kohana/route.php
+++ b/classes/kohana/route.php
@@ -67,7 +67,7 @@ class Kohana_Route {
 	public static $cache = FALSE;
 
 	/**
-	 * @var  array 
+	 * @var  array
 	 */
 	protected static $_routes = array();
 
@@ -254,6 +254,11 @@ class Kohana_Route {
 	protected $_callback;
 
 	/**
+	 * @var  array  route filters
+	 */
+	protected $_filters = array();
+
+	/**
 	 * @var  string  route URI
 	 */
 	protected $_uri = '';
@@ -340,7 +345,7 @@ class Kohana_Route {
 	 *         'controller' => 'welcome',
 	 *         'action'     => 'index'
 	 *     ));
-	 * 
+	 *
 	 * If no parameter is passed, this method will act as a getter.
 	 *
 	 * @param   array   $defaults   key values
@@ -354,6 +359,41 @@ class Kohana_Route {
 		}
 
 		$this->_defaults = $defaults;
+
+		return $this;
+	}
+
+	/**
+	 * Filters to be run before route parameters are returned:
+	 *
+	 *     $route->filter(
+	 *         function(Route $route, $params)
+	 *         {
+	 *             if ($params AND $params['controller'] === 'welcome')
+	 *             {
+	 *                 $params['controller'] = 'home';
+	 *             }
+	 *
+	 *             return $params;
+	 *         }
+	 *     );
+	 *
+	 * Callbacks can be functions, methods, or closures.
+	 *
+	 * [!!] To prevent a route from matching, return `FALSE`.
+	 *
+	 * @throws  Kohana_Exception
+	 * @param   array   $callback   callback string, array, or closure
+	 * @return  $this
+	 */
+	public function filter($callback)
+	{
+		if ( ! is_callable($callback))
+		{
+			throw new Kohana_Exception('Invalid Route::callback specified');
+		}
+
+		$this->_filters[] = $callback;
 
 		return $this;
 	}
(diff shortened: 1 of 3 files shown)


shadowhand:
This probably needs a test or two before merging.

kiall:
Dont expect a comment from Jenkins anytime soon! will be offline for a bit.. HW node panicked 3 times last night due to memory issues.. trying to identify which stick is bad now.. memtest will likely take a few hours ;)

kiall:
+1 on this change btw, looks like it could be quite useful for APIs, unless we give routes a more direct way to handle HTTP methods..

Also, I think @Zeelot has something similar in hyla. Zeelot: how does this compare?

shadowhand:
I think @Zeelot is running the callbacks from inside `Request` so his callback signature looks like `function(Request $request, Route $route, $params)`... but he can confirm that.

I still want a different way of handling HTTP methods.

shadowhand:
@Zeelot's way is probably the correct solution, as it the most flexible without modifying `Route::matches()`.

shadowhand:
The [code in question](https://github.com/hyla/hyla/blob/1.0.x/develop/hyla/core/classes/route.php). In this code, Request has been modified to send itself to `Route::matches()` instead of the URI. I think that might be too large of change to make at this point, unless it fits in with #194 somehow.

zeelot:
Yeah I prefer having access to the entire request object so that the callbacks can look at more than just the URI which I think is important.

zeelot:
I am allowing the user to specify the parameters for the callbacks similarly to Validation rules but I'm not sure if you guys find that useful or if we should just always pass in the same things. The reason I had it was so you could reuse callbacks a little more easily like this:

```
->callback(array('Router', 'match_methods'), array(':request', array('GET', 'POST')))
```

I know you want a different way of handling http methods but maybe it's still useful for other things.

zeelot:
Looks like you're currently forcing the user to return the array of params on all filters, what do you think of just replacing params if the callback returns an array? From my experience, there are 2 types of callbacks I run; callbacks to enforce a specific thing (return FALSE in a condition, do nothing otherwise), and callbacks that slightly alter the params (if id is not set, change the action)

zeelot:
Sorry last comment =P I don't think the changes to Route and Request I made in Hyla are too large, In my opinion, routes need to match an entire request instead of just the URI so I would really like the change to make it in. I'm not sure at this point how much work it is with all the changes that @kiall has made (stopped keeping track to the request work)

kiall:
I'm going to try and break that up into a few smaller pulls ;)

shadowhand:
@Zeelot wrote:

> what do you think of just replacing params if the callback returns an array?

That is probably the least error prone way to do it, good idea.

> In my opinion, routes need to match an entire request instead of just the URI so I would really like the change to make it in.

I agree with this as well, it makes a lot of sense to do `Route::matches(Request $request)`.

At this point, I am not sure what to do about the variable parameters, it seems like a lot of extra overhead for something that could be easily accomplished by creating a closure that calls another method.

@kiall The build failed on something completely unrelated... any idea what that is about?

kiall:
Yea - Looks like I managed to move the wrong version of the HTTP_Cache class from core -> cache.. I think I grabbed the copy from my request/resonse/HVMC isolation stuff. Sorting it now..

kiall:
!build

shadowhand:
If everyone is in agreement, I vote for this to be merged right now, then we can start looking at changing how `Route::matches()` is called.

shadowhand:
!build

kiall:
Just a FYI, "build" is only needed to force a re-test when there have been no code changes, if you add/edit commits it will pick it up on the next check (every 15 mins)

shadowhand:
@kiall Okay, good to know.

classes/kohana/route.php
```diff
@@ -359,6 +364,42 @@ public function defaults(array $defaults = NULL)
...
+	 * [!!] Default parameters are added before filters are called!
+	 *
+	 * @throws  Kohana_Exception
+	 * @param   array   $callback   callback string, array, or closure
+	 * @return  $this
+	 */
+	public function filter($callback)
+	{
+		if ( ! is_callable($callback))
+		{
```
kiall, line 393: Minor Coding Style issue, the brackets aren't required for return/throw etc.. From memory I think our rules pick up on this..

shadowhand, line 393: Ah crap, I already merged. Can be fixed in develop branch if CS picks it up.

PR #197, opened 2012-01-05
Using closure at validation rule. Change comments and userguide

loguntsov:
In the Kohana userguide use of anonymous functions in rules validation isn't described.

### DIFF
--- a/classes/kohana/validation.php
+++ b/classes/kohana/validation.php
@@ -189,8 +189,11 @@ class Kohana_Validation implements ArrayAccess {
 	 *     // The "password" field must match the "password_repeat" field
 	 *     $validation->rule('password', 'matches', array(':validation', 'password', 'password_repeat'));
 	 *
+	 *     // Using closure (anonymous function)
+	 *     $validation->rule('index', function( $value ) { return $value > 6 && $value < 10 } , array(':value'));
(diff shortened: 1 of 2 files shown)


shadowhand:
Did you actually test this? I don't think you can use closures because no error name can be determined. (All closures have a callable name of `Closure::__invoke`.)

zeelot:
You can use a closure but you need to pass it `:validation` and set the error manually, fix the example to show this (and the branch issue, obviously) and we can pull it in :)

loguntsov:
Thanks for my mistake. 

Can be for closure's rules to make so:

$validation->rule('index', function($value) { return $value > 6 ? TRUE : 'error name'; }, array(':value'))

That is closure return TRUE if no errors else ERROR NAME for the error.
Good idea ? 

p.s. in my opinion to transfer validation object into closure is not good but it's universal way.

shadowhand:
Committed to 3.2 b9729110826c9974b7ab9fe7378d6033d9da311a and merged into 3.3 9c58f38aad8011857f3392c74fcf1e2525757205.

zeelot:
P.S. @loguntsov, feel free to file a feature request for allowing callbacks to return the error string instead of booleans.

PR #198, opened 2012-01-06
#3141 - Handle indexed arrays consistently in Arr::merge()

cbandy:
The first commit adds more tests for existing behavior. The second commit changes behavior for indexed arrays, and appears faster, too.

### DIFF
--- a/classes/kohana/arr.php
+++ b/classes/kohana/arr.php
@@ -391,7 +391,10 @@ class Kohana_Arr {
 	}
 
 	/**
-	 * Merges one or more arrays recursively and preserves all keys.
+	 * Recursively merge two or more arrays. Values in an associative array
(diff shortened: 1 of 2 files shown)


shadowhand:
This seems completely straight forward to me. +1

cbandy:
@acoulton The issue you are describing is captured elsewhere: http://dev.kohanaframework.org/issues/4324

acoulton:
@cbandy yes, I saw that these two are related - just wanted to note for clarity that although this fix will make the server arrays be merged as expected, it doesn't actually resolve #4324 which will still need those config arrays to be switched to associative. Some of the comment in #3141 isn't clear about this.

Definitely think this should be merged.

cs278:
The type hint has been removed from the parameters?

PR #199, opened 2012-01-09
3.3/feature/4299 driver based i18n

averyaube:


shadowhand:
@kiall Builds look broken again.

kiall:
@shadowhand nope - looks like the pull is at fault. Kinda..

> PHP Strict Standards:  Non-static method Kohana_I18n::lang() should not be called statically in /var/lib/jenkins/jobs/kohana-3.3-pull/workspace/application/bootstrap.php on line 56
> PHP Fatal error:  Using $this when not in object context in /var/lib/jenkins/jobs/kohana-3.3-pull/workspace/system/classes/kohana/i18n.php on line 61

Since this pull requires changes to the bootstrap, which is in a different repo we have a typical chicken and egg problem :) Although, it looks like the Fatal Error is an issue that needs fixing...

zeelot:
Can't change API in `3.3`… we're going to rework this class soon and I'll leave a link to this PR in the ticket.

PR #203, opened 2012-01-15
Clean URLs tutorial update

badsyntax:
Just a short bit of information as to why you might not get an 500 error (rewrite module installed correctly), but do get a 404 error (Apache not reading .htaccess file). I believe this information could be useful to some. 

This commit also includes a small grammar fix.


shadowhand:
Merged into 3.3/develop in 374419b1b75bf42e070b4d02beccaf1a930908ea.

PR #204, opened 2012-01-16
Fixed an assignment in a condition

guntars:


shadowhand:
99% sure this is not correct and breaks expected functionality.

guntars:
allright, you may be 99% right. false alarm by ide. closing.

PR #205, opened 2012-01-23
3.2/develop

ellisio:
Updated the PHPDoc block on the Request factory to enable code completion for most IDEs. Tested with Eclipse and PHPStorm.

Let me know if you want me to open a ticket somewhere, but I think this is extremely minor.

### DIFF
--- a/classes/kohana/request.php
+++ b/classes/kohana/request.php
@@ -48,7 +48,7 @@ class Kohana_Request implements HTTP_Request {
 	 * @param   string  $uri URI of the request
 	 * @param   Cache   $cache
 	 * @param   array   $injected_routes an array of routes to use, for testing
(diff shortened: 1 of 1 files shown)


shadowhand:
Merged into `3.3/develop` as well.

classes/kohana/request.php
```diff
@@ -48,7 +48,7 @@ class Kohana_Request implements HTTP_Request {
 	 * @param   string  $uri URI of the request
 	 * @param   Cache   $cache
 	 * @param   array   $injected_routes an array of routes to use, for testing
-	 * @return  void
+	 * @return  void|Request
```
cs278, line 51: Return value is always an instance of `Request` no need for `void|`

PR #208, opened 2012-01-24
#4417 - Support for additional parameters to Kohana_Log::add()

kiall:
See http://dev.kohanaframework.org/issues/4417

### DIFF
--- a/classes/kohana/log.php
+++ b/classes/kohana/log.php
@@ -21,17 +21,6 @@ class Kohana_Log {
 	const NOTICE    = LOG_NOTICE;   // 5
 	const INFO      = LOG_INFO;     // 6
 	const DEBUG     = LOG_DEBUG;    // 7
-	const STRACE    = 8;
-
-	/**
-	 * @var  string  timestamp format for log entries
-	 */
-	public static $timestamp = 'Y-m-d H:i:s';
-
-	/**
-	 * @var  string  timezone for log entries
-	 */
-	public static $timezone;
 
 	/**
 	 * @var  boolean  immediately write when logs are added
@@ -130,7 +119,7 @@ class Kohana_Log {
 	 * @param   array   $values     values to replace in the message
 	 * @return  Log
 	 */
-	public function add($level, $message, array $values = NULL)
+	public function add($level, $message, array $values = NULL, array $additional = NULL)
 	{
 		if ($values)
 		{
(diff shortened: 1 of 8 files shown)


shadowhand:
There is a lot of code duplication here. Shouldn't `format_entry()` be inside of `Log_Writer` if it is duplicated in each writer? Same goes for the time format stuff.

kiall:
@shadowhand yea, I'm already cleaning this up a bit

Re the Date stuff, I'm thinking I'll let it use the defaults from the Date class itself rather than duplicating that info in the log writers.. 

`format_entry()` is inside the log writer itself because It's not really common to every log writer, eg syslog doesn't use it. Either way, I'm still thinking how best to clean it up..

shadowhand:
@kiall Just because one adapter does not use a piece of code does not justify duplicating it in every other adapter. As for the Date formatting, **do not** use the Date defaults. The duplication was for a very specific reason, namely: I might set the Date defaults based on the user's locale, which would cause log entries to be inconsistently formatted. Dig into the commit history and you will see what I mean.

kiall:
@shadowhand - how about this latest version?

I've defaulted to the Date options now, which mimics the old behavior, but have allowed for the options to be changed.

kiall:
Also worth noting that the default log format has changed slightly.

From "time --- level: body" to "time --- level: body in file:line"

shadowhand:
That looks better.

classes/kohana/log.php
```diff
@@ -130,7 +119,7 @@ public function detach(Log_Writer $writer)
 	 * @param   array   $values     values to replace in the message
 	 * @return  Log
 	 */
-	public function add($level, $message, array $values = NULL)
+	public function add($level, $message, array $values = NULL, array $additional = NULL)
```
cbandy, line 122: Needs doc comment.

shadowhand, line 122: Thanks @cbandy, that is an excellent point. @kiall, does CS catch missing `@param` comments?

classes/kohana/log/writer.php
```diff
@@ -48,4 +61,30 @@
...
+
+		$string = strtr($format, $message);
+
+		if (isset($message['additional']['exception']))
+		{
+			// Re-use as much as possible, just resetting the body to the trace
+			$message['body'] = $message['additional']['exception']->getTraceAsString();
+			$message['level'] = 'STRACE';
+
+			$string .= PHP_EOL.strtr($format, $message);
```
cbandy, line 84: Is this an intentional change in behavior? The old code combined the error and trace using `"\n--\n"`.

kiall, line 84: Nope - well spotted.

But - Personally, I think the change is okay..  it was redundant information anyway!

shadowhand, line 84: For parsing, I think the error and trace need to be separated in some way.

PR #210, opened 2012-01-25
Fix for $request->redirect() bug when Kohana:$index_file is activated

buremba:
There was a problem in $request->redirect() when Kohana:$index_file is activated. Kohana:$index_file must be a string value or FALSE not TRUE.

### DIFF
--- a/classes/kohana/request.php
+++ b/classes/kohana/request.php
@@ -242,7 +242,8 @@ class Kohana_Request implements HTTP_Request {
 
 	/**
 	 * Automatically detects the URI of the main request using PATH_INFO,
-	 * REQUEST_URI, PHP_SELF or REDIRECT_URL.
+	 * REQUEST_URI, PHP_SELF or 
+	 _URL.
 	 *
 	 *     $uri = Request::detect_uri();
 	 *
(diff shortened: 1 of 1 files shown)


classes/kohana/request.php
```diff
@@ -936,13 +937,13 @@ public function redirect($url = '', $code = 302)
 
 		if (strpos($referrer, '://') === FALSE)
 		{
-			$referrer = URL::site($referrer, TRUE, Kohana::$index_file);
+			$referrer = URL::site($referrer, TRUE, TRUE);
```
MazeXD, line 940: I recommend this snippet cause it's valid for both enabled and disabled index_file.

$referrer = URL::site($referrer, TRUE, !empty(Kohana::$index_file));

classes/kohana/request.php
```diff
@@ -936,13 +937,13 @@ public function redirect($url = '', $code = 302)
...
 		}
 
 		if (strpos($url, '://') === FALSE)
 		{
 			// Make the URI into a URL
-			$url = URL::site($url, TRUE, Kohana::$index_file);
+			$url = URL::site($url, TRUE, TRUE);
```
MazeXD, line 946: I recommend this snippet cause it's valid for both enabled and disabled index_file.

$url = URL::site($url, TRUE, !empty(Kohana::$index_file));

PR #213, opened 2012-02-08
Fixes #4434 - Request::$initial and Request::$current changed to protected properties.

ghost:
#4434 - Request::$initial and Request::$current should be protected properties and only accessed via their corresponding methods Request::initial() and Request::current().

### DIFF
--- a/classes/kohana/view.php
+++ b/classes/kohana/view.php
@@ -49,7 +49,7 @@ class Kohana_View {
 		if (View::$_global_data)
 		{
 			// Import the global view variables to local namespace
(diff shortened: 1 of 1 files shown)


kemo:
This doesn't follow Kohanas conventions

radsystems:
In what way?

kemo:
Protected properties must begin with an underscore

radsystems:
I've closed this pull request as it is not a sufficient fix, since Request::$initial and Request::$current is used throughout the system core. Therefore renaming them or changing their scope will require more refactoring than was provided here. This is probably best left for a core developer to address if they decide it is a valid issue.

PR #215, opened 2012-02-10
Issue #4353 - Request::execute() should be able to follow Location headers

kiall:
See: http://dev.kohanaframework.org/issues/4353


kiall:
!build

We had some failing tests in the core code that triggered this failure.

PR #216, opened 2012-02-10
3.2/master

buremba:


shadowhand:
Multiple issues here:
1. This is targeting the wrong branch. It should be targeting `3.2/develop` or `3.3/develop`.
2. Please make a ticket reference when changing code. If no ticket has been created, make one.
3. I doubt this is actually valid.

buremba:
@shadowhand  the third parameter of URL::site expects a boolean value because it calls URL::base with that value (https://github.com/buremba/core/blob/3.2/master/classes/kohana/url.php#L112).
And URL::base puts index prefix when $index===TRUE (https://github.com/buremba/core/blob/3.2/master/classes/kohana/url.php#L60)

When you use URL::site($referrer, TRUE, Kohana::$index_file), because Kohana::$index_file is a string value (take a look at this line: https://github.com/buremba/core/blob/3.2/master/classes/kohana/url.php#L60) $index===TRUE returns FALSE because of "===" operator.

shadowhand:
@buremba Ah, thank you for the explanation and you are 100% correct. Please create a ticket and a new pull request that targets the correct branch.

buremba:
@shadowhand I realized that @BRMatt already commited a fixed version of request.php to develop branch. (https://github.com/kohana/core/commit/b7869b16248fd7ed24acd63a9604cd5d6a6bfeec) Because I worked with master branch, I couldn't realized his commit. I'm newbie on Github, sorry. :)
However I think you should update master branch for this bug. Otherwise redirect() doesn't work if Kohana::$index_file is activated.

PR #217, opened 2012-02-13
Add max_numeric and min_numeric to Valid class.

digitaljohn:
For example. If you have a field that must be greater than 10 but has no maximum number.

http://dev.kohanaframework.org/issues/4446

### DIFF
--- a/classes/kohana/valid.php
+++ b/classes/kohana/valid.php
@@ -478,6 +478,30 @@ class Kohana_Valid {
 	{
 		return ($number >= $min AND $number <= $max);
 	}
+	
+	/**
+	 * Tests if a number is small enough.
+	 *
+	 * @param   string   number to check
+	 * @param   integer  maximum value
+	 * @return  boolean
+	 */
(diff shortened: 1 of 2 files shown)


shadowhand:
First, this is targeting the wrong branch. All new features must be targeted at the current "develop" branch, (currently `3.3/develop`).  The Kohana bot will be along shortly to close this pull request for that reason.

Second, please do not include merge commits in pull requests. New features should be developed in separate branches to prevent merges from cluttering up pull requests.

Third, is there any value in this compared to using the `range` rule?

digitaljohn:
Doh... ill re-branch...

How can I accomplish the same using the range rule?

E.g.1 ) I want a field to be validated so it is greater than 1... but has no maximum value set.
E.g.2 ) I want a field to be validated so it is less than 100... but has no minimum value set.

shadowhand:
It probably makes sense to rename these rules `less_than` and `greater_than`, only because it is slightly more accurate terminology. But it could be argued that is purely semantics.

PR #218, opened 2012-02-14
3.3/develop

thinkclay:
$end_char was already defaulting to null, so the check of null just to set to an ellipses was unnecessary duplication of logic. Furthermore, the character in the file was an actual single U+2026 character which would cause errors in some applications and IDE's depending on UTF support and font (see http://www.fileformat.info/info/unicode/char/2026/fontsupport.htm) so I've changed it to (3) period characters instead to comply with UTF-8 and fonts.

### DIFF
--- a/classes/kohana/text.php
+++ b/classes/kohana/text.php
@@ -87,10 +87,8 @@ class Kohana_Text {
 	 * @return  string
 	 * @uses    UTF8::strlen
 	 */
-	public static function limit_chars($str, $limit = 100, $end_char = NULL, $preserve_words = FALSE)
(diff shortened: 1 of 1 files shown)


shadowhand:
This doesn't follow the usual Kohana conventions. Usually we default to `NULL` for optional parameters, then apply a default inside the function. This makes it possible to do: `echo limit_chars($text, 50, NULL, TRUE)` without having to worry about defining the end character.

However, I _do_ think we should look at the history of this method and see why the ellipsis character is was chosen instead of three periods. IIRC, Geert was the one responsible for this method.

PR #221, opened 2012-02-17
3.3/develop

thinkclay:
Following along with the pull request I closed here (https://github.com/kohana/core/pull/218#issuecomment-3953700) I've researched a bit more, and found that the ellipsis character does indeed cause issues with many IDE's and editors depending on font support for the character. This was clearly not an intentional use of the ellipsis character (you can look at TextTest.php and see the inconsistency in the output) but most likely due to the like of spell checkers in the editor or OS (Coda for Mac will automatically convert (3) periods into an ellipsis for example).

### DIFF
--- a/classes/Kohana/Text.php
+++ b/classes/Kohana/Text.php
@@ -60,7 +60,7 @@ class Kohana_Text {
 	public static function limit_words($str, $limit = 100, $end_char = NULL)
 	{
 		$limit = (int) $limit;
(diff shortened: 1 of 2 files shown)


shadowhand:
Did you create a Redmine ticket for this? If not, please create one. Pull requests cannot be merged without ticket references. Thanks!

thinkclay:
Ticket created http://dev.kohanaframework.org/issues/4450

PR #223, opened 2012-02-24
Fixes for #4353 - strict_redirect option and tests

acoulton:
Per the ticket at http://dev.kohanaframework.org/issues/4353 I've implemented the new strict_redirect option as discussed. I've also added tests and docs and resolved a couple of errors in the setters implemented in previous commits.

### DIFF
--- a/classes/Kohana/Request/Client.php
+++ b/classes/Kohana/Request/Client.php
@@ -76,13 +76,13 @@ abstract class Kohana_Request_Client {
 	{
 		$response = Response::factory();
 
-		if ($this->_cache instanceof HTTP_Cache)
-			return $this->_cache->execute($this, $request, $response);
+		if (($cache = $this->cache()) instanceof HTTP_Cache)
+			return $cache->execute($this, $request, $response);
 
 		$response = $this->execute_request($request, $response);
 
 		// Do we need to follow a Location header ?
-		if ($this->_follow AND in_array($response->status(), array(201, 301, 302, 303, 307))
+		if ($this->follow() AND in_array($response->status(), array(201, 301, 302, 303, 307))
 			AND $response->headers('Location'))
 		{
 			// Figure out which method to use for the follow request
@@ -99,7 +99,7 @@ abstract class Kohana_Request_Client {
 					break;
 				case 302:
 					// Cater for sites with broken HTTP redirect implementations
-					if ($this->_strict_redirect)
+					if ($this->strict_redirect())
 					{
 						$follow_method = $request->method();
 					}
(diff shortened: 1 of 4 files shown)


acoulton:
@shadowhand that's the property accessors implemented as requested. There's still two open points on the ticket that I had asked about and haven't had a reply:
- Currently the original request body is not resubmitted on a POST/PUT redirect - should it be?
- @samsoir and @kiall had discussed providing some of the follow logic with callbacks for even more flexibility (I can make that a separate ticket if you like, as it's perhaps more of a new feature than a completion of this?)

samsoir:
@acoulton the redirect follow lambda was meant to be part of this ticket, @kiall was working on the implementation. We should await his feedback before merging this feature.

kiall:
Heya - Sorry, real life has kept me busy recently :)

I'll re-read everything I've missed here and in the ticket tonight or tomorrow..

acoulton:
@samsoir, @kiall - no problem, just let me know if I can do any more to help

kiall:
Will do! (See below :P)

I've not had any time to work on the implementation, but unless something was discussed here/the ticket that I'm not aware of, the sample syntax I [proposed](http://dev.kohanaframework.org/issues/4353#note-13) would still be my preference - but it still needs some work!

The only 2 issues I'm aware of are:

A) the requirement that in certain cases, the user agent must be told explicitly to follow the request.

This can probably be solved by supplying the callback, but requiring the user enabled it in `header_callbacks` (By passing it in..). Another issue here - is where this callback should live.. 

I think this is an acceptable solution. But would be open to other suggestions!

B) No params can be passed to the implementation, meaning no easy way to emulate the browser behavior.

This can probably be solved by providing 2 callbacks (One main implementation, with 3 params, and 1 alias to change the 3rd param into a TRUE/FALSE - whatever makes sense). Maybe we should also allow (or just use) a tuple?

Finally - I think I would like the callbacks to be able to return either a `Request`, or `Response` object.

If a `Request` is returned, it should be executed and and it's response passed to the next callback.
If a `Response` is returned, it should just be passed to the next callback as usual.

(Note: I'm suggesting the original `Request` object is always supplied to every callback! This is mainly for consistency reasons, If a CB returns a new `Response` - we don't have access to the `Request` that generated it..)

I'd love to see an implementation of this if you have time! I'm really overstretched at the moment :)

kiall:
Also - I'm happier with your current implementation than I am with my own, merging it in now in case we don't get the CB stuff done ASAP.. Lets continue this conversation in the ticket!

PR #225, opened 2012-02-26
3.3/develop

jtreminio:
Exception should not exit.

### DIFF
--- a/classes/Kohana/Kohana/Exception.php
+++ b/classes/Kohana/Kohana/Exception.php
@@ -88,7 +88,7 @@ class Kohana_Kohana_Exception extends Exception {
 		// Send the response to the browser
 		echo $response->send_headers()->body();
 
-		exit(1);
+		return TRUE;
 	}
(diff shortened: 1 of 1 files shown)


kiall:
Actually, this exit is correct.

The only place this can get called is if an execption occours somewhere between index.php and and the first use of Request_Client_Internal, so exit is the appropriate response as there can't be a subrequest (or even request) to interrupt.

kiall:
This has been fixed by:

https://github.com/kohana/core/commit/c2fed8b19aaad4b6e6c3121c760e581803c67a4d
and
https://github.com/kohana/userguide/commit/f6569199e7e49af8665e6d93a221a97cb3a49b68

But - I believe both cases should be removed from Kohana.

The Markdown one feels completely wrong, and looks like it was added specifically for this "example".

and View::__toString() should be either removed, or allowed to throw the exception and cause a confusing error rather than behaving differently to View::render();

PR #226, opened 2012-02-28
Correct the param label from param2 to param3 for rule "matches"Correct the param label from param2 to param3 

si1en2i0:
EXAMPLE:

rule('password_confirm', 'matches', array(':validation', ':field', 'password')

if not. When the rule is failed it will cause error "password confirm must be the same as password confirm"


shadowhand:
Please modify the unit tests that are failing.

zeelot:
I think this is functional in the latest 3.3 version, let me know if I misunderstood the issue.

PR #227, opened 2012-03-01
Inflector updates

hkdobrev:
Inflector::plural should not pluralize plural form of an irregular verb.

Inflector::underscore should replace not only white-space with a underscore, but some other characters as well.

### DIFF
--- a/classes/Kohana/Inflector.php
+++ b/classes/Kohana/Inflector.php
@@ -183,6 +183,10 @@ class Kohana_Inflector {
 		{
 			$str = Inflector::$irregular[$str];
 		}
+		elseif (in_array($str, Inflector::$irregular))
+		{
+			// Do nothing
+		}
 		elseif (preg_match('/[sxz]$/', $str) OR preg_match('/[^aeioudgkprt]h$/', $str))
 		{
 			$str .= 'es';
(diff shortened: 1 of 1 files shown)


classes/Kohana/Inflector.php
```diff
@@ -249,7 +253,7 @@ public static function decamelize($str, $sep = ' ')
 	 */
 	public static function underscore($str)
 	{
-		return preg_replace('/\s+/', '_', trim($str));
+		return preg_replace('/(\s|-|~|=)+/', '_', trim($str));
```
hkdobrev, line 256: Perhaps some more characters could be added in the future. They could be even set as a param to the underscore method or as a config array.

zeelot:
Do we have a redmine ticket for this? Sorry for the late reply but I'm finally catching up to all the PRs!

zeelot:
I'm ok with the first commit but the delimiters change is a bit arbitrary… I'm going to cherry pick the one commit into develop and we can work on adding a new parameter to the function for the list of characters to turn into underscores in the next major release (API change).

Thanks for the PR and sorry for taking so long to get to it (trying not to let it happen again).

zeelot:
Taken care of here: https://github.com/kohana/core/pull/385

PR #228, opened 2012-03-02
3.2/develop

kemo:
http://dev.kohanaframework.org/issues/4414
http://dev.kohanaframework.org/issues/4465


shadowhand:
Merged into `3.3/develop` as well.

PR #229, opened 2012-03-05
Header callback implementation for Request_Client - completes #4353

acoulton:
@kiall, I think I've implemented the remaining work on http://dev.kohanaframework.org/issues/4353 - see the update there for the details of what I've done.

The code is unit tested and passes phpcs but there may be remaining formatting issues around the nested arrays particularly. It's a bit complex in places, but I think this is a slightly complex requirement to implement.

Comments very welcome!

### DIFF
--- a/classes/Kohana/Request/Client.php
+++ b/classes/Kohana/Request/Client.php
@@ -1,4 +1,4 @@
-<?php defined('SYSPATH') or die('No direct script access.');
+<?php defined('SYSPATH') OR die('No direct script access.');
 /**
  * Request Client. Processes a [Request] and handles [HTTP_Caching] if
  * available. Will usually return a [Response] object as a result of the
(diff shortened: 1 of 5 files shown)


kiall:
A quick read of the code, and it looks good.

I'll have a chance to test this later this evening - cmon 3.3! ;)

acoulton:
@kemo I'd need to look again and execute the relevant test, but I think that's correct - the check is executed as part of the next request execution (eg after `$this->callback_depth()` has been incremented and before the next recursion actually triggers a request).

Although thinking about it, it might be safer to use >=, in case the callback_depth parameter is incremented by more than one for any reason.

acoulton:
@kemo sorry, I hadn't spotted earlier that I was using > rather than === already. I think this is safer - if a callback either accidentally increments by more than one, or executes a couple of additional requests before returning the exception will still be thrown when control eventually returns to that line.

I have however updated the test case to specifically verify the number of requests executed before the exception (rather than just checking the exception is thrown eventually). I have also changed callback_depth to start at 1 for the original request rather than 0 which I think makes more sense (so following a traditional request/redirect pair it will be 2, etc).

kemo:
@acoulton  looks great, keep up with good work!

kiall:
I **will** get to this today! I will. (Tell myself that enough times and its bound to happen, right?)

Anyway - I'm pretty sure I'm happy with it, and will merge later this afternoon unless I uncover something unexpected when fully testing it out..

kiall:
Today turned into next week, but still. Done!

Looks good @acoulton - Merged, Thanks!

acoulton:
Great! Thanks @kiall - roll on 3.3! :)

PR #231, opened 2012-03-14
3.2/master

dospole:
nginx rewrite query args;
no ob_start when CLI mode;


shadowhand:
Please split these two commits.

PR #238, opened 2012-04-11
Kohana environment values are illogical

ThePixelDeveloper:
http://dev.kohanaframework.org/issues/4494

Right now the following statement is false:

```
Kohana::DEVELOPMENT < Kohana::PRODUCTION
```

The values should be switched so they make logical sense when being used in an integer comparison.

The change cleans up the following code:

```
if (Kohana::$environment === Kohana::DEVELOPMENT OR
    Kohana::$environment === Kohana::TESTING)
```

to

```
if (Kohana::$environment < Kohana::STAGING)
```

### DIFF
--- a/classes/Kohana/Core.php
+++ b/classes/Kohana/Core.php
@@ -20,10 +20,10 @@ class Kohana_Core {
 	const CODENAME = 'Kolibri';
 
 	// Common environment type constants for consistency and convenience
-	const PRODUCTION  = 10;
-	const STAGING     = 20;
-	const TESTING     = 30;
-	const DEVELOPMENT = 40;
+	const PRODUCTION  = 40;
+	const STAGING     = 30;
+	const TESTING     = 20;
+	const DEVELOPMENT = 10;
 
 	// Security check that is added to all generated PHP files
 	const FILE_SECURITY = '<?php defined(\'SYSPATH\') OR die(\'No direct script access.\');';


shadowhand:
This is just change for the sake of change. -1

ThePixelDeveloper:
There's a logical and thought out reason for it.

shadowhand:
Yup, your reason is "I think this thing that goes left should go right". There's no reason for this.

lukemorton:
+1

lukemorton:
Environments are progressive. Development leads to testing, testing to staging and staging to production. Currently their values indicate that production leads to development.

yakatz:
I think this is reasonable. I have actually wanted to be able to do something like 

``` php
if (Kohana::$environment < Kohana::STAGING)
```

zeelot:
-1, to me, production as `0` or `1` would make most sense, as you can't get any closer to live than production. Everything else is moving away from the live state and adding environments doesn't mean possibly having to change the production value. However, like @shadowhand mentioned, even that would be a change for the sake of change so I don't think this matters enough to be modified.

My point was mostly that you can think of it in a few different ways ;)

zeelot:
Closing for now. Also want to point out that putting these things into a config and loading different configs based on environment will remove the need to make the comparisons all over your code. There are a few ways to do this and my preferred one is to simply load a config group based on the environment so I can replace any config item on a specific environment. This also works much better in `3.3` since you can replace a single item without rewriting the entire config file.

PR #244, opened 2012-04-24
Docs edit: Bootstrap & Upgrading core/system guides

randomecho:
- Tense and sentence structure about what the bootstrap is doing 
- Wanting ORM to be capitalised as it is an abbreviation
- ETag not in a possessive use, so should drop the apostrophe


shadowhand:
Some of these whitespace changes may not be correct. Have you verified that the HTML output is the same?

randomecho:
Compared on a local version of docs for the copy edits, which appears to look the same for text content. 
Could have been editor faffing up the encoding which ends up with these whitespace issues.

PR #245, opened 2012-04-26
Normalize the Accept-Language header so it matches on upper-cased country codes (en-US)

jerfowler:
Normalize the Accept-Language header so it matches on upper-cased country codes (en-US). This happens in Internet Explorer (9), Safari on Windows. 

Fixes: http://dev.kohanaframework.org/issues/4502

Modified tests to check for this scenario.


shadowhand:
Thanks for including a unit test!

shadowhand:
Merged into 3.3 in 1cd79e31b8e6ae7f2a05e51f398da953d864ef0a.

PR #246, opened 2012-04-27
Arr::get docs fix

hkdobrev:
The previous version of the PHPDoc for the `Arr::get` method was not accurate. It said you are retrieving a key, when you are actually retrieving a value for a key.


zombor:
O_o

Please follow the development guidelines: https://github.com/kohana/kohana/wiki/developers

You need to open a redmine issue and do your pull request properly.

PR #247, opened 2012-04-28
Fixes #4505 - Log::STRACE doesn't exist.

ThePixelDeveloper:
Kohana issue: http://dev.kohanaframework.org/issues/4505


kiall:
Its the reference to STRACE that should be removed, the stack trace is no longer logged as a seperate message..

zeelot:
Looks to be the same as https://github.com/kohana/core/pull/266

Sorry, saw his first!

PR #250, opened 2012-05-06
$_cache is protected field. ...

GendelfLugansk:
...So, using of I18n::$_cache isn't correct. self::$_cache is better

### DIFF
--- a/classes/kohana/i18n.php
+++ b/classes/kohana/i18n.php
@@ -95,9 +95,9 @@ class Kohana_I18n {
 	 */
 	public static function load($lang)
 	{
-		if (isset(I18n::$_cache[$lang]))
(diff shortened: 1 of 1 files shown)


isaiahdw:
Actually in this case using the class name instead of self is correct. Self refers to current class, not the extended class, so due to the way Kohana does transparents extensions using the class name is the correct way. In php 5.3+ static:: could be used instead of the class name.

cs278:
PHP >= 5.3 actually.

isaiahdw:
Yes, that was a typo. Changing the class name to static:: could be applied to the 3.3/develop branch, but we still support php 5.2.x in 3.2 so it can't be changed here.

PR #252, opened 2012-05-09
Passes CodeSniffer with  exceptions:

yakatz:
- I was not able to modify UTF-8 file: "expected OR but found or" core/utf8/transliterate_to_ascii.php:1
- "Function name "__" is invalid; only PHP magic functions should be prefixed with a double underscore" core/classes/Kohana/I18n.php:155
- "Method name "exchangeArray" is not in all lowercase using underscores for word separators" core/classes/Kohana/HTTP/Header.php:417
- "Whitespace before a closing parenthesis is not allowed" core/classes/Kohana/Arr.php:443
- "Whitespace before a closing parenthesis is not allowed" core/classes/Kohana/Arr.php:475

### DIFF
--- a/classes/Arr.php
+++ b/classes/Arr.php
@@ -1,3 +1,3 @@
-<?php defined('SYSPATH') or die('No direct script access.');
+<?php defined('SYSPATH') OR die('No direct script access.');
 
 class Arr extends Kohana_Arr {}
--- a/classes/Config.php
+++ b/classes/Config.php
@@ -1,3 +1,3 @@
-<?php defined('SYSPATH') or die('No direct script access.');
+<?php defined('SYSPATH') OR die('No direct script access.');
 
 class Config extends Kohana_Config {}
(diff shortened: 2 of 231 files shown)


evanpurkhiser:
:+1:

I have always wondered about the security code `<?php defined('SYSPATH') or die('No direct script access.');` using the lowercase `OR` operator.

kiall:
Thanks :)

I think Ive merged all your pulls now..

evanpurkhiser:
Would it be a good idea to add in PHP Code Sniffer ignore tags on the exceptions to the standards?

```
// @codingStandardsIngoreStart
...
// @codingStandardsIgnoreEnd
```

More information on using these tags can be found [on the PHPCS manual](http://pear.php.net/manual/en/package.php.php-codesniffer.advanced-usage.php). (Under the 'Ignoring Parts of a File' header)

PR #253, opened 2012-05-09
Fix linkrot on intro page, remove other 301 redirects

randomecho:
Links on the index are all pointing to URLs that have moved either out there in the Internet or internally to the guide itself. 

This fixes up those linkrot issues as well as point to the canonical location on several Wikipedia articles instead of their redirects.

Also 
- "Stack Overflow" the site has a space between the words
- "Git" should be capitalised for the application.


shadowhand:
Thank you!

PR #261, opened 2012-05-16
A better documentation for Encryption userguide.

diogotozzi:
A better documentation for Encryption userguide.

### DIFF
--- a/guide/kohana/security/encryption.md
+++ b/guide/kohana/security/encryption.md
@@ -1,107 +1,86 @@
 # Encryption
 
-Kohana supports built-in encryption and decryption via the [Encrypt] class, which is a convenient wrapper for the [Mcrypt library](http://www.php.net/mcrypt).
+The [Encryption] class performs a very easy-to-use way to encode and decode strings.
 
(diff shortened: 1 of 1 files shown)


guide/kohana/security/encryption.md
```diff
@@ -1,107 +1,86 @@
 # Encryption
 
-Kohana supports built-in encryption and decryption via the [Encrypt] class, which is a convenient wrapper for the [Mcrypt library](http://www.php.net/mcrypt).
+The [Encryption] class performs a very easy-to-use way to encode and decode strings.
```
zeelot, line 3: `performs` doesn't make sense in this sentence, it would make more sense to use `provides`. My other issue with this is that we aren't linking to the PHP extension. What was your reasoning for that?

guide/kohana/security/encryption.md
```diff
@@ -1,107 +1,86 @@
...
 
-To use the class, first start by ensuring you have the Mcrypt extension loaded to your PHP config. See the [Mcrypt Installation page](http://www.php.net/manual/en/mcrypt.installation.php) on php.net. The Mcrypt extension requires [libmcrypt](http://sourceforge.net/projects/mcrypt/files/).
+[!!] You have to enable Mcrypt extension in your PHP installation.
```
zeelot, line 5: You are missing a `the` in front of `Mcrypt` and maybe we can use this notice to link to the PHP page:

> [!!] You have to enable the [Mcrypt extension](http://www.php.net/mcrypt) before using this class.

zeelot:
Started leaving comments about things to fix but I think it's better to discuss how to improve this page with smaller Pull Requests so I sent you an email instead.

PR #263, opened 2012-05-17
Fix for bug #4472

AlexC:


zeelot:
unnecessary since I'm about to merge https://github.com/kohana/core/pull/262 which contains the same commit

PR #264, opened 2012-05-17
Fixed bug #4240

AlexC:
This also effects 3.3/develop and will need merging into that branch too

### DIFF
--- a/classes/kohana/url.php
+++ b/classes/kohana/url.php
@@ -147,7 +147,7 @@ class Kohana_URL {
 			else
 			{
 				// Merge the current and new parameters
-				$params = array_merge($_GET, $params);
+				$params = Arr::merge($_GET, $params);
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


zeelot:
Could you write some tests for this? Seems like we have `URL::query()` tests already so adding a recursive array to `provider_query()` should do the trick.

stillman:
The patch is incorrect!

``` php
<?php
$_GET['a'] = 'a';
echo urldecode(URL::query(array('a' => 'b'))); // Produces ?a[0]=a&a[1]=b instead of ?a=b
```

zeelot:
I guess it should use `Arr::merge()` instead? Anyway, that's why we should write some tests ;)

AlexC:
@stillman Shouldn't your example return '?a=b' instead?

I've just forced-pushed to my branch which changes to Arr::merge() and adds some unit test data.

stillman:
@AlexC Yes, sure! Thanks

zeelot:
Perfect, many thanks! :)

PR #265, opened 2012-05-17
Fixed bug #4208

AlexC:
This also effects 3.3/develop

### DIFF
--- a/classes/kohana/text.php
+++ b/classes/kohana/text.php
@@ -365,7 +365,7 @@ class Kohana_Text {
 		$text = preg_replace_callback('~\b(?<!href="|">)(?:ht|f)tps?://[^<\s]+(?:/|\b)~i', 'Text::_auto_link_urls_callback1', $text);
 
 		// Find and replace all naked www.links.com (without http://)
-		return preg_replace_callback('~\b(?<!://|">)www(?:\.[a-z0-9][-a-z0-9]*+)+\.[a-z]{2,6}\b~i', 'Text::_auto_link_urls_callback2', $text);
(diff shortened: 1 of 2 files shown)


GeertDD:
This change introduces a new bug as demonstrated below.

```
Text::auto_link_urls('www.example.be'); // Fails
Text::auto_link_urls('www.example.com'); // Succeeds
```

The problem is that in this part of the regex `\.[a-z]{2,6}[^<\s]+`, you are _requiring_ a character to follow the top-level domain extension. In the examples above, no characters follow after it, so the regex has to backtrack. The backtracking works only as long as the top-level domain extension consist of at least three characters.

Here's a quick fix (I didn't test it very thoroughly):
`'~\b(?<!://|">)www(?:\.[a-z0-9][-a-z0-9]*+)+\.[a-z]{2,6}+\b(?:[^<\s]*(?:/|\b))?~i'`

AlexC:
Thanks, good catch. Looking further into the regex, I can't see why the ending of this regex is as it is (I didn't make the original, just amended for this bug). The following regex works just fine for me under many examples that I tried. Do you see any issues with this one?

```
~\b(?<!://|">)www(?:\.[a-z0-9][-a-z0-9]*+)+\.[a-z]{2,6}[^<\s]*~i
```

GeertDD:
The ending is important for dealing with trailing punctuation. Your last suggestion matches the commas as part of the URL too. It's tricky stuff. The more tests the better.

```
Text::auto_link_urls('www.example.be,');
Text::auto_link_urls('www.example.be/');
Text::auto_link_urls('www.example.be/,');
```

AlexC:
The original bug #4208 should now be fixed with the change I just made. I've also included some more unit tests for this function to ensure its behaviour.

The issue with trailing punctuation is another bug, mostly (only?) to do with trailing slashes now. Commas work as expected (they do not become part of the URL). I'll report this as another bug and see what I can do.

zeelot:
@GeertDD are you fine with this now? Just trying to merge in any final fixes before the release :)

PR #267, opened 2012-05-18
Strace level to debug

AlexC:


AlexC:
See Kohana bug #4519

zeelot:
@kiall leaving this one for you, I guess, since you're the one that made most of this (I think). :)

PR #268, opened 2012-05-24

yakatz:
This should be targeted to 3.4/develop, but that does not exist yet.


yakatz:
Found some complications. Closing for now.

PR #269, opened 2012-05-24

yakatz:
This should be targeted to 3.4/develop, but that does not exist yet.

(This is working in production for me.)


zeelot:
umm… what happened here?

yakatz:
I managed to corrupt the history in my fork and I recreated it which probably automatically closed the pull request.
I will open a new one.

yakatz:
https://github.com/kohana/core/pull/300

PR #270, opened 2012-05-24
Route::__construct inline documentation update

jessedc:
refs #4525

http://dev.kohanaframework.org/issues/4525


zeelot:
Thanks :)

PR #275, opened 2012-06-12
revamped Feed Helper

chandeeland:
I modified the feed helper to handle ATOM format as well as providing the ability to assign nested tags and to handle tag attributes.  The static function create() was preserved but considered deprecated, and should be regarded as an example of the new usage.

### DIFF
--- a/classes/kohana/feed.php
+++ b/classes/kohana/feed.php
@@ -1,4 +1,4 @@
-<?php //defined('SYSPATH') or die('No direct script access.');
+<?php defined('SYSPATH') or die('No direct script access.');
 /**
  * RSS and Atom feed helper.
  *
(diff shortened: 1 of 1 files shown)


classes/kohana/feed.php
```diff
@@ -1,15 +1,184 @@
-<?php defined('SYSPATH') or die('No direct script access.');
+<?php //defined('SYSPATH') or die('No direct script access.');
```
randomecho: Is there a reason why this would be commented out?

classes/kohana/feed.php
```diff
@@ -1,15 +1,184 @@
...
+	{
+		if (!empty($this->current_entry))
+		{
+			$this->entry[] = $this->current_entry;
+			$this->current_entry = array();
+		}
+	}
+
+	/**
+	 * ouput XML from data collected
```
randomecho: Typo and doesn't follow the convention of starting your comment with a capital letter

classes/kohana/feed.php
```diff
@@ -1,15 +1,184 @@
...
+		}
+
+		// DOM is preferred, as it generates more readable XML
+		if (function_exists('dom_import_simplexml'))
+		{
+			$feed = dom_import_simplexml($feed)->ownerDocument;
+			$feed->formatOutput = TRUE;
+			return $feed->saveXML();
+		}
+
```
randomecho: Missing else for when `dom_import_simplexml` doesn't exist

classes/kohana/feed.php
```diff
@@ -70,107 +239,38 @@ public static function parse($feed, $limit = 0)
...
-		$feed = simplexml_load_string($feed);
+        $info += array('title' => 'Generated Feed', 'link' => '', 'generator' => 'KohanaPHP');
 
-		foreach ($info as $name => $value)
-		{
-			if ($name === 'image')
-			{
-				// Create an image element
-				$image = $feed->channel->addChild('image');
+		$feed = new self();
```
evanpurkhiser: Using the PHP `self` keyword is discouraged in Kohana due to the Cascading Filesystem.

Because the `self` keyword always refers to the class it was written in (in this case `Kohana_Feed`) that class will always be instantiated, NOT the extended version of the class (which is just `Feed`). So if you wanted to extend the Feed class to add some functionality, you wouldn't be able to use the `Feed::create()` method, as it wouldn't create an instance of your extended class.

classes/kohana/feed.php
chandeeland: no reason. i was running a test, its a typo.

classes/kohana/feed.php
```diff
@@ -1,15 +1,184 @@
-<?php defined('SYSPATH') or die('No direct script access.');
+<?php //defined('SYSPATH') or die('No direct script access.');
 /**
  * RSS and Atom feed helper.
  *
  * @package    Kohana
  * @category   Helpers
- * @author     Kohana Team
```
chandeeland, line 7: you guys can put the attribution back if you want, I didn't want to presume...

classes/kohana/feed.php
chandeeland: the else is unnecessary due to the use of the RETURN

classes/kohana/feed.php
chandeeland: thanks!  i changed it to use Feed:: instead of self::  please let me know if there's anything else.

enov:
Thanks @chandeeland. Could you rebase this on 3.4/develop? Any chance that you could also provide unit tests?

PR #276, opened 2012-06-26
Guide/custom error pages: Ensure an initial request object exists

leth:
:)


leth:
http://dev.kohanaframework.org/issues/4561

zeelot:
The documentation for this seems very different in `3.3/develop`, closing but let me know if there are still issues in those pages! Sorry for taking so long to reply…

PR #277, opened 2012-07-19
Created Text::twitter() helper

robbytaylor:
Adds HTML links to URLs, Twitter screen names, and hashtags to text. This is useful when taking text from the twitter API, for instance when embedding a twitter feed into a web page.

### DIFF
--- a/classes/kohana/text.php
+++ b/classes/kohana/text.php
@@ -603,4 +603,29 @@ class Kohana_Text {
 		return $str;
 	}
 
+	/**
+	 * Parses text representing tweet (from twitter) and adds links to URLs, twitter screen names, and hashtags.
+	 * Useful after getting tweets from the twitter API and adding links in the text.
+	 *
+	 *     echo Text::twitter($text);
+	 *
+	 * @uses Text::auto_link()
+	 * @param string $text Tweet text.
+	 * @return string
+	 */
+	public static function twitter( $text )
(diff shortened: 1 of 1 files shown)


shadowhand:
I would prefer this was made more generic.

Also, the formatting does not conform to our code style.

zombor:
I don't think we should have such a specific vendor helper in the core code base.

shadowhand:
Yes, a more generic helper that would define matching regex and the replacements would be more desirable.

robbytaylor:
I appreciate the point about it being vendor specific but embedding a twitter feed into a webpage is a very common task because of the size of Twitter these days. There's a balance perhaps to be met?

How would you suggest making it more generic? Are there any other situations where linking @... or #... would be useful?

zombor:
Facebook is also very large, but we don't have any desire to put any Facebook helpers in our core codebase. This would be better placed in a twitter helper and delivered as a module, imo.

shadowhand:
First, I think a generic helper that `regex => replace` is a useful thing:

```
$text = Text::some_clever_name($text, array('/@[a-z0-9]+/i' => ...));
```

Then I would create a module that adds a `Text::twitter_links($text)` method that proxies to the `some_clever_name` method.

robbytaylor:
An excellent suggestion. I'll do that.

PR #278, opened 2012-07-19
Added Text::plural() helper

robbytaylor:
Text::plural() concatenates a number and the result of Inflector::plural().

This can be useful when, in a view, the count is returned from a calculation and you don't want to dirty your view by assigning the result of the calculation to a variable or do the calculation twice (once to display the number and the again to pass to Inflector::plural()).

### DIFF
--- a/classes/kohana/text.php
+++ b/classes/kohana/text.php
@@ -603,4 +603,22 @@ class Kohana_Text {
 		return $str;
 	}
 
+	/**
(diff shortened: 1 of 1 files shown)


shadowhand:
This is potentially confusing and I don't see a good reason for it. I vote no.

zeelot:
I agree with @shadowhand.

PR #279, opened 2012-07-20
Text::preg_replace_array()

robbytaylor:
Following on from https://github.com/kohana/core/pull/277#issuecomment-7102571

A Text helper to perform an array of replacements on a string.

### DIFF
--- a/classes/kohana/text.php
+++ b/classes/kohana/text.php
@@ -607,8 +607,8 @@ class Kohana_Text {
 	 * Performs an array of regular expression replacements on a string.
 	 *
 	 * @param	string	$text	The text for perform replacements on.
-	 * @param	array	$regex	Array of regular expression replacements in the form $pattern => $replacement
-	 * @param	int		$limit 	The maximum number of replacements for each pattern/replacement in $replacements
+	 * @param	array	$replacements	Array of regular expression replacements in the form $pattern => $replacement
+	 * @param	integer		$limit 	The maximum number of replacements for each pattern/replacement in $replacements
 	 * @return	string
 	 */
 	public static function preg_replace_array($text, array $replacements, $limit = -1)
--- a/classes/kohana/text.php
+++ b/classes/kohana/text.php
@@ -606,9 +606,9 @@ class Kohana_Text {
 	/**
 	 * Performs an array of regular expression replacements on a string.
 	 *
-	 * @param	string	$text	The text for perform replacements on.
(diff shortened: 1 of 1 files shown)


shadowhand:
Not sure I like the name of the method, but the functionality is solid.

robbytaylor:
I don't much like the name either to be honest but I figured it describes what the method does, and I couldn't think of anything which made it clearer.

classes/kohana/text.php
```diff
@@ -603,4 +603,22 @@ public static function widont($str)
 		return $str;
 	}
 
+	/**
+	 * Performs an array of regular expression replacements on a string.
+	 *
+	 * @param	string	$text	The text for perform replacements on.
+	 * @param	array	$regex	Array of regular expression replacements in the form $pattern => $replacement
+	 * @param	int		$limit 	The maximum number of replacements for each pattern/replacement in $replacements
```
zeebinz: Should be: $replacements, integer

classes/kohana/text.php
```diff
@@ -603,4 +603,22 @@ public static function widont($str)
 		return $str;
 	}
 
+	/**
+	 * Performs an array of regular expression replacements on a string.
+	 *
+	 * @param	string	$text	The text for perform replacements on.
+	 * @param	array	$replacements	Array of regular expression replacements in the form $pattern => $replacement
+	 * @param	integer		$limit 	The maximum number of replacements for each pattern/replacement in $replacements
```
evanpurkhiser: The spacing is all wonky here. They should probably be aligned.

classes/kohana/text.php
```diff
@@ -603,4 +603,22 @@ public static function widont($str)
...
+	 * Performs an array of regular expression replacements on a string.
+	 *
+	 * @param	string	$text			The text for perform replacements on.
+	 * @param	array	$replacements	Array of regular expression replacements in the form $pattern => $replacement
+	 * @param	integer	$limit 			The maximum number of replacements for each pattern/replacement in $replacements
+	 * @return	string
+	 */
+	public static function preg_replace_array($text, array $replacements, $limit = -1)
+	{
+		foreach ($replacements as $pattern => $replacement)
```
cs278, line 616: `return preg_replace(array_keys($replacements), $replacements, $text, $limit);`

zeelot:
Could you make a ticket for this please?

zeelot:
What @cs278 said seems to indicate that we don't need a method like this. Let me know if I missed something.

PR #280, opened 2012-07-20
Added Arr::count() helper

robbytaylor:
This fixes issue #4516 (http://dev.kohanaframework.org/issues/4516)

I did the second option for an unexpected parameter, but used a Kohana exception rather than creating an UnexpectedValueException (or similar). I figured an unexpected value exception might be a bit weird in the Kohana filesystem in terms of needing a exception file in a value directory in an unexpected directory. At a loss for a better approach I went with a Kohana exception. Happy to implement suggestions on that count.

### DIFF
--- a/classes/kohana/arr.php
+++ b/classes/kohana/arr.php
@@ -583,15 +583,15 @@ class Kohana_Arr {
 	}
 
 	/**
-	 * Count the number of elements in an array or a countable object.
+	 * Count the number of elements in an array or a Countable object.
 	 *
-	 * @param   mixed   array or instance of countable
-	 * @throws Kohana_Exception
-	 * @return  int
+	 * @param   mixed   array or instance of Countable
+	 * @throws  Kohana_Exception
+	 * @return  integer
 	 */
 	public static function count($array)
 	{
-		if (is_array($array) OR $array instanceof countable)
+		if (is_array($array) OR $array instanceof Countable)
 		{
 			return count($array);
 		}
@@ -608,7 +608,7 @@ class Kohana_Arr {
 			$type = gettype($array);
 		}
 
-		throw new Kohana_Exception("Parameter 1 for Arr::count() must be array or object of type countable, :type given.",
+		throw new Kohana_Exception("Parameter 1 for Arr::count() must be array or object of type Countable, :type given.",
 			array(":type" => $type));
 	}
 
--- a/classes/kohana/arr.php
+++ b/classes/kohana/arr.php
@@ -584,7 +584,7 @@ class Kohana_Arr {
 
 	public static function count($array)
 	{
-		if (is_array($count))
+		if (is_array($array))
 		{
 			return count($array);
 		}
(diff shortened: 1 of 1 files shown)


kemo:
UnexpectedValueException is native to PHP (not that I oppose using Kohana_Exception, quite contrary - I encourage it).

And yes, PHP will internally call countable::count() if countable passed to count(), so that part is ambiguous :)

robbytaylor:
Indeed it is. Can't beat Friday afternoon code :)
Updated to combine the if blocks.

A semi related idea, would it be worth implementing some form of error reporting levels into Kohana? Then funtions like this could throw a Kohana_Exception at a certain reporting level but keep quiet at other times.

classes/kohana/arr.php
```diff
@@ -582,4 +582,34 @@ public static function flatten($array)
 		return $flat;
 	}
 
+	/**
+	 * Count the number of elements in an array or a countable object.
+	 *
+	 * @param   mixed   array or instance of countable
```
zeebinz: Should be: instance of Countable

classes/kohana/arr.php
```diff
@@ -582,4 +582,34 @@ public static function flatten($array)
...
+	 * @throws Kohana_Exception
+	 * @return  int
```
zeebinz: Should be: integer

classes/kohana/arr.php
```diff
@@ -582,4 +582,34 @@ public static function flatten($array)
...
+	 */
+	public static function count($array)
+	{
+		if (is_array($array) OR $array instanceof countable)
```
zeebinz: Should be: instance of Countable

classes/kohana/arr.php
```diff
@@ -582,4 +582,34 @@ public static function flatten($array)
...
+		if (is_object($array))
+		{
+			$type = get_class($array);
+		}
+		else
+		{
+			$type = gettype($array);
+		}
+
+		throw new Kohana_Exception("Parameter 1 for Arr::count() must be array or object of type countable, :type given.",
```
zeebinz: Should be: of type Countable

classes/kohana/arr.php
```diff
@@ -582,4 +582,34 @@ public static function flatten($array)
...
+		{
+			$type = get_class($array);
+		}
+		else
+		{
+			$type = gettype($array);
+		}
+
+		throw new Kohana_Exception("Parameter 1 for Arr::count() must be array or object of type Countable, :type given.",
+			array(":type" => $type));
```
zeelot, line 612: can we simplify this part a bit? Also some small styling fixes:

```
throw new Kohana_Exception('Parameter for Arr::count() must be array or object of type Countable, :type given', array(
    ':type' => (is_object($array)) ? get_class($array) : gettype($array),
))
```

zeelot:
Had someone create a new PR for `3.4/develop` and merged that in, thanks for the code!

PR #281, opened 2012-07-26
Fixed bug #4573

AlexC:


zeelot:
so does this properly format the header when it gets sent?

AlexC:
Yes, there is existing code throughout that handles headers with array values, for example:

https://github.com/kohana/core/blob/3.2/master/classes/kohana/http/header.php#L320
https://github.com/kohana/core/blob/3.2/master/classes/kohana/http/header.php#L860

However because of the casting to string these functions never actually get the array.

PR #282, opened 2012-07-30
nginx + apache https

iswitch:
Fix for nginx + apache


cs278:
Basically the same as #304

zeelot:
Left a reason for not wanting this in #304

PR #283, opened 2012-08-05
Use existing curl_opt equest methods instead of custom request

Ikke:
For POST and PUT, use the existing curlopt options instead of
custom_request.

Fixes [#4515](http://dev.kohanaframework.org/issues/4515)

### DIFF
--- a/classes/kohana/request/client/curl.php
+++ b/classes/kohana/request/client/curl.php
@@ -24,8 +24,10 @@ class Kohana_Request_Client_Curl extends Request_Client_External {
 		// Response headers
 		$response_headers = array();
 
+		$options = array();
+
 		// Set the request method
-		$options[CURLOPT_CUSTOMREQUEST] = $request->method();
+		$options = $this->_set_curl_request_method($request, $options);
 
 		// Set the request body. This is perfectly legal in CURL even
 		// if using a request other than POST. PUT does support this method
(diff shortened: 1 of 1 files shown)


classes/kohana/request/client/curl.php
```diff
@@ -24,8 +24,10 @@ public function _send_message(Request $request)
 		// Response headers
 		$response_headers = array();
 
+		$options = array();
```
Yahasana, line 27: these two lines is unnecessary

classes/kohana/request/client/curl.php
```diff
@@ -107,4 +109,27 @@ public function _send_message(Request $request)
...
 	}
 
+	/**
+	 * Sets the appropriate curl request options. Uses the responding options
+	 * for POST and PUT, uses CURLOPT_CUSTOMREQUEST otherwise
+	 * @param Request $request
+	 * @param array $options
+	 * @return array
+	 */
+	public function _set_curl_request_method(Request $request, array $options)
```
Yahasana, line 119: if it's public, shouldn't start with _

classes/kohana/request/client/curl.php
```diff
@@ -24,8 +24,10 @@ public function _send_message(Request $request)
 		// Response headers
 		$response_headers = array();
 
+		$options = array();
+
 		// Set the request method
-		$options[CURLOPT_CUSTOMREQUEST] = $request->method();
+		$options = $this->_set_curl_request_method($request, $options);
```
Yahasana, line 30: `CURLOPT_CUSTOMREQUEST` is ok for things like "GET", "POST","PUT", "DELETE", "CONNECT" and so on

Ikke, line 30: If you read the issue linked to this pull request, you'll see that it actually gave a problem to someone.

classes/kohana/request/client/curl.php
Ikke, line 119: I'm just following the style of this file. Most methods that start with a _ are public.

classes/kohana/request/client/curl.php
Ikke, line 27: Although they aren't necessary, I do always add them for clarity. It tells someone reading the code there's going to be an array.

PR #286, opened 2012-08-09
Add IS.METHOD functions

nike-17:
Little fix that's fix annoying problem
-- all the time i need to check type of request 0method in controller i should make ugly check like 
$this->request->method() === 'POST'
it can be reason of some bugs e.g.

$this->request->method() === 'post'

is think 
$this->request->is_post() 
is more readable and more bug less way

### DIFF
--- a/classes/kohana/request.php
+++ b/classes/kohana/request.php
@@ -1264,7 +1264,43 @@ class Kohana_Request implements HTTP_Request {
 
 		return $this;
 	}
-
+	
+	/**
+	 * Returns whether this request is GET
+	 *
+	 * @return  boolean
+	 */
+	public function is_get() {
+		return ($this->method() === 'GET');
+	}
+	
+	/**
+	 * Returns whether this request is POST
+	 *
+	 * @return  boolean
+	 */
+	public function is_post() {
+		return ($this->method() === 'POST');
+	}
+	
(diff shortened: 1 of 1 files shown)


cs278:
Or just, `$this->request->method() === Request::GET`.

zombor:
@cs278 has it right.

antmat:
It is more comfortable to use is_\* method. Besides it is more safe.

sergeyklay:
Yes, that's right.

it's comfortable:
`method() === Request::GET`
`method() === Request::POST`
`method() === Request::PUT`
`method() === Request::DELETE`

nike-17:
Серега иди в жопу. Реально заябывает писать эту поебень.

sergeyklay:
:D

sergeyklay:
@nike-17 Кстати, кстати...

``` php
public function post_isset()
{
    if ( ! $this->post())
    {
        return FALSE;
    }

    $params = func_get_args();
    foreach ($params as $param)
    {
        if ($this->post($param) === NULL)
        {
            return FALSE;
        }
    }

    return TRUE;
}
```

Например

``` php
Request::current()->post_isset('post_var1', 'post_another_var', 'yet_another_var');
```

:)

PR #287, opened 2012-08-21
Update classes/kohana/request.php

beautyfree:


zeelot:
Please file a ticket at dev.kohanaframework.org and then make the pull request for the right branch (3.2/develop instead of master). We need to be able to track all progress :)

PR #291, opened 2012-09-04
Fixed issue 3499: widont for text with HTML

glamorous:
I have added some code for the cases you want widont work with HTML-text. There was now a problem when the last space between words was in a HTML-tag. This code fix this issue.

More information about the issue: http://dev.kohanaframework.org/issues/3499

### DIFF
--- a/classes/kohana/text.php
+++ b/classes/kohana/text.php
@@ -588,12 +588,22 @@ class Kohana_Text {
 	 *     echo Text::widont($text);
 	 *
 	 * @param   string  $str    text to remove widows from
+	 * @param   bool    $html	check when words are in HTML-tags, default FALSE
 	 * @return  string
 	 */
-	public static function widont($str)
+	public static function widont($str, $html = FALSE)
 	{
 		$str = rtrim($str);
-		$space = strrpos($str, ' ');
+
+		if($html === TRUE)
+		{
+			$words = explode(' ', strip_tags($str));
+			$second_last = count($words) - 2;
+			$second_last_word = $words[$second_last];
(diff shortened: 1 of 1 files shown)


classes/kohana/text.php
```diff
@@ -588,12 +588,22 @@ public static function number($number)
 	 *     echo Text::widont($text);
 	 *
 	 * @param   string  $str    text to remove widows from
```
evanpurkhiser, line 590: Not a big issue, but you could probably lose a space before the description here so it lines up with the next description.

GeertDD:
I’d like to note that `Text::widont('singleword', TRUE);` results in a "PHP Notice: Undefined offset: -1".

Also, there’s another bug, try this: `Text::widont('problem with repeated words at the end end', TRUE);`.

I don’t like the idea to `explode()` a string on spaces. For longer strings this will result in a huge array. Drupal approaches widont with a regex. Might be worth looking at: http://drupalcode.org/project/typogrify.git/blob/HEAD:/typogrify.class.php#l151

glamorous:
Thanks Geert, I will look into your link and do another pull request.

zombor:
You would need tests for this. Also, don't make a PR into master, use develop.

glamorous:
Is there a little doc about writing tests for kohana or other stuff I would look into before contributing?

zeelot:
Sorry for not replying to your question. If you're still interested in contributing to the project, I would be glad to teach you the process on IRC.

Thanks for the PR anyway, finally got around to merging someone else's updated PR.

zombor:
https://github.com/kohana/kohana/blob/3.3/master/CONTRIBUTING.md

PR #292, opened 2012-09-07
fix redirect (kohana 3.1)

halkeye:
URL::site takes in a TRUE/FALSE (and is checked with ===) not the index file.


halkeye:
Pull request is for the wrong branch, closing

PR #293, opened 2012-09-09
3.1/fix redirect

halkeye:
URL::site takes in a TRUE/FALSE (and is checked with ===) not the index file.


halkeye:
Should fix http://dev.kohanaframework.org/issues/4560

PR #294, opened 2012-09-24
Bug fix for Issue #4178 : Num::bytes & Text::bytes take different approach

evulse:
Added the use of SI prefixes to the Num helper to help align with text helper.

Added the use of a boolean to be able to disable this function for backwards compatibility.


zeelot:
Could you clean this up so only your commits are in the diff? Would also be nice to have tests for this since it's easily testable :)

evulse:
Just cleaned the code up to only include modifications. Also modified test to contain the extra variable.

zeelot:
Ok I took a quick look at the diff and we cannot add a method parameter, that's an API change and belongs in a major release. If this isn't possible without adding a parameter, it will need to go into 3.3 (maybe 3.4 if depending on what @zombor thinks)

zombor:
3.4. You also need to clean those commits up. There shouldn't be any merge commits.

PR #296, opened 2012-09-24
Bug fix for Issue #4605 : Unable to pass body params in external PUT requests (Windows)

evulse:
Change Curl class to use custom request for post requests only as put requests must be set with CURLOPT_INFILE and CURLOPT_INFILESIZE.

Original Bug #4515 only has issues with POST requests and this has not been changed with the new commits.


zeelot:
Merged, thanks

PR #299, opened 2012-10-05
Fixed ticket #4013 - use array_key_exists instead of isset check for get() method

craig-sparks:


classes/kohana/arr.php
```diff
@@ -273,8 +273,8 @@ public static function range($step = 10, $max = 100)
 	 * @return  mixed
 	 */
 	public static function get($array, $key, $default = NULL)
-	{
-		return isset($array[$key]) ? $array[$key] : $default;
+	{	
+		return array_key_exists($key, $array) ? $array[$key] : $default;
```
Yahasana, line 277: array_key_exists is poor performance. both versions work the same for `Arr::get(array('k' => NULL), 'k')`, but different for `Arr::get(array('k' => NULL), 'k', 'foo')`.  if you need to care the null value, then don't use `arr::get`

stillman:
I suggest the following solution:

``` php
return (isset($array[$key]) or array_key_exists($key, $array)) ? $array[$key] : $default;
```

zeelot:
@stillman do you think it's safe to assume that most of the time, we are looking for a key that is there? Is it really a good idea to use both `isset` and `array_key_exists`? I remember seeing an article about this but I'm still unsure of what the best way to do this is.

stillman:
Sure, why not? I think I saw the same article

rjd22:
`array_key_exists` is really bad for performance. It takes 4 times as long as `isset`: http://juliusbeckmann.de/blog/php-benchmark-isset-or-array_key_exists.html

While ugly it is the best way to make a combination of `isset` and `array_key_exists` that way we have the best of both worlds.

merryech0:
I think combination of is not best way. After reading article http://juliusbeckmann.de/blog/php-benchmark-isset-or-array_key_exists.html I made some tests on arrays with different lenght (1000, 10000, 100000 and 1000000 items). On all arrays time spent to isset and array_key_exists is constant and not depends of length for both functions. For situations where key does not exists in array we will wait for isset and then array_key_exists. Imho these cases are very frequent.

shadowhand:
I think this is not a good idea in general. NULL means "not defined", so why should it returned instead of a defined default?

zombor:
If an array has a key that is specified as NULL, that should be returned instead of the default. You are requesting the value of a key of an array. If it exists, the value should be returned, period. NULL is a value.

TBH, I thought this was the current behavior, but I see it's been "broken" for a very long time :)

shadowhand:
As I just said, NULL means "not defined", hence it is not returned, just like `isset()`. Is it really worth making one of our most-used functions 4x slower and changing the behavior of something that has been the same for 2 major versions?

zombor:
The value itself is not defined, but the key in the array is obviously defined. That's what this method should do: Give me the value of the specified key in the array. Right now it does not do that. This method is not anything like isset(), they serve two completely different roles.

merryech0:
shadowhand, are you really think that array_key_exists is slowest function in all you code? :-D

shadowhand:
Of course not.

merryech0:
I think we need right logic instead this "economy".

And I agree with zombor

zeelot:
we shouldn't be calling array_key_exists() on objects, `3.3.0` has some other implementation which is also wrong but I will write a bunch of tests soon and figure this out… I'm kind of agreeing with @shadowhand  though simply because of how much complexity we are adding when it's just as easy to understand that Arr::get() and similar methods use `isset()` internally =/ I'll get back to this soon though (also, I cannot hate php more than I do right now)

zeelot:
Turns out we can't even leave it as a simple `isset()` because ArrayObject::offsetExists()`returns`TRUE`when a value is`NULL`but it is what is called when you run`isset($arrayObject['something'])`… so our current implementation isn't even consistent. I think we need to simply check if $array is an instance of ArrayAccess and go from using isset/array_key_exists to using offsetExists()… we also need to change things like`Validation::offsetExists()`to return`TRUE`when values are`NULL` :) at least I think…

This is all probably best done in `3.4`\+ since it would change a lot of behavior.

Thoughts? This is what I tested:

```
$one = new ArrayObject(array('foo' => NULL));
$two = array('foo' => NULL);

var_dump(isset($one['foo'])); // TRUE!
var_dump(isset($two['foo'])); // FALSE!
```

stillman:
@Zeelot, what PHP version do you use? I've tested on 5.3.13 and 5.4.3 and they return FALSE in both cases. But! 5.2.11 returns values as in your example. From this we can conclude that PHP < 5.3 has the bug

stillman:
I think this behavior is the most correct

``` php
function arr_get_old($array, $key, $default)
{
    return isset($array[$key]) ? $array[$key] : $default;
}

function arr_get_new($array, $key, $default)
{
    return (isset($array) or array_key_exists($key, $array))
        ? $array[$key]
        : $default;
}

$one = new ArrayObject(array('foo' => NULL));
$two = array('foo' => NULL);

var_dump(arr_get_old($one, 'foo', 'does not exist!')); // "does not exist!", incorrect
var_dump(arr_get_old($two, 'foo', 'does not exist!')); // "does not exist!", incorrect

var_dump(arr_get_new($one, 'foo', 'does not exist!')); // "NULL", correct
var_dump(arr_get_new($two, 'foo', 'does not exist!')); // "NULL", correct</pre>
```

zeelot:
We don't currently require `5.3` so we can't actually rely on that functionality being fixed… however this is my code:

```
$data = new ArrayObject(array('foo' => NULL));

echo PHP_VERSION;

var_dump($data['foo'], isset($data['foo']));die;
```

outputs:

```
5.3.5
null
boolean true
```

I don't think we should ever have code that could possibly call `array_key_exists()` on an object.

> For backward compatibility reasons, array_key_exists() will also return TRUE if key is a property defined within an object given as search. This behaviour should not be relied upon, and care should be taken to ensure that search is an array.

This is what I think we should do:

```
public function get($array, $key, $default = NULL)
{
    if ($array instanceof ArrayAccess)
    {
        return ($array->offsetExists($key)) ? $array[$key] : $default;
    }
    else
    {
        return (isset($array) OR array_key_exists($key, $array)) ? $array[$key] : $default;
    }
}
```

Optionally we have been talking about removing support for objects in the `Arr` class which is something we could consider in `3.4`

zeelot:
In fact… none of these changes will be making it into `3.2`, we are simply talking about what to do in `3.3.0` since a change has already been made there and it seems to be incorrect.

stillman:
This is very strange... Your code returns the following for me:

<pre>5.3.13
NULL
bool(false)</pre>

> We don't currently require 5.3

Don't you? I thought that 3.3 requires 5.3

zeelot:
I don't think so but that's irrelevant since I am on `5.3` and still have this insane result :)

stillman:
Sorry for offtopic, but how about tihs? :)
http://forum.kohanaframework.org/discussion/comment/62572#Comment_62572

stillman:
By the way, 5.3.5 returns the same result as yours. Maybe required PHP version for 3.3 should be raised then?

stillman:
@shadowhand, there already was a problem with isset (http://dev.kohanaframework.org/issues/3296) which you have fixed by the way ;)

zombor:
Kohana 3.3 requires php 5.3. @Zeelot is being silly.

zeelot:
This is probably not a good enough reason to raise the minimum version. Also, I am still not ok with `array_key_exists()` ever being used on objects. It's unnecessary and asking for trouble.

paulchubatyy:
Guys, 

Does arr::get have anything to do with objects? This is "array helper" (remember ol' good times we used to call it so). Why should we care about the objects here?

zeelot:
Like I mentioned above, we could typehint the methods to only allow real arrays but this definitely has uses with ArrayAccess objects like DB Result objects, Validation, etc.

It's an option we discussed on IRC and we can consider it for `3.4.0`

rjd22:
I think it should at least support `ArrayAccess` objects. Since those are still array's. But we should make a new `Obj` class that supports normal object.

If we don't do this the code of `Arr` will become a mess.

zeelot:
We are only talking about `ArrayAccess`, no other type of class needs this stuff.

rjd22:
We should close all related pull requests and add this into 3.4. Zeelot has the following code been tested and is it acceptable for the 3.4 version? IF so I will open 1 final pull request:

https://github.com/kohana/core/pull/299#issuecomment-9505211

zeelot:
I think we should discuss what we want to do in 3.4 but I just want to get 3.3 out before I think too much about the rest :-) let's move this discussion to the redmine ticket I made so I can find it easily later

PR #300, opened 2012-10-16

yakatz:
This (maybe) should be targeted to 3.4/develop, but that does not exist yet.

(This is working in production for me.)

Replaces https://github.com/kohana/core/pull/269

http://dev.kohanaframework.org/issues/4614


enov:
@yakatz Thanks for this. This should have gone in 3.4/develop... but note that there has been some work done there to separate module definition and initialization.

See #530 and kohana/kohana#42.

PR #301, opened 2012-10-16
Fixes Issue #4615 #4022 and #4021

rjd22:
I've fixed the 3 issues in the Arr class. This fix won't work upstream since in Kohana 3.3 we start using arrayaccess. I need to checkout if these are really good for performance but I doubt it ;)

### DIFF
--- a/classes/kohana/arr.php
+++ b/classes/kohana/arr.php
@@ -274,7 +274,7 @@ class Kohana_Arr {
 	 */
 	public static function get($array, $key, $default = NULL)
 	{
-		return isset($array[$key]) ? $array[$key] : $default;
+		return (isset($array[$key]) OR array_key_exists($key, $array)) ? $array[$key] : $default;
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


kemo:
I don't think the isset() call should stay in there

rjd22:
@kemo I think it should. The problem with array_key_exists is that it's 4 times slower than isset. Using both will get us the best of both worlds.

rjd22:
@kemo look at the discussion here: https://github.com/kohana/core/pull/299

kemo:
Speed over readability does make a point in this case, I agree. The `array_key_exists()` is a fail-over for NULL / undefined values anyways so... Ignore my previous comment

zeelot:
can any of you guys explain this to me? https://bugs.php.net/bug.php?id=46354 it seems to say that since `5.3.0`, array_key_exists() does not work on any Object, but my test shows it does… so I'm confused but this would affect whether we can make this change or not (or in which version).

rjd22:
Since after a lot of debate we came to the conclusion this should be fixed in 3.4 I'm closing this pull.

PR #302, opened 2012-10-17
Fix bug when Kohana::magic_quotes == 1 in php version < 5.4

bmxnt:
In version of PHP < 5.4 Kohana::$magic_quotes is always TRUE


zombor:
A few problems with this:
- Is there a redmine issue number? Please create one if not, and link your PR to that issue.
- Put the ticket number in the commit message.
- This should target 3.2/develop, not master.

PR #303, opened 2012-10-18
Fix issue #4616

bmxnt:
In version of PHP < 5.4 Kohana::$magic_quotes is always TRUE


zeelot:
Thanks, I'll merge this as soon as I get a chance! Closing #4616 as a duplicate but I will still take this pull request :)

zeelot:
should be fixed already

PR #304, opened 2012-10-21
Set secure based on X-FORWARDED-PROTO

jimktrains:
If the HTTPS flag is not set, check the
X-FORWARDED-PROTO header to check if the
load balancer received a secure request

### DIFF
--- a/classes/kohana/request.php
+++ b/classes/kohana/request.php
@@ -126,7 +126,8 @@ class Kohana_Request implements HTTP_Request {
 					$method = HTTP_Request::GET;
 				}
 
-				if ( ! empty($_SERVER['HTTPS']) AND filter_var($_SERVER['HTTPS'], FILTER_VALIDATE_BOOLEAN))
+				if (( !empty($_SERVER['HTTPS']) AND filter_var($_SERVER['HTTPS'], FILTER_VALIDATE_BOOLEAN) )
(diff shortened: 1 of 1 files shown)


zombor:
Is this testable? Provided unit tests are required if possible.

zombor:
Also, please reference the redmine ticket number in the commits.

cs278:
Blindly trusting user data to determine if a request is secure doesn't seem very wise.

zeelot:
yeah I always thought this was just specific to how you configured your load balancer set up, no?

jimktrains:
Yes, it is a consequence of how the load balancer is set up.  I've seen
this header used quite frequently though.

The load balancer should scrub the header from input, but that doesn't help
if you don't have a load balancer.  Perhaps having the header be
configurable (and set to null by default) would be a good option?

Jim

rjd22:
Isn't it possible to just overrule the request class for these cases? I think it's not something that should be included in core since this is environment specific.

zeelot:
I agree about this being something you can extend the request class for. I don't want this to be a default behavior but we can consider making it easy to configure in the future.

cs278:
> I agree about this being something you can extend the request class for. I don't want this to be a default behavior but we can consider making it easy to configure in the future.

This is a bit of a weird answer given Kohana supports setting the remote IP based on proxy headers.

zeelot:
I agree with that as well :P but I would rather not copy our past mistakes. The request class in the future will not have any of this functionality. On a side note, we seem to validate the ip with a list of accepted proxies, so it's a tiny bit better.

PR #305, opened 2012-10-21
When setting POST params, set the request method

jimktrains:


jimktrains:
I just tested this and it appears my patch isn't needed

PR #315, opened 2012-10-25
Build URI using preg_replace_callback

cbandy:
http://dev.kohanaframework.org/issues/4630

### DIFF
--- a/classes/Kohana/Route.php
+++ b/classes/Kohana/Route.php
@@ -35,8 +35,14 @@
  */
 class Kohana_Route {
 
+	// Matches a URI group and captures the contents
+	const REGEX_GROUP   = '\((.*?)\)';
+
 	// Defines the pattern of a <segment>
 	const REGEX_KEY     = '<([a-zA-Z0-9_]++)>';
+
+	// Asserts that remainder of the string has a pair of parentheses or none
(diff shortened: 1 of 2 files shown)


classes/Kohana/Route.php
```diff
@@ -501,109 +507,88 @@ public function is_external()
...
 
-			$uri = str_replace($key, $params[$param], $uri);
-		}
+			return array($result, $required);
+		};
+
+		list($uri) = $compile($this->_uri, TRUE);
 
 		// Trim all extra slashes from the URI
 		$uri = preg_replace('#//+#', '/', rtrim($uri, '/'));
```
cbandy, line 588: This line should probably go. I didn't see any tests calling for its behavior, but I also didn't try taking it out.

zeelot:
Thanks for the work :) I'll take a look at this over the weekend… gave it a quick look yesterday but haven't tried it out!

zeelot:
Sorry for the delay… it looks good and the new tests seem to cover the cases people were complaining about. I found the formatting to be pretty confusing so I decided to break the line length rule in favor of readability. I would like it to be a bit cleaner but I'm not sure how to fix it up more at the moment and I want to move on to other tickets :)

Thanks again for the work!

PR #318, opened 2012-10-30
Fixed links to tutorial for template driven site

yakatz:
Might want to include the content instead and credit that site, just in case it changes again.

### DIFF
--- a/guide/kohana/tutorials/templates.md
+++ b/guide/kohana/tutorials/templates.md
@@ -1,7 +1,7 @@
 Making a template driven site.
 
-<http://kerkness.ca/wiki/doku.php?id=template-site:create_the_template>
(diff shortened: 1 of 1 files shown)


zombor:
We should just remove all external "tutorials" to other sites in the docs. For example, this one is is just out of date, and won't work in 3.3. There's other links that are broken as well.

Also, this PR would need to target 3.3/develop, not master.

zombor:
You should also make a ticket in redmine for any issues you file.

yakatz:
I just used the editor built into Github because it was a small and quick change. I am looking at how long it would take to create an updated doc for templates and I might just do it.

zombor:
K, just make sure you read the development guidelines:

https://github.com/kohana/kohana/wiki/Kohana-Development-Procedures
https://github.com/kohana/kohana/wiki/developers

PR #320, opened 2012-11-04
Use the current exception handler during shutdown

evanpurkhiser:
It's possible that a different exception handler is defined than the
default `Kohana_Exception::hander`.

Refs [#4627](http://dev.kohanaframework.org/issues/4627)


zeelot:
Sorry for the long delay, thanks for the patch.

evanpurkhiser:
Appreciate it! Looking forward to bringing some of these awesome new patches into my Kohana project!

PR #321, opened 2012-11-15
Update classes/Kohana/Valid.php

dkobia:
- Fixed an issue with range() excluding the minimum and maximum values
- closes #4660


zeelot:
Any chance you could update the tests to include the edge values? They should fail without your fix and pass with it :)

zeelot:
Fixed by another pull request, thanks though :)

PR #322, opened 2012-11-20
Accurating sources (removing crappy spaces)

rc5hack:
All trailing spaces removed, no code changed.


zombor:
We don't support 3.0 anymore. We currently support versions 3.2 and 3.3. If you have a patch like this for those versions, make sure you target the develop branch, not master.

PR #324, opened 2012-12-13
Changed the Security::token() method to remove the token from the Session

mAiNiNfEcTiOn:
Changed the token() method to make a get_once, so that the token does not remain in the Session.
Check the case where, the user actually uses your form once, and it is validated correctly and does the desired process.
The CSRF protection's token used in the whole process remains in the Session, therefore allowing the user to ignore your form and make posts directly to the endpoint of your form, therefore only validating with the token in the Session, until the Session ends.

### DIFF
--- a/classes/Kohana/Security.php
+++ b/classes/Kohana/Security.php
@@ -43,7 +43,7 @@ class Kohana_Security {
 		$session = Session::instance();
 
 		// Get the current token
-		$token = $session->get(Security::$token_name);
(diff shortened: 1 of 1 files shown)


randomecho:
You should retarget your PR to the develop branch instead.

zombor:
Please read the CONTRIBUTING document in the main integration branch.

zeelot:
here: https://github.com/kohana/kohana/blob/3.3/master/CONTRIBUTING.md

PR #325, opened 2012-12-14
Changed the token() method to only get_once the generated token

mAiNiNfEcTiOn:
if the user uses the form with CSRF protection once, it will generate the token, store it in the session and  when the form is submitted it will (or should) be validated using the check() method.

This is all fine, the check() will basically get the token and validate if the submitted value matches the existing in the form. All good, the process continues.

... however if the user, crafts a form post (using, for example, javascript) with the same CSRF token defined, submitting it to the same 'action' that processes the form... It will be as valid as the previous one, because it's re-using the token in the session.

With the get_once, the token is generated and returned, checked and _puff_... vanished from the session.

What do you think?

### DIFF
--- a/classes/kohana/security.php
+++ b/classes/kohana/security.php
@@ -43,7 +43,7 @@ class Kohana_Security {
 		$session = Session::instance();
 
 		// Get the current token
-		$token = $session->get(Security::$token_name);
+		$token = $session->get_once(Security::$token_name);
 
 		if ($new === TRUE OR ! $token)
 		{


randomecho:
Do you also have a related [Redmine issue](http://dev.kohanaframework.org/projects/kohana3/issues) open for this?

Please refer to [the CONTRIBUTING guide](https://github.com/kohana/kohana/blob/3.3/master/CONTRIBUTING.md#contributing-to-the-project)

> All features and bugfixes must be fully tested and reference an issue in the tracker, there are absolutely no exceptions.

mAiNiNfEcTiOn:
Hi Soon Van,

Yes, the http://dev.kohanaframework.org/issues/4675

Since I knew the 'solution' (if there's a problem) I kind of /fixed it/.

Shouldn't I ?

Sincerely,
_Ricardo Machado_

Em 14-12-2012 01:58, Soon Van escreveu:

mAiNiNfEcTiOn:
Ok, so instead of _patch_ I made a duplicate as a bug report. It's now http://dev.kohanaframework.org/issues/4676 the correct issue.

mAiNiNfEcTiOn:
Since the issue was - seemingly - ignored, because you want to "support multiple tabs opening the same form, without changing the token", I closed the pull...

Nevertheless, I do not agree, as this leads to a serious CSRF flaw. An approach to fix the issue you want to support ( the "multiple tabs" thingy ) would be maintaining a hashmap/dictionary in the Session for those forms...

After a year, you finally decided to "fake-a-look" (pun intended). This leads me to forget about helping this community. It demands too many steps to help fixing issues. Your organization lacks order (and purpose) by enforcing it.

Wishes of success!

JackEllis:
@mAiNiNfEcTiOn It was only me who questioned it, not the core development team, and they may disagree...

mAiNiNfEcTiOn:
@JackEllis thanks for your reply. First I'd like to say that what matters to me most is the community's opinion, not a _core development team_ of any framework. The same way I can use this, I can use other frameworks.

At least, you answered this issue after 1 year ... That's more than the _core development team_ right? Right.

Last but not least, I respect your opinion, regardless of disagreeing. I only am sad that I took my time - which is as valuable as the _core development team_'s time - to file an issue (or what I think it is an issue) and the only response from the _core development team_ was "Did you filed the issue in the Redmine ? Check the contribution.md" speech. Are you serious? That's how you want to 'grow' as community? By forcing people to use different platform for the same purpose?

In the end, after I actually took the time to file the issue in the Redmine platform. And still, a year has passed. Well, seems like that's an issue that it's not even worth to invalidate :-/

Now I see the point in so many comments I've heard about the _core development team_'s way of dealing with the actually community. Perhaps that's other reason to explain the (lack of) documentation of the framework.

For me, mottos like "It was hard to write. It should be hard to read" do not fit in my way of thinking.

JackEllis:
@mAiNiNfEcTiOn Preaching to the choir ;) With regards to your talk about the community, the reason I found the issue was because of this: http://forum.kohanaframework.org/discussion/12366/announcing-our-first-kohanathon#Item_11

mAiNiNfEcTiOn:
@JackEllis thanks for mentioning.  I totally understand the recent activity then :)

However, apparently it means that (after a year) they just run a _Kohanathon_ to clean bugs. With a reward !!
Open source is really motivational :D

Oh well... Hope you win it! Best of luck!

zeelot:
I'm a core developer so you can direct your anger my way :)

It seems that if you pass in `true` as the first parameter, you get the functionality you're looking for. Am I missing something?

Side note: I hate the implementation of this function entirely and want to replace it as soon as possible.

zeelot:
And sorry for letting PRs slip for so long… we're trying to catch up now.

mAiNiNfEcTiOn:
@Zeelot no anger, just facts :)

Regarding that... The problem is not on the TRUE parameter, but the in the check() function... Unless you're saying that we should do check() and right after that, just generate a new session token just for the fun and renovation.

In my point of view, the check() function should do a get_once, because it should erase the token used in the beginning from the session. Otherwise, imagine a _bot_ that requests once a form (say for sending an e-mail) and gets the csrf token, submits that form over and over again, always passing the same token...

... that would always pass in the check() until he 'reloaded the form' to actually generate a new token ( even with the _TRUE_ value ).

But that's my opinion :)

Nevertheless, I've removed that repo from my github account so, the pull has no effect I think.

zeelot:
But the session logic is only in the `token()` function. Would updating the `check()` function to pass `true` into the `token()` function behave like you expect?

The change you made is not in the function you're talking about, which is what I'm confused about, I believe.

mAiNiNfEcTiOn:
Sorry, but I didn't remember the whole situation (been quite a time) so now that I refreshed my memory, here's the thing:

1 - The _check()_ method calls the Security::token() ... So, basically it does not change the token - the way you have the code now - nor erases it from the Session.

2 - The _token()_ method does a _Session::get()_ ( instead of the mentioned/used _Session::get_once()_ in the pull request ). So, if the token doesn't exist it's created, but if the token exists, only when the TRUE is passed it's regenerated.

3 - Passing TRUE as an argument, when the _check()_ method calls the _Security::token()_, solves it... But, it's re-setting the Session variable instead of erasing it (which is what the _Session::get_once()_ does, right? ).

So, yes, the TRUE as a parameter inside of the check() would solve it, not in a perfect way but at least you wouldn't have the problem.

zeelot:
Ok I see what it was trying to do now. Unfortunately, we can't break this for people that do expect it to work with multiple tabs but I would like to get to the bottom of this so I can find a proper solution (in the next major version if I have to) that will allow people to use the stricter settings.

What I am still trying to find is how the attacker is finding the used CSRF though. Can you describe the attack that you are preventing by regenerating the token so often?

mAiNiNfEcTiOn:
Hum... the way that I see it, the attacker only needs to do the following steps:

1 - Access the form once
2 - Read the token from the hidden field
3 - Fill in the form without errors and submit it over and over again... The Session token will always be the same, and - as long as the attacker doesn't refresh the form (which generates the token) - it'll always pass :)

4 - With a get_once() the check would only get it once and remove it from the Session... the attacker would be able to submit only once and then it would re-generate one at the _check()_ time since there would be no session token :)

Was I clear?

zeelot:
Except that step 1 is not possible. How is the attacker accessing your form? If he could access the form, it wouldn't matter how many times you generate that token, the attacker has it. Am I missing something? Could you describe how step 1 is being done?

mAiNiNfEcTiOn:
Well... 

... point taken. I was so focused on automated submitting reusing the same token, that I never questioned myself about "Well, if he had the access to the form, would he need to do that? And if so, why couldn't he if I changed the code this way?"

@Zeelot I'm sorry for wasting your time with this... And thank you for having the time to do that question. Took a year, but I liked a lot the answer

zeelot:
No worries, sorry again for the long delay and I wasn't entirely sure I was right so I was trying to see if you had thought of something I was missing :) thanks for taking the time to go through this with me even after you've give up on us.

mAiNiNfEcTiOn:
Well, I didn't gave up on you :) I am still using Kohana (3.3 currently) in my projects... However, it's like I said before, the fame of the _core team_ is not good as well as the fact that it is less documented.

Nevertheless I prefer less documentation and code readability and the ability to follow the flow of the code. With frameworks like Symfony2 that rely a lot on DI and Service containers, you sometimes lose the track of the flow.

Thanks once again ;-)

PR #326, opened 2012-12-17
Update classes/kohana/response.php

egidijusz:
http://dev.kohanaframework.org/issues/4678


evanpurkhiser:
I'm pretty sure this issue is already tracked.

http://dev.kohanaframework.org/issues/4592
http://forum.kohanaframework.org/discussion/comment/74341#Comment_74341

Looks like the fix was merged into 3.3, but was never merged into 3.2. It's targeted at 3.2.3

egidijusz:
Yes, you're right. And I think woodys fix is more appropriate. Too bad it isn't merged into 3.2 :(

PR #329, opened 2013-01-15
Expression missing parentheses

dishuostec:
Missing parentheses in classes/Kohana/Validation.php.


shadowhand:
There's nothing wrong with the current code... are you adding these just for clarity?

In any case, the commit message is wrong and you are targeting the wrong branch with your pull request.

PR #330, opened 2013-02-05
Arr::get work with ArrayAccess and NULL values

hkdobrev:
This is my take on new `Arr::get()` based on the discussions on #301 and #299 and the many issues related to them.

---

Passing `NULL` as first argument would return the default.

Passing something else than `array` or instance of `ArrayAccess` would throw an exception.

Using `isset()` for performance reasons.

Using `array_key_exists()` for arrays and `ArrayAccess::offsetExists()` otherwise.

---

This change might throw an exception which was not discussed before. Please let me know if you disagree with that.

### DIFF
--- a/classes/Kohana/Arr.php
+++ b/classes/Kohana/Arr.php
@@ -267,14 +267,32 @@ class Kohana_Arr {
 	 *     // Get the value "sorting" from $_GET, if it exists
 	 *     $sorting = Arr::get($_GET, 'sorting');
 	 *
-	 * @param   array   $array      array to extract from
-	 * @param   string  $key        key name
-	 * @param   mixed   $default    default value
+	 * @param   array|ArrayAccess|NULL   $array      array to extract from
+	 * @param   int|string               $key        key name
+	 * @param   mixed                    $default    default value
 	 * @return  mixed
+	 * @throws Kohana_Exception If $array is not of expected types
 	 */
 	public static function get($array, $key, $default = NULL)
 	{
-		return isset($array[$key]) ? $array[$key] : $default;
+		// using isset for performance reasons
+		if (isset($array[$key]))
+			return $array[$key];
+
+		if ($array === NULL)
+			return $default;
+
+		if (is_array($array))
+		{
+			return array_key_exists($key, $array) ? $array[$key] : $default;
+		}
+
+		if ($array instanceof ArrayAccess)
+		{
+			return $array->offsetExists($key) ? $array[$key]: $default;
+		}
+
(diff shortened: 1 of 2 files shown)


zombor:
We would need updated tests for this to at least consider it. This is quite different behavior from the current behavior.

hkdobrev:
@zombor I've just added some extensive tests. All core tests pass. I haven't run the tests for all of the official Kohana modules, but I don't think this is in the scope of this PR.

hkdobrev:
I have rebased my commits into one. Any updates or opinions on this PR?

hkdobrev:
Hey, everyone! Any thoughts on this one?

cc @zombor

classes/Kohana/Arr.php
```diff
@@ -274,7 +274,27 @@ public static function range($step = 10, $max = 100)
 	 */
 	public static function get($array, $key, $default = NULL)
 	{
-		return isset($array[$key]) ? $array[$key] : $default;
+		// using isset for performance reasons
+		if (isset($array[$key]))
+			return $array[$key];
+
+		if (is_null($array))
```
Yahasana: `if ($array === NULL)`

hkdobrev: Is it faster?

hkdobrev: It actually is. I'll change it.

hkdobrev:
Hey, everyone! What's up with this one?

Thanks!

cc @zombor @Yahasana

classes/Kohana/Arr.php
```diff
@@ -274,7 +274,27 @@ public static function range($step = 10, $max = 100)
...
+		{
+			if (array_key_exists($key, $array))
+				return $array[$key];
+		}
+		elseif ($array->offsetExists($key))
+		{
+			return $array[$key];
+		}
+
+		return $default;
```
Yahasana:
``` php
    // using isset for performance reasons

    if (isset($array[$key]))
        return $array[$key];

    if ($array === NULL)
        return $default;

    if (is_array($array))
    {
        return array_key_exists($key, $array) ? $array[$key] : $default;
    }

    if ($array instanceof ArrayAccess)
    {
        return $array->offsetExists($key) ? $array[$key] : $default;
    }

    throw new Kohana_Exception('Argument 1 to Arr::get() must be an array or an instance of ArrayAccess');
```

hkdobrev:
@Yahasana Thank you for your input! I don't know how I did not write it like that the first time.

All done now. I've also improved the PHPDoc.

PR #331, opened 2013-02-15
regarding http://dev.kohanaframework.org/issues/4695

chandeeland:
the way HTTP_Exceptions are handled ...exceptionally in the Kohana/Request/Client/Internal.php makes it difficult to override their presentation when exceptions are thrown before Client->execute() such as the case of Route::Filters()

This change allows request to call the overloaded version of the function instead of always forcing Kohana_Exception::response(), while still using it as the default in the case of non kohana exceptions.

### DIFF
--- a/classes/Kohana/Kohana/Exception.php
+++ b/classes/Kohana/Kohana/Exception.php
@@ -103,11 +103,21 @@ class Kohana_Kohana_Exception extends Exception {
 	{
 		try
 		{
-			// Log the exception
-			Kohana_Exception::log($e);
-
-			// Generate the response
-			$response = Kohana_Exception::response($e);
+			if ($e instanceof Kohana_Exception)
+			{
+				$exception_name = get_class($e);
+				$exception_name::log($e);
+				$response = $exception_name::response($e);
+			} 
+            // non Kohana_Exception Exceptions
+			else 
+			{
+				// Log the exception
+				Kohana_Exception::log($e);
+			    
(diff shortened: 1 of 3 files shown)


chandeeland:
I noticed Redirects were showing the error page instead of redirecting properly.  I think I've fixed this issue by handling HTTP_Exception_Expected differently.

zombor:
This would need to go into 3.4 since you removed a public method.

zombor:
Also, I'm really not sure about this change. I don't think I agree with removing that method.

chandeeland:
I could put the public method back in as a wrapper to maintain backwards compat.

what is your concern regarding the change?  I think it works towards kohanas' biggest strength, which is it's cascading loading, and the consistency of over-ride behavior.

sergeyklay:
:+1:

classes/Kohana/Kohana/Exception.php
```diff
@@ -185,6 +195,12 @@ public static function response(Exception $e)
 			$line    = $e->getLine();
 			$trace   = $e->getTrace();
 
+			if ($e instanceof http_exception_expected)
```
zeelot, line 198: Can you fix the casing for this class name?

zombor:
I'm mainly concerned that this removes the http exception logic away from the request internal object, where it really belongs. This was an intentional change in 3.3 so they didn't leak all over the place.

PR #336, opened 2013-03-06
Use == instead of === for exact_length()

krische:
This is a fix for #4713


zeelot:
Reason left on redmine: http://dev.kohanaframework.org/issues/4713

PR #337, opened 2013-03-07
FIX #4712 - Exception on writing log

dm:
Fixing bug report #4712
Could also be merged into 3.3/master

~ Daniel


zeelot:
We want to only filter out things that cannot be cast to a string, not necessarily things like integers. Since we know what this array consists of, we can easily simply `unset()` the few things we know are an issue.

dm:
Hmpf... I wen't with the array_filter thinking this could come up again in the future. Aren't the line numbers returned as strings? (was convinced so, hence the use of is_string, but didn't check it...)

I can change it to unset the 'additional' and 'trace' keys instead.

zeelot:
Maybe you're right but actually, your fix will break the `if (isset($message['additional']['exception']))` that's right below in… I kinda dislike this method in general.

dm:
Terrible, didn't notice it! But then again, as would using `unset`...

Perhaps instead of changing the `$message` var I'd have the `array_filter()` on the second parameter of `strtr()` on both calls. And to avoid the issue with ints I can change it to `is_scalar()`, although that removes the NULLs as well; so maybe:

`$string = strtr($format, array_filter($message, 'is_scalar'));`

and

`$string .= PHP_EOL.strtr($format, array_filter($message, 'is_scalar'));`

If all else causes issues the `unset()` can work if duplicating into a second `$message2` for the `strtr()` calls...

Good enough?

zeelot:
Looks good but now why are there 7 commits in this pull request :(

dm:
Ouch, I messed up with my two local clones of the repo!

Damn it... fixing it now

zombor:
This needs to be rebased into one commit.

dm:
If you notice I messed and merged my local repo (clone A) into my local repo (clone B), shouldnt have worked on both the github clone and the one that's a submodule of the current working project, my bad!

zeelot:
Yeah, I saw that heh, just fix that up or make a new pull request if you need to with a single commit and I'll merge it in. Thanks for the fix.

dm:
Fixing this and issuing a new pull request: 338 ;)

Cheers

PR #338, opened 2013-03-07
3.3/develop

dm:
Left the two commits to retain my submodule repo history, hopefully that's not an issue to accept the pull request. ;)

Cheers,
~ Dani


zeelot:
thanks :) yeah, in the future, we would appreciate a single commit for things like this, you can usually do that pretty easily with git.

PR #339, opened 2013-03-14
doc: use "303 See Other" redirects after POST

ktdreyer:
Per RFC 2616 (HTTP 1.1), the HTTP response status code "303 See Other" is the correct way to redirect a client after an HTTP POST. Adjust the documentation to use a 303 redirect instead of 302.

### DIFF
--- a/guide/kohana/mvc/controllers.md
+++ b/guide/kohana/mvc/controllers.md
@@ -128,7 +128,7 @@ A user login action.
 			// Try to login
 			if (Auth::instance()->login($this->request->post('username'), $this->request->post('password')))
 			{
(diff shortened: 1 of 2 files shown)


ktdreyer:
Redmine ticket filed at http://dev.kohanaframework.org/issues/4720

zombor:
There was a reason we chose 302 vs 303 even though 303 is "correct". I'm not sure offhand, maybe it was browser support?

ktdreyer:
Closing for #343, which is filed against the proper branch (3.3/develop)

sergeyklay:
:+1:

PR #340, opened 2013-03-14
doc: fix HTML in validation view example

ktdreyer:
The unordered list in the HTML example was unclosed. Add a closing ul element to fix it.


zeelot:
Could you make this pull for `3.3/develop`? We don't merge into master branches.

PR #341, opened 2013-03-17
call from previous gets instance of route

kazelot:


zeelot:
Could you make this pull request for the `3.3/develop` branch? We don't merge into master branches.

PR #342, opened 2013-03-19
doc: fix HTML in validation view example

ktdreyer:
The unordered list in the HTML example was unclosed. Add a closing ul element to fix it.

This is filed in Redmine as http://dev.kohanaframework.org/issues/4721

Previously erroneously submitted against 3.3/master as #340


enov:
Thanks @ktdreyer , sorry it took us too long to merge this.

PR #351, opened 2013-05-14
Add option to use wildcard keys in Arr::path()

tommcdo:
In `Arr::path()`, when the boolean parameter `$wildcard_assoc` is set to `TRUE`, the array resulting from a wildcard expansion will be associative, using the keys at the wildcard position. The parameter is defaulted to `FALSE` to not alter the behaviour of existing code.


tommcdo:
Closing this request to adhere to contribution guidelines. Will re-open later.

PR #352, opened 2013-05-15
Fixes #4743: Arr::path() to return associative array when wildcard used

tommcdo:
Fixes: http://dev.kohanaframework.org/issues/4743

### DIFF
--- a/classes/Kohana/Arr.php
+++ b/classes/Kohana/Arr.php
@@ -159,11 +159,11 @@ class Kohana_Arr {
 				// Handle wildcards
 
 				$values = array();
-				foreach ($array as $arr)
+				foreach ($array as $wildcard_key => $arr)
 				{
 					if ($value = Arr::path($arr, implode('.', $keys)))
 					{
-						$values[] = $value;
+						$values[$wildcard_key] = $value;
 					}
 				}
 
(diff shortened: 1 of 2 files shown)


enov:
This is a well written bug fix, with tests. However, existing code relying on the numeric sequential keys of the output of this function will break, if we apply this fix.

@acoulton could you kindly comment what to do in this case?

acoulton:
Tricky. Since it changes the response, I think it would have to go to 3.4 with a clear note in the upgrade guide.

It will only be if end-user code is specifically looking at the keys, I think, that it's an issue - they'll still come out in the same sequence as far as I can see? It's probably quite an edge case, but still not safe for 3.3.

@tommcdo if you disagree we'd welcome your opinion.

tommcdo:
I think it's quite unlikely that users are currently relying on the numerical keys with the current implementation, since they're quite frankly not useful at all. However, this is not strictly a bug fix, so it's perfectly reasonable to hold off until 3.4.

enov:
> However, this is not strictly a bug fix, so it's perfectly reasonable to hold off until 3.4.

@tommcdo could you please PR this to 3.4/develop branch.

neo22s:
What we do with this? to 4.0.0? Looks good to me.

acoulton:
I suggest a contrib manually merges this PR to the 4.0 branch and adds to upgrade notes

PR #356, opened 2013-05-29
Fix for #4646

rjd22:
Cleaned up the documentation

### DIFF
--- a/classes/Kohana/Request.php
+++ b/classes/Kohana/Request.php
@@ -454,7 +454,7 @@ class Kohana_Request implements HTTP_Request {
 	/**
 	 * Process a request to find a matching route
 	 *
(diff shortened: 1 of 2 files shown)


zeelot:
Hmm, http://dev.kohanaframework.org/issues/4512 doesn't seem to be related to this PR… do you have the right redmine ticket? I can merge this as soon as we find the ticket :)

rjd22:
I was initially looking at this one but that seems to be fixed already. Then I saw where were still some things that needed fixing so I fixed them right away. I'm not sure if it's needed to make another ticket for it since it's only documentation.

http://dev.kohanaframework.org/issues/4646

zeelot:
I think this is old… let me know if I'm mistaken :)

PR #359, opened 2013-06-02
Allow security disabling in Upload class for unit testing purposes (#4751)

shadlaws:


shadlaws:
Issue: http://dev.kohanaframework.org/issues/4751

rjd22:
@shadlaws It would be nice if you could add unittests to kohana/kohana that test the upload class and this functionality. Just to make sure everything works like it's supposed to.

classes/Kohana/Upload.php
```diff
@@ -31,6 +32,15 @@ class Kohana_Upload {
 	public static $default_directory = 'upload';
 
 	/**
+	 * By default, the methods in this class are secure against fake uploads.  However,
+	 * for unit testing purposes, it may be useful to disable security to test functionality
+	 * downstream.  This should never be used in production installations!
+	 *
+	 * @var  boolean  allow security disabling using [Upload::disable_security]
+	 */
+	public static $allow_security_disabling = FALSE;
```
zeelot, line 41: I'm not sure we need this kind of flag. Having a property to makes one of the methods do nothing seems a bit odd.

classes/Kohana/Upload.php
```diff
@@ -131,7 +144,7 @@ public static function not_empty(array $file)
 		return (isset($file['error'])
 			AND isset($file['tmp_name'])
 			AND $file['error'] === UPLOAD_ERR_OK
-			AND is_uploaded_file($file['tmp_name']));
+			AND (static::$_disable_security OR is_uploaded_file($file['tmp_name'])));
```
zeelot, line 147: Can you use `Upload` instead of `static` to match our conventions?

classes/Kohana/Upload.php
zombor, line 41: I agree. Adding methods "for testing purposes" is wrong.

zeelot:
And just realized this can't be done in `3.3.x` as adding method and properties is changing the API. I'm going to close this pull request and move the ticket to `3.4.0` but I'll keep a reference to here in redmine so we can find the solution.

arteymix:
To solve that issue, I override Unittest_TestCase and override the Upload class in it.

``` php
class Unittest_TestCase {

}

class Upload extends Kohana_Upload {

    public static function not_empty(array $file)                                                                                                 
    {                                                                                                                                          
        return (isset($file['error']) and isset($file['tmp_name']) and $file['error'] === UPLOAD_ERR_OK);                                         
        // AND is_uploaded_file($file['tmp_name']));                                                                                              
    } 
}
```

I think we could solve this simply by documenting the solution. Adding a flag to disable security enforcment is a really tricky thing.

PR #360, opened 2013-06-05
fixed array to string conversion bug for the log writer

alphanyx:
http://forum.kohanaframework.org/discussion/11567/kohana-3-3-strtr-array-to-string-conversion-issue-in-kohana_log_writer/p1

https://bugs.php.net/bug.php?id=63893

http://php.net/manual/en/migration54.incompatible.php

On Servers with php 5.4 running, the error log function throws an Exception with the notice that some parts of the message (trace from the exception) can`t be castet to string.

With this Update from the First Forum link the bug will be fixed

### DIFF
--- a/classes/Kohana/Log/Writer.php
+++ b/classes/Kohana/Log/Writer.php
@@ -12,21 +12,21 @@ abstract class Kohana_Log_Writer {
 
 	/**
 	 * @var  string  timestamp format for log entries.
-	 * 
+	 *
 	 * Defaults to Date::$timestamp_format
 	 */
 	public static $timestamp;
 
 	/**
 	 * @var  string  timezone for log entries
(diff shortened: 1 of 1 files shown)


evanpurkhiser:
You're going to want to read the [contributing.md](https://github.com/kohana/kohana/blob/3.3/master/CONTRIBUTING.md) document.

Also, I'm fairly sure this was already taken care of?

See 82b470b2827470da37b0e6771b77c369c3d2e5fb and issue 4712 (Looks like the Kohana issue tracker is offline right now though)

zeelot:
Yeah, this was fixed in `3.3/develop`. Also, I just got the issue tracker back up (sorry).

Let me know if I missed anything.

PR #363, opened 2013-06-13
Fixes #4637: Replace iconv with mb_convert_encoding for beter support in...

rjd22:
... PHP 5.4.3+

http://dev.kohanaframework.org/issues/4637


shadowhand:
This change did not include the additional flag to make encoding skip invalid characters:

`ini_set('mbstring.substitute_character', 'none');`

Are you sure this is a compatible change?

sergeyklay:
@shadowhand  +1

rjd22:
Shadowhand We did that change in bootstrap: https://github.com/kohana/kohana/pull/12

PR #366, opened 2013-06-14
New feature from #4753

fleximus:
Provide full access to curl_getinfo() in class Kohana_Request_Client_Curl


enov:
Personally, I find this interesting, but it should target 3.4/develop branch and not 3.3/master. Sorry it took us too long to reply.

PR #368, opened 2013-07-02
Added function Arr::filter()

danieliser:
Similar to extract except that it doesnt return keys that arnt set in the first array.

EX.. 
Arr::filter(
  array('a' => 'foo', 'b' => 'bar'),
  array('a','c')
);
Returns array('a' => 'foo')
Key b is removed from results altogether.

This is particularly useful for extracting fields from $_POST for Model without leaveing unset values as NULL.
Regular extract in this situation will end up saving value "C" above as NULL even if your only attempting to change "A".

### DIFF
--- a/classes/Kohana/Arr.php
+++ b/classes/Kohana/Arr.php
@@ -303,6 +303,29 @@ class Kohana_Arr {
 
 		return $found;
 	}
+	
+	/**
+	* Retrieves multiple keys from an array only if they exist.
+	*
+	* // Get the values "username", "password" from $_POST
+	* $auth = Arr::filter($_POST, array('username', 'password'));
+	*
+	* @param array array to filter keys from
+	* @param array list of key names to search for
+	* @return array
+	*/
+	public static function filter($array, array $keys)
+	{
+		$found = array();
+		foreach ($keys as $key)
+		{
+			if(isset($array[$key]))
+			{
+				$found[$key] = $array[$key];
+			}
+		}
+		return $found;
+	}
 
 	/**
 	 * Retrieves muliple single-key values from a list of arrays.


zeelot:
Few things:
1. `filter` made me thing of http://php.net/manual/en/function.array-filter.php which this is nothing like. Maybe find a better name?
2. This is an API change and needs to go in `3.4.x`. (closing this pull request but feel free to open a new one for `3.4/develop`)
3. Could you file a ticket on redmine for this so we can track it? (http://dev.kohanaframework.org/projects/kohana3/issues)

danieliser:
any thoughts on a better name?

Will submit ticket after finding suitable name.

How about Arr::retrieve()? Sounds a bit like it should be the same as Arr::get() though.

zeelot:
What about `extract_existing()` or `extract_available()`?

danieliser:
Might it not be better to just modify the Arr::extract() function with an additional parameter to switch between default values for non existent values or leaving them out.

What about pull() or get_fields() or similar?

zeelot:
> Might it not be better to just modify the Arr::extract() function with an additional parameter to switch between default values for non existent values or leaving them out.

I prefer splitting functionality out into multiple similar methods instead of having an if statement (or two, or three) in a single method. It's easier to test and clearer in code (a method call like `Some_Class::some_method(FALSE)` is very confusing).

> What about pull() or get_fields() or similar?

All of your suggestions make sense, but since this is a method that is almost identical to `extract()` I feel like we should make it clear in the name. Personally, I'm not fond of the `extract` name, but changing that is probably more trouble than it's worth, so we should just pick a name that clearly indicates the relationship between the two methods.

danieliser:
Point taken.. Will go with extract_existing() since its just a variation of the extract function.

PR #369, opened 2013-07-04
Added function Arr::extract_existing()

danieliser:
Similar to extract except that it doesnt return keys that arnt set in the first array.

EX..
Arr::filter(
array('a' => 'foo', 'b' => 'bar'),
array('a','c')
);
Returns array('a' => 'foo')
Key b is removed from results altogether.

This is particularly useful for extracting fields from $_POST for Model without leaveing unset values as NULL.
Regular extract in this situation will end up saving value "C" above as NULL even if your only attempting to change "A".

### DIFF
--- a/classes/Kohana/Arr.php
+++ b/classes/Kohana/Arr.php
@@ -303,7 +303,31 @@ class Kohana_Arr {
 
 		return $found;
 	}
-
+	
+	/**
+	* Retrieves multiple keys from an array only if they exist.
+	*
+	* // Get the values "a", "b" from $arr = array('a' => 'foo', 'b' => 'bar', c => 'car')
+	* $result = Arr::extract_existing($arr, array('a', 'b'));
+	* returns array('array('a' => 'foo', 'b' => 'bar')
+	* 
+	* @param array array to filter keys from
+	* @param array list of key names to search for
+	* @return array
+	*/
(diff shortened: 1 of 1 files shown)


zeelot:
Did you ever make a redmine ticket for this?

cs278:
Bit of a mouthful, Underscore calls this [pick](http://underscorejs.org/#pick).

You could rewrite the method to:

``` php
array_intersect_key($arr, array_flip($keys))
```

But this won't work on objects that implement `\Traversable` and `\ArrayAccess`, and it'll handle nulls differently.

enov:
Thanks for this @danieliser. No need for a Redmine ticket anymore, but definitely unit tests :)

enov:
Also, I think the method name needs to change, but I am not sure.

acoulton:
This would be really useful - I agree with @enov and @cs278, `pick` might be a better name. 

Though if this were a blank sheet, I'd call this `extract` and have called the existing method something else, since it adds extra values to the result....

danieliser:
Agreed, i thought of several names since. 
`pick`, `fetch`, `only`, `pull`.

Glad to see this finally make it into core. Hell glad to see that project isn't dead. Felt like community died shortly after i started using Kohana heavily.

neo22s:
I like this, missing tests, should be for 4.0.0?

PR #370, opened 2013-07-04
Update Arr.php - allow array of keys to be passed as $path argument

roomcays:
To make Arr::set_path() consistent with Arr::path() allow $path parameter to be set directly as array, in addition to delimiter-separated string value.


melvinmt:
This is already fixed in the 3.3/develop branch: https://github.com/kohana/core/commit/aa08daf2f2ebdae42e210de882a70ef0a2b5faee

zeelot:
If it's not in `3.4` yet, it will be as soon as I merge the `3.3` branches up. Thanks for the PR!

PR #371, opened 2013-07-10
refs #4723 - FIX on URI generation example

maddes:
Basically, confusing uri() function from Route and Request.
`$this->request->uri()` is not the same as `$this->request->route()->uri()`.
Neither `Request::current()->uri` the same as `Request::current()->route()->uri`.

Also, Request::instance() was deprecated in favor of Request::current()


zeelot:
This isn't fixing the problem completely. Lets just remove the paragraph instead (left you details on IRC).

zeelot:
Closing since you made the other change already.

maddes:
Yeap, forgot to close. Thanks!

PR #372, opened 2013-07-10
refs #4723 - Chopping off bad practice example.

maddes:
Zeelot: "The recommended way is to just build the full url from a route without recycling the one in Request.
you don't know which route the user got to your page with, trying to recycle parameters from it is asking for trouble"

So, removing the example...


zeelot:
Thanks!

PR #374, opened 2013-07-28
Fix extraneous white space in Validation.php

evanpurkhiser:
I know pull requests are usually supposed to have issue references in redmine, but this was such a minor change I figured hopefully it would be ok =)


sergeyklay:
I've seen a lot of similar things. Why don't fix all then?

evanpurkhiser:
Every little bit helps right?

sergeyklay:
right :smile_cat:

zeelot:
thanks :)

PR #375, opened 2013-08-01
Update Header.php

shanliu:
run in hiphop-php 
pHipHop Fatal error: Uncaught exception 'ErrorException' with message 'Argument 2 passed to ArrayObject::__construct() must be an instance of int, null given' in /var/www/system/classes/Kohana/HTTP/Header.php:299\nStack trace:\n#0 (): Kohana_Core::error_handler()
#1 /var/www/system/classes/Kohana/HTTP/Header.php(299): ArrayObject->__construct()
#2 /var/www/system/classes/Kohana/Request.php(659): Kohana_HTTP_Header->__construct()
#3 /var/www/system/classes/Kohana/Request.php(160): Kohana_Request->__construct()
#4 /var/www/index.php(117): Kohana_Request::factory()
#5 {main}

### DIFF
--- a/classes/Kohana/HTTP/Header.php
+++ b/classes/Kohana/HTTP/Header.php
@@ -295,8 +295,11 @@ class Kohana_HTTP_Header extends ArrayObject {
 		 * HTTP header declarations should be treated as case-insensitive
 		 */
 		$input = array_change_key_case( (array) $input, CASE_LOWER);
-
-		parent::__construct($input, $flags, $iterator_class);
+		/**
+		 * @link https://github.com/facebook/hiphop-php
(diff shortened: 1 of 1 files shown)


dyron:
Seems that HTTP_Header isn't compatible to ArrayObject, so the constructor arguement should be updated to __construct(array $input = array(), int $flags = 0, $iterator_class = 'ArrayIterator') instead of casting to int in the parent call,
And you should send PR to _._/develop branches, never to _._/master.

zeelot:
Yes please send this to a develop branch and please file a ticket on redmine: http://dev.kohanaframework.org/

PR #376, opened 2013-08-01
add useful method rangelength

timmy8823:


zeelot:
This can't go in 3.3 as it is an API change and we only merge into develop branches. The appropriate branch to merge this into would be `3.4/develop`. Could you also file a ticket on redmine? http://dev.kohanaframework.org/

Thanks!

timmy8823:
I have a question. Do you plan to realize the integration with jquery pugins, for example jquery validation for form (http://jqueryvalidation.org/)
I think this feature will be useful and very much in demand

dyron:
As Kohana is a PHP framework, i doubt that jQuery is included by default.

PR #377, opened 2013-08-04
PSR-0 fix class name

WinterSilence:


zeelot:
Fixed in `3.3/develop` already, thanks.

WinterSilence:
i don`t see this:
https://github.com/kohana/core/blob/3.3/develop/classes/Kohana/I18n.php

https://github.com/kohana/core/blob/3.4/develop/classes/Kohana/I18n.php

zeelot:
The change in this PR shows this https://github.com/kohana/core/blob/3.3/develop/classes/Kohana/View.php#L238-L240

WinterSilence:
@Zeelot, sorry, misunderstood, i thought it about http://dev.kohanaframework.org/issues/4780

PR #378, opened 2013-08-05
fix is_array method

WinterSilence:
1. if $value not array not return FALSE. NULL !== FALSE
2. Easy array objects not implement Traversable
   Traversable: Interface to detect if a class is traversable using foreach. 
   ArrayAccess: Interface to provide accessing objects as arrays.


zeelot:
Did you ever look into the things we discussed on IRC?

zeelot:
Closing since this doesn't fix the problem. Feel free to file a redmine ticket and adjust the change to be closer to a real array check :)

Thanks!

PR #379, opened 2013-08-09
Fix typo causing error when a view throws an exception

ghost:


sergeyklay:
Also update please PHPDoc

``` php
    /**
     * Magic method, returns the output of [View::render]
     *
     * @return  string
     *
     * @uses    View::render
     * @uses    Kohana_Exception::_handler
     * @uses    Response::body
     */
```

zeelot:
Already fixed in `3.3/develop`, thanks. (PHPDoc is wrong since we don't throw an exception from the method)

PR #380, opened 2013-08-12
Improved examples in DocBlock.

marklocker:
This pull request adds improved examples which explicitly show how you can use the _readable_text()_ method.

This also clears up an inconsistency in the previous example, reported in the [original feature request](http://dev.kohanaframework.org/issues/4626#note-16).


zeelot:
Thanks!

PR #381, opened 2013-08-16
Detect secure HTTPS request via proxy

Manc:
Other properties already support proxy related attributes (eg. HTTP_X_FORWARDED_FOR) but not to determine whether the initial request was made via a secure HTTPS connection. This is useful if the web server is behind a proxy server.


cs278:
This means any client can fool the server into thinking the request arrived over SSL/TLS.

Manc:
Can this be solved taking Request::$trusted_proxies into account? It's important an application has a way to determine if the connection between client and server/proxy is secure.

cs278:
Yeah you should use the same check as the IP replacement code.

zeelot:
Already made the change with the proxy checks, thanks! (it's in the `3.3/develop branch`)

PR #382, opened 2013-08-19
3.3/feature/4776/composer package definition

acoulton:
As discussed in http://dev.kohanaframework.org/issues/4776, adds a composer.json to the cache module to simplify using composer to manage Kohana module dependencies.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,35 @@
+{
+	"name":        "kohana/core",
+	"description": "Core system classes for the Kohana application framework",
+	"homepage":    "http://kohanaframework.org",
+	"license":     "BSD-3-Clause",
+	"keywords":    ["kohana", "framework"],
+	"authors": [
+		{
+			"name":     "Kohana Team",
+			"email":    "team@kohanaframework.org",
+			"homepage": "http://kohanaframework.org/team",
+			"role":     "developer"
+		}
+	],
+	"support": {
+		"issues":   "http://dev.kohanaframework.org",
+		"forum":    "http://forum.kohanaframework.org",
+		"irc":      "irc://irc.freenode.net/kohana",
+		"source":   "http://github.com/kohana/core"
+	},
+	"require": {
+		"php":        ">=5.3.3"
+	},
+	"suggest": {
+		"ext-http":   "*",
+		"ext-curl":   "*",
+		"ext-mcrypt": "*"
+	},
+	"extra": {
+		"branch-alias": {
+			"dev-3.3/dev":  "3.3.x-dev",
+			"dev-3.4/dev":  "3.4.x-dev"
+		}
+	}
+}


composer.json
```diff
@@ -0,0 +1,35 @@
...
+	},
+	"suggest": {
+		"ext-http":   "*",
+		"ext-curl":   "*",
+		"ext-mcrypt": "*"
+	},
+	"extra": {
+		"branch-alias": {
+			"dev-3.3/dev":  "3.3.x-dev",
+			"dev-3.4/dev":  "3.4.x-dev"
```
zeelot, line 32: Could you explain how these aliases work? The branches you mention aren't in the repo… what is composer doing? The documentation is terribly confusing :(

acoulton, line 32: Sure.

Composer names versions (including unstable, dev, etc) based on tag and branch names. For branch names by default it looks for eg 2.0 or 2.0.x (which it will call version 2.0.x-dev). Bear in mind that unlike a tag, a branch-based version is not immutable - the 2.0.x-dev "version" will be updated to always point to the tip of the 2.0 development branch.

It if doesn't understand a branch name it will create a version "dev-{branch}" - so Kohana repos will have a "dev-3,3/dev", "dev-3.3/master" etc. See http://getcomposer.org/doc/02-libraries.md#branches

That naming is fine if I want to depend on a stable version eg `"kohana/core" : ">=3.3"` because it will always detect your tags as proper named versions. If I want the unstable version, it would still work if I knew to put `"kohana/core": "dev-3.3/dev"` in my composer.json - so long as I'm not using other packages that require >=3.3, as composer can't tell that the dev-3.3/dev meets that requirement.

But if I use the standard syntax of `"kohana/core": ">=3.3@dev"` (or use command line or other config to tell it to prefer dev versions, as you might on a CI server for spotting future API breakages) it won't know where to find the 3.3 development branch so I'll always get the most recent tagged version.

These aliases fix that by telling it that the branch version called dev-3.3/dev is the 3.3.x development branch (and likewise for 3.4).

That's all in-principle - locally it seems to be doing what I expect but I've not used aliases before and they might have to be tweaked in future. Agree that the composer docs are not the clearest in places, but I'm pretty confident I've understood them properly.

Cheers,
Andrew

zeelot, line 32: But our branches are called `3.3/develop`, not `3.3/dev`, does this need to be updated for that?

zeelot, line 32: Shouldn't it be something like`"dev-3.3/develop":  "3.3.x-dev",`?

zeelot, line 32: I read through the composer documentation with fresh eyes today and it seems to match what I was thinking. If I understand your explanation correctly, the alias should be along these lines:

```
      "dev-3.3/develop":  "3.3.x-dev",
      "dev-3.4/develop":  "3.4.x-dev"
```

I'm heading out for a bit today but if you haven't gotten a chance to reply by the time I get back tonight I'm going to merge your PR and make that change myself… hopefully I'm right. Then I'll set up the needed stuff on packagist and we'll be ready for a `3.3.1` release :D

Thanks for all the help!

PR #383, opened 2013-08-24
Fix redirect

WinterSilence:


zeelot:
How is this an issue? The controller shouldn't be trying to render anything once this is called since it results in an exception. Could you elaborate?

WinterSilence:
_confused_ sorry, mistook project

PR #387, opened 2013-08-26
3.2/change/make error view independent

rjd22:
http://dev.kohanaframework.org/issues/3577

### DIFF
--- a/views/kohana/error.php
+++ b/views/kohana/error.php
@@ -48,7 +48,7 @@ function koggle(elem)
 }
 </script>
 <div id="kohana_error">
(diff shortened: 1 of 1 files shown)


zeelot:
Do we really need to worry about this? We use the `Debug` class in here as well as the `Kohana` class… both classes that you could possibly extend and break.

rjd22:
Well I don't really expect that many people will extend those. And if you extend the Kohana class wrong everything breaks and the Debug class is for erm.... debugging so extending that one wrong and you're doomed also.

But not for the HTML class. You don't expect Kohana to fall apart when extending that one the wrong way.

PR #389, opened 2013-08-27
Refs #3916: Deprecating windowed_urls

rjd22:
http://dev.kohanaframework.org/issues/3916

We should delete this in 3.4

### DIFF
--- a/classes/kohana/html.php
+++ b/classes/kohana/html.php
@@ -48,7 +48,7 @@ class Kohana_HTML {
 
 	/**
 	 * @var  boolean  automatically target external URLs to a new window?
-	 * @deprecated deprecated since version 3.2.3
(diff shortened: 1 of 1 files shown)


sergeyklay:
In HTML5, which replaced xHTML 1.1, attribute `target` for tag `<a>` not deprecated, fully supported.
Under this could you explain to me why `$windowed_urls` should be deprecated?

rjd22:
@sergeyklay It's not really that we are deprecating it because we think it's invalid but that it is unneeded to have a global for it. You can still do this trough the attributes. Only there won't be a global setter anymore. 

This will ensure not unwanted behavior is happening and will cut loose all dependencies the function has.

sergeyklay:
@rjd22 Ok. Thanks!

PR #390, opened 2013-08-27
Rearranged sidebar menu

feketegy:


zeelot:
looks like this does the opposite of what I was doing. I want to take thing _out_ of the confusing submenus, not put them back in.

feketegy:
This is just a rearrangement for this PR https://github.com/kohana/userguide/pull/57

Next, when I'm rewriting the docs, it should change. We need to discuss the menu items for the whole docs. How to rearrange them, merging/splitting/adding/removing pages, etc.

PR #397, opened 2013-09-06
add multilang support and update view

WinterSilence:


classes/Kohana/Date.php
```diff
@@ -514,12 +514,12 @@ public static function fuzzy_span($timestamp, $local_timestamp = NULL)
 		if ($timestamp <= $local_timestamp)
 		{
 			// This is in the past
-			return $span.' ago';
+			return __($span).' '.__('ago');
```
birkir: Shouldn't this be done with parameters?
__(':span ago', array(':span' => __($span)));
as some languages will have "ago", before the actual $span.

WinterSilence: what would you suggest?

birkir: http://kohanaframework.org/3.3/guide/kohana/files/i18n#i18n-variables
I wrote an example above.

WinterSilence: ups :) not immediately understand you. make changes now. thanks!

PR #398, opened 2013-09-07
Include detailed exception message when cannot read session data.

laurent22:
When the session data cannot be read, PHP throws an exception, which is caught by Kohana. Kohana then throws a new exception with a generic message "Error reading session data.". So, in between, the real reason why the session cannot be read has been lost.

As a fix, I suggest to simply add the PHP built-in message to the basic Kohana one. That would change this kind of errors (which I recently got in PHPunit):

```
Error reading session data.
Error reading session data.
Error reading session data.
```

to something like this:

```
Error reading session data. session_start(): Cannot send session cookie - headers already sent by (output started at application/tests/ControllerTest.php:2)
Error reading session data. A session had already been started - ignoring session_start()
Error reading session data. A session had already been started - ignoring session_start()
```

### DIFF
--- a/classes/Kohana/Session.php
+++ b/classes/Kohana/Session.php
@@ -321,7 +321,7 @@ abstract class Kohana_Session {
 		catch (Exception $e)
 		{
 			// Error reading the session, usually a corrupt session.
-			throw new Session_Exception('Error reading session data.', NULL, Session_Exception::SESSION_CORRUPT);
+			throw new Session_Exception('Error reading session data. ' . $e->getMessage(), NULL, Session_Exception::SESSION_CORRUPT);
 		}
 
 		if (is_array($data))


dm:
You should issue PRs to the development branch. And it would be best to include an issue report in redmine.

Other than that, +1

evanpurkhiser:
You probably want to use parameters too for consistency with the rest of the framework. So instead of it looking like this

``` php
throw new Session_Exception('Error reading session data. ' . $e->getMessage(), NULL, Session_Exception::SESSION_CORRUPT);
```

It would look like this

``` php
throw new Session_Exception('Error reading session data. :message', array(':message' => $e->getMessage()), Session_Exception::SESSION_CORRUPT);
```

laurent22:
Thanks for the feedback, I will apply the commit to the dev branch and EvanPurkhiser's suggested change.

By the way, I have opened the Redmine issue there: http://dev.kohanaframework.org/issues/4787

ljyf5593:
this bug is existed in version 3.3.2

randomecho:
Closing in favour of if this goes again it would be re-targeted to the develop branch.

PR #400, opened 2013-09-18
3.4/feature/enable seperation word pick text number

rjd22:
Fixes http://dev.kohanaframework.org/issues/4000

### DIFF
--- a/classes/Kohana/Text.php
+++ b/classes/Kohana/Text.php
@@ -513,7 +513,7 @@ class Kohana_Text {
 	 * @return  string
 	 * @since   3.0.8
 	 */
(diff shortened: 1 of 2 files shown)


melvinmt:
I don't think it's a good idea if the separator parameter includes spaces for formattting, it's not what I would expect.

rjd22:
@melvinmt The problem with not doing this will be a loss of control. Some may want to add a comma to the string. Others only a space.

By adding spacing to the separator we're giving the users full control over what is between the 2 numbers.

If you have a better solution of solving this flexibility problem tell me. :)

PR #401, opened 2013-09-26
Only copy original request headers when following redirects [fixes #4790]

acoulton:
When following a redirect, the client should not add NULL headers
from the follow_headers list that were not set on the original
request. This prevents problems when requesting URLs that
include an authorization signature calculated from the request URL
and headers.

Fixes http://dev.kohanaframework.org/issues/4790


zeelot:
Thanks!

acoulton:
Thanks @Zeelot

PR #402, opened 2013-09-26
Don't set empty body on external curl requests [Fixes #4791]

acoulton:
When the CURLOPT_POSTFIELDS option is present, curl adds a
default Content-Type header which can be changed but not
removed, causing authentication problems with signed requests.

The option should only be set if a request body is being sent.

Fixes http://dev.kohanaframework.org/issues/4791 - see issue discussion regarding tests for this issue and fix.


zeelot:
Can you file this under one of the develop branches? `3.2/develop` should be fine. Can't merge into master branches.

PR #403, opened 2013-09-26
Don't set empty body on external curl requests [Fixes #4791]

acoulton:
When the CURLOPT_POSTFIELDS option is present, curl adds a
default Content-Type header which can be changed but not
removed, causing authentication problems with signed requests.

The option should only be set if a request body is being sent.

Fixes http://dev.kohanaframework.org/issues/4791 - see issue discussion regarding tests for this issue and fix.

This is a replacement for #402 which I'd accidentally targeted to 3.3/master instead of 3.3/develop, and I've subsequently rebased back to 3.2/develop.

### DIFF
--- a/classes/kohana/request/client/curl.php
+++ b/classes/kohana/request/client/curl.php
@@ -33,7 +33,10 @@ class Kohana_Request_Client_Curl extends Request_Client_External {
 		// if using a request other than POST. PUT does support this method
 		// and DOES NOT require writing data to disk before putting it, if
 		// reading the PHP docs you may have got that impression. SdF
-		$options[CURLOPT_POSTFIELDS] = $request->body();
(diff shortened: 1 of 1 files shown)


biakaveron:
What method do you use in request? PUT or POST? It seems like only POST will automatically set Content-Type, while S3 requires PUT (but I may be wrong, cause I'm not familiar with it).

acoulton:
@biakaveron the issue I was having was actually with GET requests. Curl was adding the Content-Type header and causing the signature to fail authorisation. 

S3 supports POST as well as PUT afaik, though obviously for S3 I can't think of a case for sending either POST or PUT without a body.

I guess we could change my conditional to be on request type, but it felt to me more obvious that the content-type header only be forced if the request has content.

acoulton:
Thanks @Zeelot

PR #407, opened 2013-10-14
Update Text.php

thinkclay:
/e is deprecated, updating code to reflect better standard. Unit tests are passing with this update.


shadowhand:
I think the proper solution is to backport #517 instead of using this method.

thinkclay:
Yeah probably. I'll kill this one.

PR #408, opened 2013-10-22
Added two attributes to the Form::Select helper for prompt value and bla...

aurimasniekis:
I think this feature is really needed in Kohana :) Added the same functionality as the Ruby on Rails do.


zombor:
This needs to point at the develop branch, and it needs tests.

PR #409, opened 2013-10-22
Added prompt value attribute to the Form::Select helper function

aurimasniekis:
Added prompt value attribute to the Form::Select helper function


classes/Kohana/Form.php
```diff
@@ -332,6 +332,15 @@ public static function select($name, array $options = NULL, $selected = NULL, ar
 					$options[$value] = '<option'.HTML::attributes($option).'>'.HTML::chars($name, FALSE).'</option>';
 				}
 			}
+			
+			// Check if prompt value is set
+                        if ($attributes && $attributes["prompt"]) {
```
evanpurkhiser, line 337: You should use `AND` instead of `&&` here.

zombor:
I'm not sure why a user wouldn't just add this to the options array with `arr::merge()` or something.

evanpurkhiser:
I'm a little confused what 'prompt' means in this context.

Some other things:
- You'll probably need to [open a issue on the redmine tracker](http://dev.kohanaframework.org/projects/kohana/issues) for this. 
- You should get rid of those two travis.yml commits (`git reset --hard 5a6f0fd`)
- The formatting appears to be wonky (tabs / spaces mixup?)

Not sure what's going on with the CI build, @zombor might be able to help you there.

zombor:
We don't use travis for core repos, only the `kohana/kohana` repo.

classes/Kohana/Form.php
aurimasniekis, line 337: Good point :+1:  It's my first push to Kohana :)

aurimasniekis:
I am sorry for travis **\* i thought it will not go the pull request... I was editing with Github editor so maybe this is why it formated  spaces tabes

rjd22:
@gcds Could you please fix the issues and redo the pull request? Also please link the ticket you're fixing or make a new one and link that :)

classes/Kohana/Form.php
alexeyco, line 337: Double quotes - it's wrong way to, I think. You should use single quotes, like `$attributes['prompt']`.

For code indentation - you need to use tabs instead spaces.

See https://github.com/kohana/core/blob/3.3/master/guide/kohana/conventions.md

sergeyklay, line 337:
``` php
if ($attributes AND isset($attributes['prompt']))
```

enov:
@gcds sorry to close. Would you kindly redo this by fixing the code style as suggested here? We're not using the Redmine tracker anymore so no need to create a ticket there. Just PR anew with a better description - I am not sure if _prompt_ is the right word in this context.

PR #410, opened 2013-10-25
Move cookie salt description up.

aaronott:
The cookie salt description belongs inside the bootstrap.php
documentation. Putting this above the filesystem changes and with the
other bootstrap.php changes just makes sense.


enov:
I took care of this in 337d812619b5c8e82abf59dd74e31ed3bbcdffcd. It will be also available for 3.4. Thanks :)

PR #411, opened 2013-10-29
Fixes issue that would render integer 0 as blank

JackEllis:
When passing in a 0 (integer) as a value, HTML::attributes would render it blank because it was detecting it as false. With this little change, it'd be looking for a !== match.

### DIFF
--- a/classes/Kohana/HTML.php
+++ b/classes/Kohana/HTML.php
@@ -332,7 +332,7 @@ class Kohana_HTML {
 			// Add the attribute key
 			$compiled .= ' '.$key;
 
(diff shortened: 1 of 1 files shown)


zombor:
This needs tests and should be pointed at develop branch.

rjd22:
Hey @JackEllis 

Thank you for you pull request. Could you please do the following things?:
- Please link the ticket you're fixing. Else create on and link that.
- Add some tests to your fix so you don't break any other things.
- Please request a pull on kohana:3.3/develop master is our stable branch

PR #412, opened 2013-11-07
Replaced '://' with '//' for SDN URL's

maddes:
The current check doesn't handle propperly SDN addresses, for example:

```
HTML::script('//ajax.googleapis.com/ajax/libs/jquery/1.10.2/jquery.min.js');
```

### DIFF
--- a/classes/Kohana/HTML.php
+++ b/classes/Kohana/HTML.php
@@ -118,7 +118,7 @@ class Kohana_HTML {
 		}
 		else
 		{
-			if (strpos($uri, '://') !== FALSE)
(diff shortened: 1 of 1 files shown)


randomecho:
You mean CDN?

On contributing: 
- Target the develop branch
- Include the [Redmine issue number](http://dev.kohanaframework.org/projects/kohana/issues) that goes with it
- Don't forget your tests

zombor:
We've decided in the past not to support this feature. In this case, it's just better to write the html src by hand. But yes, read the contributing guide: https://github.com/kohana/kohana/blob/3.3/master/CONTRIBUTING.md

PR #417, opened 2013-12-22
Added convenience methods for request type checking

marklocker:
Added convenience methods for the four most common request types.

Related issue: http://dev.kohanaframework.org/issues/4606

### DIFF
--- a/classes/Kohana/Request.php
+++ b/classes/Kohana/Request.php
@@ -1328,4 +1328,72 @@ class Kohana_Request implements HTTP_Request {
 		return $this;
 	}
 
+	/**
+	 * Check to see if the current request is a POST request.
+	 *
+	 * @return bool Whether the request is a POST request or not.
+	 */
+	public function is_post()
+	{
+		if ($this->method() === HTTP_Request::POST)
+		{
+			return TRUE;
+		}
+		else
+		{
+			return FALSE;
+		}
+	}
+
+	/**
+	 * Check to see if the current request is a GET request.
+	 *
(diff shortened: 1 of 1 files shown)


classes/Kohana/Request.php
```diff
@@ -1328,4 +1328,72 @@ public function post($key = NULL, $value = NULL)
...
+	public function is_post()
+	{
+		if ($this->method() === HTTP_Request::POST)
+		{
+			return TRUE;
+		}
+		else
+		{
+			return FALSE;
+		}
```
sergeyklay:
``` php
return ($this->method() === HTTP_Request::POST)
```

birkir:
``` php
return ($this->_method === HTTP_Request::POST)
```

sergeyklay: @birkir yep, you're right, but canonical

``` php
return (HTTP_Request::POST === $this->_method)
```

marklocker:
Quite right - a little too concerned with adhering to Kohana's "braces on new lines" guideline I think!

classes/Kohana/Request.php
evanpurkhiser: Are the parentheses needed here?

**Edit:** Nevermind, I see that this is the style follow through the rest of the file.

zeelot:
There seem to be merge conflicts here, could you merge the latest `3.4/develop` into your branch? Also, can we add unit tests for this? It should be simple enough.

neo22s:
I like this for 4.0.0 but needs to be done manually.

PR #418, opened 2013-12-24
Refs #4587: Request::redirect calls URL::site using the wrong parameters

enov:
URL::site and consequently URL::base can take care of
the default protocol and index file


zeelot:
Closing based on comment in ticket :) let me know if I missed something and we can reopen it.

PR #419, opened 2013-12-26
Updated hello world to help beginners get past Kohana basic setup.

miket3:
For new users setting up Kohana in a local environment, a missing .htaccess file can become a road block for getting past Hello World example. Just because the Hello World example works, it does not mean the localhost environment is setup sufficiently to move on to real world examples. Therefore a beginner might assume proper environment setup because Hello World works, and waste time investigating other reasons for a real world example failing such as posting to a controller from a web page.


randomecho:
You'll need to target **develop** branch. No pull requests are done against **master**.

zeelot:
Right, please file PRs like this to our develop branch (`3.4/develop` or `3.3/develop`).

PR #423, opened 2014-01-09
Refs #3461: Do not execute action if Controller::before() returns FALSE

tmihalik:
Do not execute action if Controller::before() returns FALSE

http://dev.kohanaframework.org/issues/3461


zeelot:
That ticket was closed as "wontfix" with explanations for why we don't want to support this. Feel free to comment on redmine if you want to make a case for it to be changed.

PR #424, opened 2014-01-09
Ref #4802: HTTP_Header not compatible with ArrayObject

enov:
HHVM complains about this. The Zend engine silently ignore the bug, probably casting the NULL internally to int 0.

Could not create a test for it, as this works with Zend, but it's fatal in HHVM.


dm:
Yup, good catch!

The $flags param should be int :thumbsup:

PR #306, opened 2012-10-21
URL encode all parameters to Route::uri

jimktrains:

PR #427, opened 2014-01-11
#4079 - Route::uri() Should encode parameters

JackEllis:
Route::uri() does not currently handle special characters very well.

route.php modification was copied from @jimktrains (#306), but I've used rawurlencode to achieve OP's expected result.

See: http://dev.kohanaframework.org/issues/4079

### DIFF
--- a/classes/kohana/route.php
+++ b/classes/kohana/route.php
@@ -456,7 +456,7 @@ class Kohana_Route {
 	 */
 	public function uri(array $params = NULL)
 	{
-		if (is_array($params))
+		if ($params)
 		{
 			$params = array_map('rawurlencode', $params);
 		}
--- a/classes/kohana/route.php
+++ b/classes/kohana/route.php
@@ -456,6 +456,11 @@ class Kohana_Route {
 	 */
 	public function uri(array $params = NULL)
 	{
+		if (is_array($params))
+		{
+			$params = array_map('rawurlencode', $params);
+		}
+
 		// Start with the routed URI
 		$uri = $this->_uri;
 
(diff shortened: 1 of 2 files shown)


classes/kohana/route.php
```diff
@@ -456,6 +456,11 @@ public function is_external()
 	 */
 	public function uri(array $params = NULL)
 	{
+		if (is_array($params))
```
Yahasana: `if ($params !== NULL)`

zeelot: that or even just `if ($params)`. No need to check for an array specifically.

tests/kohana/RouteTest.php
```diff
@@ -715,4 +715,41 @@ public function test_external_route_includes_params_in_uri($route, $defaults, $e
...
+	 * @test
+	 * @covers Route::get
+	 * @ticket 4079
+	 * @dataProvider provider_route_uri_encode_parameters
+	 */
+	public function test_route_uri_encode_parameters($name, $uri_callback, $defaults, $expected)
+	{
+		Route::set($name, $uri_callback)->defaults($defaults);
+
+		$get_route_uri = Route::get($name)->uri(array('article_name' => 'Article name with special chars \\ ##'));
```
zeelot: Could you move the key and value into the data provider as well? This way we can test with multiple versions in the future.

JackEllis:
@Zeelot Sorted ^

enov:
Hey @JackEllis ,

This is conflicting with the userguide.

Could you try to re-patch this by rawurlencoding everything except slash and backslash?

See more in the bug tracker:

http://dev.kohanaframework.org/issues/4079

PR #429, opened 2014-01-12
Bug 3373: Avoid division by zero

enov:
This bug is hardly reproducible.

http://dev.kohanaframework.org/issues/3373

### DIFF
--- a/views/profiler/stats.php
+++ b/views/profiler/stats.php
@@ -35,7 +35,7 @@ $application_cols = array('min', 'max', 'average', 'current');
 				<div>
 					<div class="value"><?php echo number_format($stats[$key]['time'], 6) ?> <abbr title="seconds">s</abbr></div>
 					<?php if ($key === 'total'): ?>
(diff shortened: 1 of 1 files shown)


views/profiler/stats.php
```diff
@@ -47,7 +47,7 @@
 				<div>
 					<div class="value"><?php echo number_format($stats[$key]['memory'] / 1024, 4) ?> <abbr title="kilobyte">kB</abbr></div>
 					<?php if ($key === 'total'): ?>
-						<div class="graph" style="left: <?php echo max(0, 100 - $stats[$key]['memory'] / $group_stats[$group]['max']['memory'] * 100) ?>%"></div>
+						<div class="graph" style="left: <?php echo $group_stats[$group]['max']['memory'] ? max(0, 100 - $stats[$key]['memory'] / $group_stats[$group]['max']['memory'] * 100) : '' ?>%"></div>
```
zeelot: Do these need to default to 0? seems like the code is being put into a style tag so the default of the empty string wouldn't be valid, right?

enov: Well, either 0, or 100, or n/a... Whatever someone translates 0/0... I choose n/a :)

PR #434, opened 2014-01-13
#4801 - update model guide

WinterSilence:


rjd22:
Sadly this is incorrect. In web frameworks models don't directly notify the controller and the view. Can you please rewrite this to be applied to web MVC frameworks?

WinterSilence:
Help, if you can, I don't know English very well

PR #436, opened 2014-01-13
Request::process_uri() must ignore external routes

biakaveron:
Pull request for http://dev.kohanaframework.org/issues/4235. External routes can be used for reverse routing only.

PR #437, opened 2014-01-21
Bug 4235 exclude external routes2

biakaveron:
Pull request for http://dev.kohanaframework.org/issues/4235. External routes can be used for reverse routing only. (clean version of https://github.com/kohana/core/pull/436)


zeelot:
Thanks!

PR #439, opened 2014-01-22
Update core.php

goper:
$var = true and false;
// $var = true;


goper:
if i'm right send me a message to goper@tut.by

feketegy:
This is already fixed here https://github.com/kohana/core/pull/438
Issue: http://dev.kohanaframework.org/issues/4810

zeelot:
Alright, closing then. Thanks!

PR #442, opened 2014-01-23
Fixes #3499 widont with html tags

enov:
with code courtesy Typogrify project:
http://code.google.com/p/typogrify/


zeelot:
Thanks!

PR #445, opened 2014-01-27
Fixes #3604: default to ini setting if Cookie::$domain not available

enov:
We are using Cookie parameters for Session Native with session_set_cookie_params.

session_set_cookie_params overrides `session.cookie_domain` ini setting.

When the developer specifies `session.cookie_domain` ini setting and keeps Cookie::$domain NULL, PHP defaults to server name (which is not what the developer wants).

From PHP manual:

> Default is none at all meaning the host name of the server which generated the cookie

http://www.php.net/manual/en/session.configuration.php#ini.session.cookie-domain

### DIFF
--- a/classes/kohana/session/native.php
+++ b/classes/kohana/session/native.php
@@ -31,7 +31,7 @@ class Kohana_Session_Native extends Session {
 		// see issue #3604
 		//
 		// set to Cookie::$domain if available, otherwise default to ini setting, 
-		$session_cookie_domain = empty(Cookie::$domain) ?
+		$session_cookie_domain = Cookie::$domain ?
 			Cookie::$domain :
 			ini_get('session.cookie_domain');
 		
--- a/classes/kohana/session/native.php
+++ b/classes/kohana/session/native.php
@@ -30,11 +30,11 @@ class Kohana_Session_Native extends Session {
 		//
 		// see issue #3604
 		//
-		// set to Cookie::$domain if available, otherwise default to ini setting, 
-		$session_cookie_domain = Cookie::$domain
-		    ? Cookie::$domain
-		    : ini_get('session.cookie_domain');
-		
+		// set to Cookie::$domain if available, otherwise default to ini setting
+		$session_cookie_domain = empty(Cookie::$domain)
+		    ? ini_get('session.cookie_domain')
+		    : Cookie::$domain;
+
 		// Sync up the session cookie with Cookie parameters
 		session_set_cookie_params(
 			$this->_lifetime,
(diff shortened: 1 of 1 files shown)


classes/kohana/session/native.php
```diff
@@ -25,7 +25,14 @@ public function id()
 	protected function _read($id = NULL)
 	{
 		// Sync up the session cookie with Cookie parameters
-		session_set_cookie_params($this->_lifetime, Cookie::$path, Cookie::$domain, Cookie::$secure, Cookie::$httponly);
+		session_set_cookie_params(
+			$this->_lifetime,
+			Cookie::$path,
+			// set to Cookie::$domain if available, otherwise default to ini setting, see issue #3604
+			Cookie::$domain ? : ini_get('session.cookie_domain'),
```
loonies: > Since PHP 5.3, it is possible to leave out the middle part of the ternary operator. Expression expr1 ?: expr3 returns expr1 if expr1 evaluates to TRUE, and expr3 otherwise.

http://php.net/manual/en/language.operators.comparison.php

Kohana 3.2 doesn't require PHP 5.3.

It would be also better to create a variable and pass into function as argument (comment as well)..

enov: Alright, I will update the PR.

loonies:
Please fix ternaries as per coding standard http://kohanaframework.org/3.2/guide/kohana/conventions#ternaries

loonies:
btw, you can squash commits with rebase and force push before this get merged into develop, so we get one nice commit

enov:
I think it is compliant with the coding standard, except for the spaces (instead of tabs). Is that what you are referring to?

> you can squash commits with rebase and force push

Not a git guru here... I will see what I can do  :)

loonies:
Should read

``` php
$session_cookie_domain = empty(Cookie::$domain)
    ? Cookie::$domain
    : ini_get('session.cookie_domain');
```

enov:
Yeah... I see now... Thanks!

enov:
@loonies would you mind to check again before I squash/rebase/push/whatever-git-fu and have one nice commit?

loonies:
hm...

I wasn't paying attention, but shouldn't that line read

``` php
$session_cookie_domain = empty(Cookie::$domain)
    ? ini_get('session.cookie_domain')
    : Cookie::$domain;
```

or

``` php
$session_cookie_domain = (Cookie::$domain === NULL)
    ? ini_get('session.cookie_domain')
    : Cookie::$domain;
```

?

enov:
> hm...

Is that a :+1: or a :-1: ?

loonies:
sorry, posted too early by accident

enov:
It seems that if I rebase, I need to issue a new PR, and our discussion will be lost.

http://stackoverflow.com/questions/7293271/rebase-on-github-branch

evanpurkhiser:
That SO answer is a little over two and a half years old. I'm pretty sure you can force push a branch with a pending pull request and it will update the PR.

zeelot:
closing in favor of #447

PR #448, opened 2014-01-29
Fix typo and small bugs, remove Kohana::version

WinterSilence:
Bugs:

https://github.com/kohana/core/blob/3.4/develop/classes/Kohana/Core.php#L255
https://github.com/kohana/core/blob/3.4/develop/classes/Kohana/Core.php#L258

```
mkdir($settings['cache_dir'], 0755, TRUE);
```

```
chmod($settings['cache_dir'], 0755);
```

low chmod

https://github.com/kohana/core/blob/3.4/develop/classes/Kohana/Core.php#L442

```
if (is_array($value) OR is_object($value))
```

incorrect condition: is_object($value)

https://github.com/kohana/core/blob/3.4/develop/classes/Kohana/Core.php#L1025

```
if (Kohana::$errors AND $error = error_get_last() AND in_array($error['type'], Kohana::$shutdown_errors))
```

incorrect condition: error_get_last() return array or null

https://github.com/kohana/core/blob/3.4/develop/classes/Kohana/Core.php#L1028

```
ob_get_level() AND ob_clean();
```

incorrect condition: ob_get_level()  return integer

Delete Kohana::version(): this method is not used.

### DIFF
--- a/classes/Kohana/Core.php
+++ b/classes/Kohana/Core.php
@@ -10,14 +10,14 @@
  * @package    Kohana
  * @category   Base
  * @author     Kohana Team
- * @copyright  (c) 2008-2012 Kohana Team
+ * @copyright  (c) 2008-2014 Kohana Team
  * @license    http://kohanaframework.org/license
  */
 class Kohana_Core {
 
 	// Release version and codename
-	const VERSION  = '3.3.0';
-	const CODENAME = 'badius';
+	const VERSION  = '3.3.1';
+	const CODENAME = 'peregrinus';
 
 	// Common environment type constants for consistency and convenience
 	const PRODUCTION  = 10;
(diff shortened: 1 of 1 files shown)


classes/Kohana/Core.php
```diff
@@ -252,15 +252,17 @@ public static function init(array $settings = NULL)
 				try
 				{
 					// Create the cache directory
-					mkdir($settings['cache_dir'], 0755, TRUE);
+					mkdir($settings['cache_dir'], 0777, TRUE);
```
KostyaTretyak, line 255: @WinterSilence, why it is called a bug? I have the cache works well with "755"

WinterSilence:
@KostyaTretyak This forestalls potential problems: cache files use chmod 0777, but folders have chmod 0755. 
TODO: Optimally use default rules: for files 644 (-rw-r--r--), for folders 755 (drwxr-xr-x).

PR #449, opened 2014-01-29
Fixed typo in comments for Kohana_View::__toString()

robbytaylor:
I haven't opened a ticket since it's such a small comment change...

### DIFF
--- a/classes/Kohana/View.php
+++ b/classes/Kohana/View.php
@@ -232,7 +232,7 @@ class Kohana_View {
 			/**
 			 * Display the exception message.
 			 *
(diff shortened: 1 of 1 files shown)


evanpurkhiser:
Is this actually a typo?

> We use this method here because it's impossible to throw an exception from __toString().

This sounds right to me.

zombor:
`and` vs `an`

zombor:
Should this target 3.3/develop? Is the typo there as well?

robbytaylor:
I noticed it while looking through the View class in 3.3/master, so yeah, you're right, it should target 3.3/develop.

PR #452, opened 2014-02-05
Fixes #3499: widont with html tags

enov:
This PR will basically fix https://github.com/kohana/core/pull/442
where the regex was without a delimiter and modifier


enov:
I see that my commit from PR kohana/core/3.2/develop to enov/core/3.2/develop appears here.
But the "Files Changed" section above correctly shows only 2 files changed.

Is this OK? Or should I do something about it?

zeelot:
That's fine, thanks!

PR #453, opened 2014-02-07
Update HTML.php

fabiotnt:
To support less files

### DIFF
--- a/classes/Kohana/HTML.php
+++ b/classes/Kohana/HTML.php
@@ -219,7 +219,7 @@ class Kohana_HTML {
 		$attributes['rel'] = empty($attributes['rel']) ? 'stylesheet' : $attributes['rel'];
 
 		// Set the stylesheet type
-		$attributes['type'] = 'text/css';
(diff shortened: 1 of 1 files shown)


sergeyklay:
The [HTML5 spec](http://dev.w3.org/html5/spec/Overview.html#the-link-element) says that the type attribute is purely advisory and explains in detail how browsers should act if it's omitted.

It seems `$attributes['type']` should be removed. If the user really needs it - it could specify it explicitly

shadowhand:
@sergeyklay agree, we should be transitioning towards HTML5, but as it stands, we still target XHTML, which says that `type` is recommended for stylesheets.

enov:
@fabiotnt could you kindly point this towards 3.3/develop? Thanks!

enov:
BTW, I also noticed that the `type` attribute is recommended to be `text/css` similar to the standard CSS `link` tag. See [lesscss.org](http://lesscss.org):

> `<link rel="stylesheet/less" type="text/css" href="styles.less" />`

So @shadowhand, would you kindly let me know if we still need to have this?

PR #454, opened 2014-02-11
3.2/develop

Asenar:
theses 2 commits remove, in comments, the mention of Request::instance(), which does not exists in 3.2


shadowhand:
That's a lot more than 2 commits, I think you targeted the wrong branch.

Asenar:
yes sorry, I just noticed

PR #455, opened 2014-02-11
Fix documentation mention of Request::instance

Asenar:
Various place in guide and php comments were mentionning
Request::instance() to be used. I replaced it either by
Request::factory() or something more appropriate.


zeelot:
thanks!

PR #456, opened 2014-02-12
Fixes #4819 replace preg replace with /e preg replace callback

enov:
contains an ugly hack to support both PHP 5.2 and PHP 5.5:
used a temporary static protected variable to "pass" argument
from function `censor` to `_censor_preg_replace_callback`

### DIFF
--- a/classes/kohana/text.php
+++ b/classes/kohana/text.php
@@ -48,6 +48,17 @@ class Kohana_Text {
 	);
 
 	/**
+	 * Temporarily stores the string to replace badwords while censoring
+	 * This is part of an ugly hack to support both PHP 5.2 and PHP 5.5
+	 * Used in the functions censor and _censor_preg_replace_callback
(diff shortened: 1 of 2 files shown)


shadowhand:
Why not just use an `if (version_compare(...))` fork to handle both situations? This seems like an overly complicated fix.

enov:
Hmmm... very interesting, I did not thought of that.

That would definitely create a more elegant solution. Will issue a new PR soon.

Thanks @shadowhand .

enov:
I believe this is the one that should go.

If you don't like the protected static temp variable, we can have a class under `Kohana_Text` with a `$replacement` property. We can instantiate that class with `$replacement` from `censure` and have a method in there for callback, where we can definitely call `$this->replacement`.

enov:
Fixed with `create_function` in https://github.com/kohana/core/pull/464

Thanks @Zeelot

enov:
Closed in favor of #465

PR #457, opened 2014-02-12
Fixes #4819 replace `preg_replace` with /e modifier with preg_replace_callback

enov:
This PR amends https://github.com/kohana/core/pull/456 with the suggestion by @shadowhand

### DIFF
--- a/classes/kohana/text.php
+++ b/classes/kohana/text.php
@@ -235,8 +235,8 @@ class Kohana_Text {
 	/**
 	 * Uppercase words that are not separated by spaces, using a custom
 	 * delimiter or the default.
-	 * 
-	 *      $str = Text::ucfirst('content-type'); // returns "Content-Type" 
+	 *
+	 *      $str = Text::ucfirst('content-type'); // returns "Content-Type"
 	 *
 	 * @param   string  $string     string to transform
 	 * @param   string  $delimiter  delimiter to use
@@ -269,6 +269,9 @@ class Kohana_Text {
 	 *         'frick' => '#####',
 	 *     ));
 	 *
+	 * If argument $replacement is a single character, it will be used to replace
+	 * the characters in the badword, otherwise it will replace the badword completely
+	 *
 	 * @param   string  $str                    phrase to replace words in
 	 * @param   array   $badwords               words to replace
 	 * @param   string  $replacement            replacement string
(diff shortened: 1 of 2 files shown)


classes/kohana/text.php
```diff
@@ -235,8 +235,8 @@ public static function random($type = NULL, $length = 8)
 	/**
 	 * Uppercase words that are not separated by spaces, using a custom
 	 * delimiter or the default.
-	 * 
-	 *      $str = Text::ucfirst('content-type'); // returns "Content-Type" 
+	 *
+	 *      $str = Text::ucfirst('content-type'); // returns "Content-Type"
```
shadowhand: this bit looks like (unnecessary) white space change, perhaps tabs?

utf8/ucwords.php
```diff
@@ -15,9 +16,19 @@ function _ucwords($str)
 
 	// [\x0c\x09\x0b\x0a\x0d\x20] matches form feeds, horizontal tabs, vertical tabs, linefeeds and carriage returns.
 	// This corresponds to the definition of a 'word' defined at http://php.net/ucwords
-	return preg_replace(
-		'/(?<=^|[\x0c\x09\x0b\x0a\x0d\x20])[^\x0c\x09\x0b\x0a\x0d\x20]/ue',
-		'UTF8::strtoupper(\'$0\')',
+	return preg_replace_callback(
```
shadowhand, line 19: wouldn't this break in php < 5.5?

classes/kohana/text.php
enov: That's courtesy of my IDE

enov: I think it is removing spaces from end of lines

utf8/ucwords.php
enov, line 19: No, I don't think so. preg_replace_callback seems to exist long time ago. I am on PHP 5.4 and tests are running fine.

enov:
@shadowhand ,

I don't think `if (version_compare(...))` will work.

See this SO question:

http://stackoverflow.com/questions/14036174/variant-anonymous-function-definitions-based-on-php-version

classes/kohana/text.php
evanpurkhiser: When you stage your changes in git you may want to use `git add -p` so you can selectively pick only the relative changes to the PR =)

enov: @EvanPurkhiser thanks for being around and sharing your notes. Forgive my ignorance. I will search about and see what that `git add -p` will do for me :+1: .

enov: @EvanPurkhiser Yeah I see now what it does, and already used it while staging the changes for the PR that followed this one. Thanks.

classes/kohana/text.php
```diff
@@ -293,12 +296,25 @@ public static function censor($str, $badwords, $replacement = '#', $replace_part
 
 		$regex = '!'.$regex.'!ui';
 
+		// if $replacement is a single character: replace each of the characters of the badword with $replacement
 		if (UTF8::strlen($replacement) == 1)
 		{
-			$regex .= 'e';
-			return preg_replace($regex, 'str_repeat($replacement, UTF8::strlen(\'$1\'))', $str);
+			// issue #4819: preg_replace_callback mainly for 5.5.0, where preg_replace with /e modifier is depricated
+			if (version_compare(PHP_VERSION, '5.5.0') >= 0)
```
cs278, line 303: Why not use this for PHP ≥ 5.3?

enov, line 303: Hey @cs278 ,

Yeah, we could have used `5.3` in there. As I commented on the commit message of 2861ac6, I just _felt_ that putting `5.5` there seemed faster while running the tests. Also, it makes it more readable to put `5.5` since the bugfix is for PHP 5.5.

In ALL cases, this PR should be closed, because even if the code is forked with an `if` condition, it breaks for PHP 5.2 with E_PARSE error.

enov:
I am favoring PR #456 over this one, as this will E_PARSE error on PHP 5.2

PR #460, opened 2014-02-14
Fixes #3931: allow Kohana shutdown properly, exit(1) in case of exception

enov:


zeelot:
This looks ok to me… seemed to work. Will not merge this into 3.3+ since we have a different fix in there that involves a better API.

enov:
Hey @Zeelot , would you mind also closing #4275 and #4794 on Redmine? They depended on this bug and this PR would fix them too.

zeelot:
Done, thanks.

classes/kohana/core.php
```diff
@@ -967,11 +967,24 @@ public static function error_handler($code, $error, $file = NULL, $line = NULL)
...
+	 * on the second visit.
+	 * @see issue #3931
 	 *
 	 * @return  void
 	 */
 	public static function shutdown_handler()
 	{
+		// test if we have already shutdown, if TRUE, exit(1)
+		static $shutdown_handler_visited = FALSE;
+		if ($shutdown_handler_visited) {
```
evanpurkhiser, line 983: Not to be too picky, but the brace here should be on the next line.

enov, line 983: @EvanPurkhiser it seems I need to PR anew?

zeelot, line 983: yeah if you want to :) I don't actually care all that much.

enov, line 983: @Zeelot: voilà https://github.com/kohana/core/pull/463

PR #461, opened 2014-02-14
Fixes #3931: Allow Kohana to shutdown properly and then exit(1) in case of exception

enov:


zeelot:
Adding parameters, even if optional, is still breaking the API :(

I think I'm ok with globals like @shadowhand mentioned since this is an old version and we aren't really looking for a clean architecture… it's much more important that we don't break anyone's app when they bring in a bugfix release.

zeelot:
This one can't be done in a minor version so I'm closing it. Looking at the other one now.

PR #462, opened 2014-02-19
Update Encrypt.php

sarciszewski:
Read this please:
http://phpsecurity.readthedocs.org/en/latest/Insufficient-Entropy-For-Random-Values.html

Also, this should probably be wrapped inside of an authentication mechanism (Encrypt-then-HMAC) if it isn't already employed elsewhere.

### DIFF
--- a/classes/Kohana/Encrypt.php
+++ b/classes/Kohana/Encrypt.php
@@ -151,7 +151,7 @@ class Kohana_Encrypt {
 				}
 				else
 				{
-					// Use the system random number generator
+					// Uses the C function php_rand() which is the same as rand()
 					Encrypt::$_rand = MCRYPT_RAND;
 				}
 			}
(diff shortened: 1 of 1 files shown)


classes/Kohana/Encrypt.php
```diff
@@ -162,6 +162,9 @@ public function encode($data)
 			// The system random number generator must always be seeded each
 			// time it is used, or it will not produce true random results
 			mt_srand();
```
shadowhand, line 164: This should be fixed, [mcrypt_create_iv docs](http://php.net/manual/en/function.mcrypt-create-iv.php) state that using srand() is no longer required when using MCRYPT_RAND. These lines should be removed. I'm not sure where using mt_srand() came from.

sarciszewski, line 164: Well, if there's mystery code then this merits further examination.

shadowhand:
This is awesome, but it is insufficient. Since PHP 5.3, additional modes are available in Windows, which means that we can just error out when neither of the other modes are available.

shadowhand:
And finally, this pull request is targeting the wrong branch. All pull requests must be made against develop branch, not master.

sarciszewski:
My mistake.

zombor:
Please see the contribution documentation here: https://github.com/kohana/kohana/blob/3.3/master/CONTRIBUTING.md

PR #463, opened 2014-02-24
Re-fixes #3931: brace on the next line according to the coding standard

enov:
changed the word "twice" to "again" in phpdoc
also the brace of `if` should go on the next line
thanks @EvanPurkhiser


zeelot:
Thanks :)

PR #464, opened 2014-02-25
Fixes #4819 replace preg replace with /e preg replace callback

enov:
uses `create_function` to support both PHP 5.2 and PHP 5.5
`create_function` uses `eval` internally so probably it's
better to have this fix only for Kohana v.3.2

Thank you @Zeelot for the help

### DIFF
--- a/utf8/ucwords.php
+++ b/utf8/ucwords.php
@@ -7,6 +7,7 @@
  * @copyright  (c) 2007-2012 Kohana Team
  * @copyright  (c) 2005 Harry Fuecks
  * @license    http://www.gnu.org/licenses/old-licenses/lgpl-2.1.txt
+ * @uses _ucwords_preg_replace_callback callback for preg_replace_callback
  */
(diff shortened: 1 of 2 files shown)


utf8/ucwords.php
```diff
@@ -15,9 +16,19 @@ function _ucwords($str)
...
 		$str
 	);
+}
+/**
+ * helper callback function for _ucwords' preg_replace_callback
+ *
+ * @param array $matches
+ * @return string
+ */
+function _ucwords_preg_replace_callback($matches)
```
zeelot, line 31: Did you just miss this one?

enov, line 31: heh! So that's considered an API change too? Alright, will add another `create_function` in there... :soon:

enov, line 31: PR https://github.com/kohana/core/pull/465

Cheers!

enov:
Closed in favor of #465

PR #466, opened 2014-02-28
Update license URLs in PHPDoc - Ref: #4821

randomecho:
Found links pointing to old location of license instead of where they are now at http://kohanaframework.org/license

Ref: http://dev.kohanaframework.org/issues/4821


shadowhand:
Thanks!

PR #467, opened 2014-03-05
Refix #4079: Decode slashes back after rawurlencode

enov:
Apache's `AllowEncodedSlashes` directive is `Off` by default.
URLs with encoded slashes will bypass Apache's `AcceptPathInfo` setting.
Such URLs will not be redirected to index.php


zeelot:
Thanks!

PR #469, opened 2014-03-24
#4827 Fixes to speed up, cross links for userguide

WinterSilence:
pluck: `array_column()` work faster than `foreach()`.
count: if `$array` is string `instanceof` try autoload\find class for check, new version work faster, also rename second attribute as `$recursive` - it's more comprehensible.
is_assoc: I add notice, but maybe add validation to 0 for the first key? It's expensive and not too much more reliable.

### DIFF
--- a/classes/Kohana/Arr.php
+++ b/classes/Kohana/Arr.php
@@ -5,18 +5,21 @@
  * @package    Kohana
  * @category   Helpers
  * @author     Kohana Team
- * @copyright  (c) 2007-2012 Kohana Team
+ * @copyright  (c) 2007-2014 Kohana Team
  * @license    http://kohanaframework.org/license
  */
-class Kohana_Arr {
+abstract class Kohana_Arr {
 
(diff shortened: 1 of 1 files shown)


classes/Kohana/Arr.php
```diff
@@ -5,18 +5,21 @@
...
- * @copyright  (c) 2007-2012 Kohana Team
+ * @copyright  (c) 2007-2014 Kohana Team
  * @license    http://kohanaframework.org/license
  */
-class Kohana_Arr {
+abstract class Kohana_Arr {
 
 	/**
-	 * @var  string  default delimiter for path()
+	 * @var  string  Default delimiter for [Arr::path()]
```
shadowhand, line 14: isn't it `Arr::path` without `()`?

WinterSilence, line 14: @shadowhand  Both version are possible,  it's more comprehensible.

PR #475, opened 2014-04-08
Move Log_StdOut \ Log_StdErr into minion module (FR 4830)

WinterSilence:
http://dev.kohanaframework.org/issues/4830


kemo:
Why does it need to be separated? Also, in that chunk of code you default to STDERR if no stream is provided. That should probably be detected based on the message level when `Log_Cli::write()` is called, e.g. `$stream = ($message['level'] <= Log::ERROR) ? STDERR : STDOUT`

WinterSilence:
> Why does it need to be separated?

Log_StdOut \ Log_StdErr use only in CLI, so it should be moved to minion module. Other classes are arranged similarly: Config_Database, Session Database, Auth_ORM.

> Also, in that chunk of code you default to STDERR if no stream is provided.

See [Log_Writer::$_log_levels](https://github.com/kohana/core/blob/3.3/master/classes/Kohana/Log/Writer.php#L31) - 5 еrrors / 3 debug.

> That should probably be detected based on the message level when Log::write() is called, e.g. $stream = ($message['level'] <= Log::ERROR) ? STDERR : STDOUT

Cool idea! Small fix: Log::WARNING it's too error message: 

```
$stream = ( ! isset($message['level']) OR $message['level'] < Log::NOTICE) ? STDERR : STDOUT;
```

PR #476, opened 2014-04-12
Simplify Kohana::globals() test

jeffchan:
This simplifies the `test_globals_removes_user_def_globals` test
to take advantage of PHPUnit's backupGlobals option which is on
by default. There's no need to backup globals.

In addition, changed to avoid directly assigning to $GLOBALS,
since it is not allowed in HHVM. See:
https://github.com/facebook/hhvm/blob/master/hphp/doc/inconsistencies#L127


kemo:
Can you please open another one targeted at 3.3/develop?

jeffchan:
Oops! Sorry about that. See #477

PR #478, opened 2014-04-15
UTF8: fixed typos and bugs

WinterSilence:
Bugs: 
UTF8::clean():
check UTF8::$server_utf8 not specified
Error result: http://kohanaframework.org/3.3/guide-api/Kohana ErrorException [Fatal Error]: Call to undefined function mb_convert_encoding().

UTF8::$server_utf8:
Set Kohana_UTF8::$server_utf8 not true, because UTF8 use property as UTF8::$server_utf8

### DIFF
--- a/classes/Kohana/UTF8.php
+++ b/classes/Kohana/UTF8.php
@@ -1,30 +1,30 @@
 <?php defined('SYSPATH') OR die('No direct script access.');
 /**
- * A port of [phputf8](http://phputf8.sourceforge.net/) to a unified set
- * of files. Provides multi-byte aware replacement string functions.
+ * A port of [phputf8](http://phputf8.sourceforge.net) to a unified set of files.
+ * Provides multi-byte aware replacement string functions.
  *
  * For UTF-8 support to work correctly, the following requirements must be met:
  *
  * - PCRE needs to be compiled with UTF-8 support (--enable-utf8)
- * - Support for [Unicode properties](http://php.net/manual/reference.pcre.pattern.modifiers.php)
+ * - Support for [Unicode properties](http://php.net/reference.pcre.pattern.modifiers)
  *   is highly recommended (--enable-unicode-properties)
  * - The [mbstring extension](http://php.net/mbstring) is highly recommended,
  *   but must not be overloading string functions
  *
  * [!!] This file is licensed differently from the rest of Kohana. As a port of
- * [phputf8](http://phputf8.sourceforge.net/), this file is released under the LGPL.
+ * [phputf8](http://phputf8.sourceforge.net), this file is released under the LGPL.
  *
  * @package    Kohana
  * @category   Base
  * @author     Kohana Team
- * @copyright  (c) 2007-2012 Kohana Team
+ * @copyright  (c) 2007-2014 Kohana Team
  * @copyright  (c) 2005 Harry Fuecks
- * @license    http://www.gnu.org/licenses/old-licenses/lgpl-2.1.txt
+ * @license    http://gnu.org/licenses/old-licenses/lgpl-2.1.txt
  */
-class Kohana_UTF8 {
+abstract class Kohana_UTF8 {
 
 	/**
-	 * @var  boolean  Does the server support UTF-8 natively?
+	 * @var  bool  Does the server support UTF-8 natively?
 	 */
 	public static $server_utf8 = NULL;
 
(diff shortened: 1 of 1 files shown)


kemo:
Any reason why `require` calls have been replaced with `require_once`?

WinterSilence:
Why do we need to reload the file containing the function?

kemo:
We don't reload it, it's `if not loaded, load it`. Reloading it would surely cause an error about function being defined already. And `require_once` is slower than `require` or at least it used to be

WinterSilence:
Difficulty translation :( for example: 

```
    public static function substr_replace($str, $replacement, $offset, $length = NULL)
    {
        if ( ! isset(UTF8::$called[__FUNCTION__]))
        {
            // This code called only once!!!
            require Kohana::find_file('utf8', __FUNCTION__);
            // Function has been called
            UTF8::$called[__FUNCTION__] = TRUE;
        }

        return _substr_replace($str, $replacement, $offset, $length);
    }
}
```

require_once protects reload the file containing the function. require against the file causing the error. i use require_once instead require an additional foolproof. 
loss rate in this case is minimal.

kemo:
`if ( ! isset(UTF8::$called[__FUNCTION__]))` will prevent reloading the file twice, those files aren't meant to be included from anywhere else anyway

WinterSilence:
replace

fleep:
Is there a reason the UTF8 class acts as an interface to _substr_replace() and is it conceivable that we would add this for other functions defined in [/core/utf8](https://github.com/kohana/core/tree/3.3/master/utf8)?  If so, then this "load the file only once" functionality should be punched out into a protected method that can be reused in interfaces to other methods.

In that case, we should also probably catch catch cases where Kohana::find_file() returns FALSE.

classes/Kohana/UTF8.php
```diff
@@ -325,24 +328,25 @@ public static function strtolower($str)
...
-	 * @uses    UTF8::$server_utf8
 	 * @uses    Kohana::$charset
+	 * @uses    Kohana::find_file
 	 */
 	public static function strtoupper($str)
 	{
 		if (UTF8::$server_utf8)
+		{
 			return mb_strtoupper($str, Kohana::$charset);
+		}
```
kemo, line 345: Why do you add braces to a terminating condition? These are one-liners

classes/Kohana/UTF8.php
```diff
@@ -325,24 +328,25 @@ public static function strtolower($str)
...
 
 		if ( ! isset(UTF8::$called[__FUNCTION__]))
 		{
 			require Kohana::find_file('utf8', __FUNCTION__);
-
 			// Function has been called
```
kemo, line 350: Why is whitespace removed here?

classes/Kohana/UTF8.php
```diff
@@ -758,8 +760,8 @@ public static function from_unicode($arr)
 
 }
 
-if (Kohana_UTF8::$server_utf8 === NULL)
+if (UTF8::$server_utf8 === NULL)
 {
 	// Determine if this server supports UTF-8 natively
-	Kohana_UTF8::$server_utf8 = extension_loaded('mbstring');
+	UTF8::$server_utf8 = extension_loaded('mbstring');
```
kemo, line 766: This breaks the unit tests, UTF8 isn't loaded at this point.

WinterSilence, line 766: it's because file should contain only class. We can add  method for set\get server_utf8 and delete this code.

classes/Kohana/UTF8.php
WinterSilence, line 350: what's the point in it?

classes/Kohana/UTF8.php
WinterSilence, line 345: it's considered good practice

Ikke, line 345: Did you forget about the 'goto fail'  bug in OSX? And iirc, the code standard says even oneline if-statements should use  braces.

classes/Kohana/UTF8.php
Ikke, line 350: Readability, just like paragraphs. Especially before comments.

classes/Kohana/UTF8.php
```diff
@@ -351,21 +355,21 @@ public static function strtoupper($str)
...
-	 * Makes a UTF-8 string's first character uppercase. This is a UTF8-aware
-	 * version of [ucfirst](http://php.net/ucfirst).
+	 * Makes a UTF-8 string's first character uppercase.
+	 * This is a UTF8-aware version of [ucfirst](http://php.net/ucfirst).
 	 *
 	 *     $str = UTF8::ucfirst($str);
 	 *
 	 * @author  Harry Fuecks <hfuecks@gmail.com>
-	 * @param   string  $str mixed case string
+	 * @param   string  $str  Mixed case string
```
Ikke, line 364: This ads an unnecessary space.

classes/Kohana/UTF8.php
kemo, line 345: http://kohanaframework.org/3.3/guide/kohana/conventions#single-line-statements

PR #479, opened 2014-04-16
Update Core.php

dimonbat:
Hi!
is it good idea?
;)

### DIFF
--- a/classes/Kohana/Core.php
+++ b/classes/Kohana/Core.php
@@ -16,8 +16,8 @@
 class Kohana_Core {
 
 	// Release version and codename
-	const VERSION  = '3.3.1';
-	const CODENAME = 'peregrinus';
+	const VERSION  = '3.3.2';
+	const CODENAME = 'luscinia';
 
 	// Common environment type constants for consistency and convenience
 	const PRODUCTION  = 10;


WinterSilence:
3.2 & 3.3 closed

enov:
Thanks @dimonbat, but we can't directly merge to 3.3/master.

But this reminds me that we need to change those before releasing 3.3.3.
cc @acoulton

acoulton:
@enov is there a "what to do for a release" document anywhere that you know of? Maybe we should make one?

enov:
I am not aware of such document. @shadowhand definitely knows about this.

lenton:
@acoulton @enov https://github.com/kohana/core/blob/3.4/develop/guide/kohana/upgrading.md

enov:
@lenton "what are the steps to release a new Kohana version" :)

enov:
It's basically
- Change the version (maybe also the codename)
- git merge to master
- git tag the version
- prepare a zip to be downloaded from the website
- prepare the guides for official modules in case of a new version (this seems tricky)
- etc...

@shadowhand?

lenton:
Oh right, well the contributing doc mentions a little about a release branch: https://github.com/kohana/kohana/blob/3.4/develop/CONTRIBUTING.md#branch-name-meanings

enov:
Hmm.. you're right:

> release branches are for maintenance work before a release. This branch should be branched from the develop branch only. Change the version number/code name here, and apply any other maintenance items needed before actually releasing. Merges from master should only come from this branch. It should be merged to develop when it's complete as well. This branch is deleted after it's done.

shadowhand:
We basically use [git flow](http://nvie.com/posts/a-successful-git-branching-model/), slightly modified to include the various dev version branches.

acoulton:
@shadowhand I was more meaning the non-git steps - changing version number in files, etc - @enov's [list](https://github.com/kohana/core/pull/479#issuecomment-63636915)  has most of what I can think of, anything else you know about?

PR #480, opened 2014-04-18
Initialize UTF8::$server_utf8 moved into Kohana_UTF8::native_utf8()

WinterSilence:
Class file should contain only class itself. Initialize UTF8::$server_utf8 moved into Kohana_UTF8::native_utf8().


shadowhand:
This should be combined with #481. It doesn't make sense as a standalone PR.

PR #481, opened 2014-04-18
Add native_utf8, fix _load, clean, is_ascii

WinterSilence:
`UTF8::native_utf8()`: added to preserve purity of class file

`UTF8::_load()`: need use require_once because `UTF8::$called` is public property:

```
unset(UTF8::$called['strlen']);
UTF8::strlen($str);
```

`UTF8::clean()`: reordered `UTF8::native_utf8()` and `UTF8::is_ansii()` - basic condition "server_utf8 = true".

`UTF8::is_ascii()`: added support for traversable objects

### DIFF
--- a/classes/Kohana/UTF8.php
+++ b/classes/Kohana/UTF8.php
@@ -34,16 +34,34 @@ abstract class Kohana_UTF8 {
 	public static $called = array();
 
 	/**
-	 * Includes function files (once)
+	 * Check extension [mbstring](http://php.net/mbstring)
+	 * and initialize [UTF8::$server_utf8].
 	 *
-	 * @param  string $function
-	 * @return void
+	 * @return bool
+	 */
+	public static function native_utf8()
+	{
+		if ( ! is_bool(UTF8::$server_utf8))
+		{
+			// Determine if this server supports UTF-8 natively
+			UTF8::$server_utf8 = extension_loaded('mbstring');
+		}
+
+		return UTF8::$server_utf8;
+	}
+
+	/**
+	 * Include once file with "utf8" function.
+	 *
+	 * @param   string  $function  Function name
+	 * @return  void
+	 * @uses    Kohana::find_file
 	 */
 	protected static function _load($function)
 	{
 		if ( ! isset(UTF8::$called[$function]))
 		{
-			require Kohana::find_file('utf8', $function);
+			require_once Kohana::find_file('utf8', $function);
 
 			// Function has been called
 			UTF8::$called[$function] = TRUE;
@@ -58,9 +76,10 @@ abstract class Kohana_UTF8 {
 	 *     UTF8::clean($_GET); // Clean GET data
 	 *
 	 * @param   mixed   $var      Variable to clean
-	 * @param   string  $charset  Character set, defaults to Kohana::$charset
+	 * @param   string  $charset  Character set, defaults to [Kohana::$charset]
 	 * @return  mixed
 	 * @uses    Kohana::$charset
+	 * @uses    Arr::is_array
 	 */
 	public static function clean($var, $charset = NULL)
 	{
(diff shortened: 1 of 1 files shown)


classes/Kohana/UTF8.php
```diff
@@ -34,16 +34,34 @@
...
+	 * @param   string  $function  Function name
+	 * @return  void
+	 * @uses    Kohana::find_file
 	 */
 	protected static function _load($function)
 	{
 		if ( ! isset(UTF8::$called[$function]))
 		{
-			require Kohana::find_file('utf8', $function);
+			require_once Kohana::find_file('utf8', $function);
```
shadowhand: absolutely not. require_once is magnitudes slower than require: http://gazehawk.com/blog/php-require-performance/

classes/Kohana/UTF8.php
```diff
@@ -104,13 +123,16 @@ public static function clean($var, $charset = NULL)
 	 *
 	 *     $ascii = UTF8::is_ascii($str);
 	 *
-	 * @param   mixed  $str  String or array of strings to check
+	 * @param   mixed  $str  String or array (or traversable object) of strings to check
 	 * @return  bool
+	 * @uses    Arr::is_array
```
shadowhand, line 133: introducing tighter coupling into a core class? i don't like it.

classes/Kohana/UTF8.php
```diff
@@ -473,9 +490,8 @@ public static function strcspn($str, $mask, $offset = NULL, $length = NULL)
 	 * @param   string  $str               Input string
 	 * @param   int     $final_str_length  Desired string length after padding
 	 * @param   string  $pad_str           String to use as padding
-	 * @param   string  $pad_type          Padding type: STR_PAD_RIGHT, STR_PAD_LEFT, or STR_PAD_BOTH
+	 * @param   string  $pad_type          Padding type: STR_PAD_RIGHT, STR_PAD_LEFT or STR_PAD_BOTH
```
shadowhand: what the hell? why would anyone want to remove an [oxford comma](https://en.wikipedia.org/wiki/Serial_comma)?!

oxford comma

classes/Kohana/UTF8.php
```diff
@@ -34,16 +34,34 @@
...
+		{
+			// Determine if this server supports UTF-8 natively
+			UTF8::$server_utf8 = extension_loaded('mbstring');
+		}
+
+		return UTF8::$server_utf8;
+	}
+
+	/**
+	 * Include once file with "utf8" function.
```
shadowhand, line 54: sentence backwards it is.

classes/Kohana/UTF8.php
WinterSilence, line 133: We already add Arr::is_array() in UTF8::clean()

shadowhand, line 133: Then it should be removed.

WinterSilence, line 133: I suggested this elsewhere, but was not accepted.

classes/Kohana/UTF8.php
WinterSilence, line 54: Сomment your opinion.

kemo:
`UTF8::_load()` shouldn't use `require_once` but `UTF8::$called` should be `protected $_called` instead

kemo:
+1 for moving that condition into a method, I just moved your code where it 'belonged' (not really but it - at least - made unit tests pass)

classes/Kohana/UTF8.php
kemo, line 54: Maybe just `UTF8::is_native()`?

WinterSilence, line 54:
```
protected static UTF8::$_server_utf8;

public static function server_utf8() 
{
    // ...
}
```

I think I better make the property protected

PR #486, opened 2014-04-24
URL and Eclipse project fix

farazmunj:
Added some conditions in HTML class to accept external links starting with // 
also added Eclipse project files to ignore list.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 *~
 *.swp
+/.settings
+/.project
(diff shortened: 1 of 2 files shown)


randomecho:
A few things to note at least:
- Target the `develop` branch, not `master`
- Use `AND` instead of `&&`
- Do you have tests?

zombor:
This feature has already been rejected in the past, iirc. There's probably details about it in Redmine.

PR #487, opened 2014-05-07
Fixed bugs, typos

WinterSilence:
fixed bug with undefined _protocol, moved from protocol() to __construct() - it will be faster

`return UTF8::strlen($this->_body);` correctly than strlen()

`if (function_exists('http_build_cookie'))` [function_exists faster than extension_loaded](http://www.merkushin.com/function_exists-vs-extension_loaded/) sick! but true 0_o

### DIFF
--- a/classes/Kohana/Response.php
+++ b/classes/Kohana/Response.php
@@ -400,7 +400,7 @@ abstract class Kohana_Response implements HTTP_Response {
 			$options['delete'] = FALSE;
 
 			// Get the content size
-			$size = UTF8::strlen($this->_body);
+			$size = $this->content_length();
 
 			// Create a temporary file to hold the current response
 			$file = tmpfile();
--- a/classes/Kohana/Response.php
+++ b/classes/Kohana/Response.php
@@ -1,15 +1,14 @@
 <?php defined('SYSPATH') OR die('No direct script access.');
 /**
  * Response wrapper. Created as the result of any [Request] execution
- * or utility method (i.e. Redirect). Implements standard HTTP
+ * or utility method (i.e. redirect). Implements standard HTTP
  * response format.
  *
  * @package    Kohana
  * @category   Base
  * @author     Kohana Team
- * @copyright  (c) 2008-2012 Kohana Team
- * @license    http://kohanaphp.com/license
- * @since      3.1.0
+ * @copyright  (c) 2008-2014 Kohana Team
+ * @license    http://kohanaframework.org/license
  */
 class Kohana_Response implements HTTP_Response {
 
@@ -23,7 +22,7 @@ class Kohana_Response implements HTTP_Response {
 	 *      // Create a new response with headers
 	 *      $response = Response::factory(array('status' => 200));
 	 *
-	 * @param   array    $config Setup the response object
+	 * @param   array  $config  Setup the response object
 	 * @return  Response
 	 */
 	public static function factory(array $config = array())
@@ -31,7 +30,9 @@ class Kohana_Response implements HTTP_Response {
 		return new Response($config);
 	}
 
-	// HTTP status codes and messages
+	/**
+	 * @var  array  HTTP status codes and messages
+	 */ 
 	public static $messages = array(
 		// Informational 1xx
 		100 => 'Continue',
@@ -87,7 +88,7 @@ class Kohana_Response implements HTTP_Response {
 	);
 
 	/**
-	 * @var  integer     The response http status
+	 * @var  integer  The response HTTP status
 	 */
 	protected $_status = 200;
 
@@ -97,43 +98,47 @@ class Kohana_Response implements HTTP_Response {
 	protected $_header;
 
 	/**
-	 * @var  string      The response body
+	 * @var  string  The response body
 	 */
 	protected $_body = '';
 
 	/**
-	 * @var  array       Cookies to be returned in the response
+	 * @var  array  Cookies to be returned in the response
 	 */
 	protected $_cookies = array();
 
 	/**
-	 * @var  string      The response protocol
+	 * @var  string  The response protocol
 	 */
 	protected $_protocol;
 
 	/**
 	 * Sets up the response object
 	 *
-	 * @param   array $config Setup the response object
+	 * @param   array  $config  Setup the response object
 	 * @return  void
 	 */
 	public function __construct(array $config = array())
 	{
 		$this->_header = new HTTP_Header;
 
+		if (isset($config['_header']))
+		{
+			$this->headers($config['_header']);
+			unset($config['_header']);
+		}
+
 		foreach ($config as $key => $value)
 		{
 			if (property_exists($this, $key))
 			{
-				if ($key == '_header')
-				{
-					$this->headers($value);
-				}
-				else
-				{
-					$this->$key = $value;
-				}
+				$this->$key = $value;
 			}
+		}
+
+		if (!$this->_protocol)
+		{
+			$this->_protocol = HTTP::$protocol;
 		}
 	}
 
@@ -162,10 +167,9 @@ class Kohana_Response implements HTTP_Response {
 	}
 
 	/**
-	 * Gets or sets the HTTP protocol. The standard protocol to use
-	 * is `HTTP/1.1`.
+	 * Gets or sets the HTTP protocol.
 	 *
-	 * @param   string   $protocol Protocol to set to the request/response
+	 * @param   string  $protocol  Protocol to set to the request/response
 	 * @return  mixed
 	 */
 	public function protocol($protocol = NULL)
@@ -174,11 +178,6 @@ class Kohana_Response implements HTTP_Response {
 		{
 			$this->_protocol = strtoupper($protocol);
 			return $this;
-		}
-
-		if ($this->_protocol === NULL)
-		{
-			$this->_protocol = HTTP::$protocol;
 		}
 
 		return $this->_protocol;
@@ -194,12 +193,13 @@ class Kohana_Response implements HTTP_Response {
 	 *      // Get the current status
 	 *      $status = $response->status();
 	 *
-	 * @param   integer  $status Status to set to this response
+	 * @param   integer  $status  Status to set to this response
 	 * @return  mixed
+	 * @throws  Kohana_Exception
 	 */
-	public function status($status = NULL)
+	public function status($status = 0)
 	{
-		if ($status === NULL)
+		if ($status === 0)
 		{
 			return $this->_status;
 		}
@@ -210,7 +210,10 @@ class Kohana_Response implements HTTP_Response {
 		}
 		else
 		{
-			throw new Kohana_Exception(__METHOD__.' unknown status value : :value', array(':value' => $status));
+			throw new Kohana_Exception
+				':method: unknown status value: :value',
+				array(':method' => __METHOD__, ':value' => $status)
+			);
 		}
 	}
 
@@ -231,9 +234,9 @@ class Kohana_Response implements HTTP_Response {
 	 *       // Set multiple headers
 	 *       $response->headers(array('Content-Type' => 'text/html', 'Cache-Control' => 'no-cache'));
 	 *
-	 * @param mixed $key
-	 * @param string $value
-	 * @return mixed
+	 * @param   mixed   $key
+	 * @param   string  $value
+	 * @return  mixed
 	 */
 	public function headers($key = NULL, $value = NULL)
 	{
@@ -241,7 +244,7 @@ class Kohana_Response implements HTTP_Response {
 		{
 			return $this->_header;
 		}
-		elseif (is_array($key))
+		elseif (Arr::is_array($key))
 		{
 			$this->_header->exchangeArray($key);
 			return $this;
@@ -258,14 +261,13 @@ class Kohana_Response implements HTTP_Response {
 	}
 
 	/**
-	 * Returns the length of the body for use with
-	 * content header
+	 * Returns the length of the body for use with content header
 	 *
 	 * @return  integer
 	 */
 	public function content_length()
 	{
-		return strlen($this->body());
+		return UTF8::strlen($this->_body);
 	}
 
 	/**
@@ -280,11 +282,9 @@ class Kohana_Response implements HTTP_Response {
 	 *          'expiration' => 12352234
 	 *     ));
 	 *
-	 * @param   mixed   $key    cookie name, or array of cookie values
-	 * @param   string  $value  value to set to cookie
-	 * @return  string
-	 * @return  void
-	 * @return  [Response]
+	 * @param   mixed   $key    Cookie name, or array of cookie values
+	 * @param   string  $value  Value to set to cookie
+	 * @return  mixed
 	 */
 	public function cookie($key = NULL, $value = NULL)
 	{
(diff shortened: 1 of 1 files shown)


classes/Kohana/Response.php
```diff
@@ -1,29 +1,27 @@
...
  * @category   Base
  * @author     Kohana Team
- * @copyright  (c) 2008-2012 Kohana Team
- * @license    http://kohanaphp.com/license
- * @since      3.1.0
+ * @copyright  (c) 2008-2014 Kohana Team
+ * @license    http://kohanaframework.org/license
  */
-class Kohana_Response implements HTTP_Response {
+abstract class Kohana_Response implements HTTP_Response {
```
shadowhand, line 12: what... this makes no sense.

classes/Kohana/Response.php
```diff
@@ -1,29 +1,27 @@
...
 
 	/**
 	 * Factory method to create a new [Response]. Pass properties
 	 * in using an associative array.
 	 *
-	 *      // Create a new response
-	 *      $response = Response::factory();
+	 *     // Create a new response
+	 *     $response = Response::factory();
```
shadowhand, line 19: why are you reformatting whitespace?

classes/Kohana/Response.php
```diff
@@ -97,44 +97,48 @@ public static function factory(array $config = array())
...
-				{
-					$this->$key = $value;
-				}
+				$this->$key = $value;
 			}
 		}
+
+		if (!$this->_protocol)
+		{
+			$this->_protocol = HTTP::$protocol;
```
shadowhand, line 140: and where does `$protocol` come from? it's not defined...

classes/Kohana/Response.php
```diff
@@ -176,30 +179,25 @@ public function protocol($protocol = NULL)
 			return $this;
 		}
 
-		if ($this->_protocol === NULL)
```
shadowhand, line 179: looks like that was a mistake...

classes/Kohana/Response.php
```diff
@@ -176,30 +179,25 @@ public function protocol($protocol = NULL)
...
+	 *     // Get the current status
+	 *     $status = $response->status();
 	 *
-	 * @param   integer  $status Status to set to this response
+	 * @param   integer  $status  Status to set to this response
 	 * @return  mixed
+	 * @throws  Kohana_Exception
 	 */
-	public function status($status = NULL)
+	public function status($status = 0)
```
shadowhand, line 198: why?

classes/Kohana/Response.php
```diff
@@ -210,7 +208,10 @@ public function status($status = NULL)
 		}
 		else
 		{
-			throw new Kohana_Exception(__METHOD__.' unknown status value : :value', array(':value' => $status));
+			throw new Kohana_Exception(
```
shadowhand, line 211: why?

classes/Kohana/Response.php
```diff
@@ -248,24 +249,21 @@ public function headers($key = NULL, $value = NULL)
 		}
 		elseif ($value === NULL)
 		{
-			return Arr::get($this->_header, $key);
-		}
-		else
-		{
-			$this->_header[$key] = $value;
-			return $this;
+			return $this->_header->offsetGet($value);
```
shadowhand, line 252: why

classes/Kohana/Response.php
```diff
@@ -248,24 +249,21 @@ public function headers($key = NULL, $value = NULL)
...
-	 * Returns the length of the body for use with
-	 * content header
+	 * Returns the length of the body for use with content header
 	 *
 	 * @return  integer
 	 */
 	public function content_length()
 	{
-		return strlen($this->body());
+		return UTF8::strlen($this->_body);
```
shadowhand, line 266: ~~finally! a good change.~~

edit: nope, see next comment.

classes/Kohana/Response.php
```diff
@@ -379,55 +376,37 @@ public function send_headers($replace = FALSE, $callback = NULL)
...
+			$size = $this->content_length();
 
 			// Create a temporary file to hold the current response
 			$file = tmpfile();
 
 			// Write the current response into the file
-			fwrite($file, $file_data);
-
-			// File data is no longer needed
-			unset($file_data);
```
shadowhand, line 430: what... why?

classes/Kohana/Response.php
```diff
@@ -443,34 +422,29 @@ public function send_file($filename, $download = NULL, array $options = NULL)
 			// Get the file size
 			$size = filesize($filename);
 
-			if ( ! isset($mime))
```
shadowhand, line 446: wtf...

classes/Kohana/Response.php
```diff
@@ -443,34 +422,29 @@ public function send_file($filename, $download = NULL, array $options = NULL)
...
-			));
+			throw new Kohana_Exception(
+				':method: Could not read file to send: :file', 
+				array(':method' => __METHOD__, ':file' => $download)
+			);
 		}
 
 		// Inline or download?
-		$disposition = empty($options['inline']) ? 'attachment' : 'inline';
+		$disposition = isset($options['inline']) ? 'inline' : 'attachment';
```
shadowhand, line 438: what if i pass `inline = FALSE` ?

classes/Kohana/Response.php
```diff
@@ -443,34 +422,29 @@ public function send_file($filename, $download = NULL, array $options = NULL)
...
 
 		// Calculate byte range to download.
 		list($start, $end) = $this->_calculate_byte_range($size);
 
-		if ( ! empty($options['resumable']))
+		if (isset($options['resumable']))
```
shadowhand, line 443: same thing, what about `resumable = false` ?

classes/Kohana/Response.php
```diff
@@ -479,23 +453,34 @@ public function send_file($filename, $download = NULL, array $options = NULL)
...
 			// Naturally, IE does not act like a real browser...
-			if (Request::$initial->secure())
+			if (Request::initial()->secure())
 			{
-				// http://support.microsoft.com/kb/316431
+				// @link http://support.microsoft.com/kb/316431
 				$this->_header['pragma'] = $this->_header['cache-control'] = 'public';
 			}
 
-			if (version_compare(Request::user_agent('version'), '8.0', '>='))
```
shadowhand, line 496: wtf..

classes/Kohana/Response.php
```diff
@@ -503,10 +488,12 @@ public function send_file($filename, $download = NULL, array $options = NULL)
 		// Send all headers now
 		$this->send_headers();
 
-		while (ob_get_level())
+		// Flush all output buffers
+		$levels = ob_get_level();
```
shadowhand, line 492: uh, no...

classes/Kohana/Response.php
```diff
@@ -518,16 +505,19 @@ public function send_file($filename, $download = NULL, array $options = NULL)
...
 
-		// Send data in 16kb blocks
-		$block = 1024 * 16;
+		// Send data in 16kb (1024*16) blocks
+		$block = 16384;
 
 		fseek($file, $start);
 
-		while ( ! feof($file) AND ($pos = ftell($file)) <= $end)
+		while ( ! connection_aborted() AND ! feof($file))
```
shadowhand, line 513: again, no.

classes/Kohana/Response.php
```diff
@@ -544,7 +534,7 @@ public function send_file($filename, $download = NULL, array $options = NULL)
 		// Close the file
 		fclose($file);
 
-		if ( ! empty($options['delete']))
+		if (isset($options['delete']) AND $options['delete'])
```
shadowhand, line 537: wow, pointless changes abound.

classes/Kohana/Response.php
```diff
@@ -556,13 +546,10 @@ public function send_file($filename, $download = NULL, array $options = NULL)
 				// Create a text version of the exception
 				$error = Kohana_Exception::text($e);
 
-				if (is_object(Kohana::$log))
+				if (Kohana::$log instanceof Log)
```
shadowhand, line 549: no.

shadowhand:
i stopped reviewing half way through. if you can't separate your opinions from your bug fixes, please stop submitting pull requests. it is a huge waste of everyone's time when you do this.

WinterSilence:
@shadowhand Nobody forces to spend your time, you can ignore my requests. Some of the issues you mentioned, I described in the comments to request.
I do not understand why asking questions and then close the query, you probably are not interested in the answers.

shadowhand:
@WinterSilence i should ignore pull requests? okay... i'll quote you on that next time someone complains that no one pays attention to them.

classes/Kohana/Response.php
WinterSilence, line 12: Then half of the available classes.
Eg: `abstract class Kohana_Session {`
https://github.com/kohana/core/blob/3.4/develop/classes/Kohana/Session.php#L11

classes/Kohana/Response.php
WinterSilence, line 19: remove extra spaces, should be 4 spaces for code + 1 border

shadowhand, line 19: so what? it has **nothing** to do with the pull request goals of "fixed bugs"

classes/Kohana/Response.php
WinterSilence, line 140: At current version default protocol is set only when calling `protocol()`, this wrong

classes/Kohana/Response.php
WinterSilence, line 19: I'm trying to tidy up the code and comments to him. Request title: "Fixed bugs, typos"

classes/Kohana/Response.php
WinterSilence, line 179: See latest version https://github.com/WinterSilence/core/blob/9e6d8368d4cfde907706ccf191ce7f45a01fb99b/classes/Kohana/Response.php#L174

classes/Kohana/Response.php
shadowhand, line 19: each bug fix should be a separate pull request, as a **clear**, **single** unit of change. you can put all of your opinion changes (like all the bogus whitespace changes) into a separate pull request.

classes/Kohana/Response.php
WinterSilence, line 198: `@param   integer  $status` NULL not integer value

shadowhand, line 198: yes, but `function (array $foo = NULL)` is totally acceptable. NULL literally means "not defined" and is an acceptable value for any **type** of optional parameter.

this fixes nothing and adds absolutely no value.

classes/Kohana/Response.php
WinterSilence, line 211: Cut long string, corrected passing arguments

classes/Kohana/Response.php
WinterSilence, line 252: Why native method better? are you serious?

shadowhand, line 252: no, you modified the logic, for no apparent reason... why did you modify the if/else statements? this is not fixing anything, it is an opinion change.

classes/Kohana/Response.php
WinterSilence, line 430: Extra variables are not needed

classes/Kohana/Response.php
WinterSilence, line 446: Merge duplicate code and posted below.

classes/Kohana/Response.php
WinterSilence, line 438: what if you pass nothing? - notice. I remove an additional check because the value is not used, its only flag

classes/Kohana/Response.php
WinterSilence, line 443: By analogy to `$options['inline']`

classes/Kohana/Response.php
WinterSilence, line 496: I already wrote that this version is faster.

classes/Kohana/Response.php
WinterSilence, line 492: > If you enable zlib.output_compression then level count will be increased by 1 and then this code will just freeze your script.
> http://ru2.php.net/manual/ru/function.ob-end-flush.php#103387

classes/Kohana/Response.php
WinterSilence, line 513: Explain your answer

classes/Kohana/Response.php
WinterSilence, line 549: Explain your answer

classes/Kohana/Response.php
Ikke, line 492: That's why you should create individual commits that mention these kinds of
details, not one mass change commit that is difficult to accept while there
are still issues and conveys very little information about why certain
changes are being made, especially when looking back at the commits.

Ikke, line 492: My bad, I should have watched better, there are separate commits.

classes/Kohana/Response.php
WinterSilence, line 198: For "strong" typing is not the correct value. Just look listing functions and methods of php

classes/Kohana/Response.php
shadowhand, line 438: by "pass nothing" i assume you mean `$options` being null, or `$options['inline']` not being defined. but in reality, [empty](http://php.net/empty) works like `isset` and doesn't produce notices or warnings for undefined array offsets, as per the manual:

>  That means empty() is essentially the concise equivalent to !isset($var) || $var == false.

classes/Kohana/Response.php
shadowhand, line 496: that's not a bug fix or a type, that's a performance change. hence, should be a separate pull request.

classes/Kohana/Response.php
shadowhand, line 492: No, his pull request message mentions none of this. You're absolutely right that each of these should be a separate pull request!

classes/Kohana/Response.php
WinterSilence, line 438: This is only true in this case as after checking the option not used. Сейчас добавлю для этих опций дополнительную проверку

shadowhand, line 438: You're wrong:

``` php
<?php
error_reporting(E_ALL);
// this produces no notice, warning, or error
var_dump(empty($foo["bar"]));
```

WinterSilence:
I am engaged my fork, if I find any errors or optimize code, then I send it here too. I don't have time for a more detailed request, if something looks interesting, you can add its own separate request.

shadowhand:
> I don't have time for a more detailed request, if something looks interesting, you can add its own separate request.

You are acting as if your time is more valuable and important than anyone else's. It would have taken you matter of minutes to separate these commits into distinct pull requests, but you instead you argued with my review.

I appreciate that you are trying to help, but if you would just focus your effort on using the correct process, rather than arguing, we would be able to use your contributions.

WinterSilence:
judging by the fact that a lot of criticism, and no help, then is not true. if no one wants to develop a project, makes no sense to send new requests, rather spend this time on my fork.

shadowhand:
> judging by the fact that a lot of criticism, and no help

I made it very clear what the issues with this pull request were. No one will collaborate with you on a project if you don't follow the process or make any attempt to address concerns.

classes/Kohana/Response.php
cs278, line 266: Except it's wrong, the content length header should express the [number of bytes not the number of characters](http://mark.koli.ch/remember-kids-an-http-content-length-is-the-number-of-bytes-not-the-number-of-characters).

shadowhand, line 266: good catch chris!

WinterSilence, line 266: @cs278 :+1:

PR #489, opened 2014-05-08
Fixed bug in Text::user_agent - wrong version

WinterSilence:
Alt. version:

```
if (preg_match('#'.preg_quote($search).'[^0-9.]*+([0-9.a-z]*)#i', $agent, $matches))
{
    // Set the version number
    // Detect version for IE >= 8 (Trident/4.0 - IE/8.0, Trident/5.0 - IE/8.0, etc)
    $info['version'] = $search != 'Trident' ? $matches[1] : 4.0 + $matches[1];
}
```

### DIFF
--- a/classes/Kohana/Text.php
+++ b/classes/Kohana/Text.php
@@ -649,7 +649,10 @@ class Kohana_Text {
 					// Set the browser name
 					$info['browser'] = $name;
 
-					if (preg_match('#'.preg_quote($search).'[^0-9.]*+([0-9.][0-9.a-z]*)#i', Request::$user_agent, $matches))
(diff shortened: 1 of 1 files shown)


mintbridge:
I'm not sure that this is the best way to do it, what happens if they change it again? another exception case?
If it is the only way then it needs some tests and i'd add some brackets around the ternary for clarity

WinterSilence:
> what happens if they change it again? 

Уou have friends in the leadership of microsoft? I don't have. So we will update the code  sometimes..
"Best way" is http://browscap.org/ + get_browser(), but weighs at least 4 megabytes ((
I dont care how you do it will add with or without braces. Was important to support new browsers.

PR #491, opened 2014-05-13
Add resistance to remote timing attacks

sarciszewski:
When cryptographic hashes are compared with the `===` operator, they are susceptible to remote timing attacks. This patch affects the cookie-based session driver as well as `Security::check`; you may wish to also patch other hash comparison functions.


sarciszewski:
Whoops. Wrong branch. D'oh!

PR #492, opened 2014-05-13
3.3/develop

sarciszewski:
Now on the correct branch!

When cryptographic hashes are compared with the `===` operator, they are susceptible to remote timing attacks. This patch affects the cookie-based session driver as well as `Security::check($token)`; you may wish to also patch other hash comparison functions.

`slow_equals()` taken from https://defuse.ca/php-pbkdf2.htm


shadowhand:
Thanks again!

PR #498, opened 2014-05-21
Add IE 11 and Win 8.1 user agent detection

LexLythius:
Similar to other solutions, but without altering real code, only config.


shadowhand:
All pull requests need to target a `develop` branch.

shadowhand:
But otherwise this request looks perfect.

Issue #503, opened 2014-06-04
HTML::anchor


helpse:
Hello, there is an issue with HTML::anchor

It doesnt allow you to use "?key=value" for href. The function always prepends a leading slash. It's because it uses internally URL::site. May be its there that it shold look for a leading '?' and do not prepend the slash.

helpse:
This would be a better solution: (HTML::anchor, line 129)

```
elseif ($uri[0] !== '#' AND $uri[0] !== '?')
{
    // Make the URI absolute for non-id anchors
    $uri = URL::site($uri, $protocol, $index);
}
```

enov:
On which version you are?

helpse:
3.2, 3.3. Doesnt matter, its the same even in development branches.

enov:
I guess you're making a point here. I'll mark this as a bug.

Please make a PR towards the development branches of 3.3 and 3.4. Also, add your tests, for both cases: the query part (?) and the fragment part (#):

https://github.com/kohana/core/blob/3.3/develop/tests/kohana/HTMLTest.php#L223

Thanks

enov:
@helpse this is fixed in 3.3/develop.

Thanks for reporting!

PR #506, opened 2014-06-09
bug #503: allow query-only links in HTML::anchor()

enov:
Added tests for query-only and fragment-only links

Ref #503


enov:
Waiting for peer review before merge.

Issue #508, opened 2014-06-24
Remove Direct Access Checks


lenton:
I propose the removal of all direct access checks placed at the start of all classes:

`<?php defined('SYSPATH') OR die('No direct script access.');`

This check makes classes unusable in other non-kohana projects without first having to define 'SYSPATH'. This line of code doesn't conform to the DRY principle since it's repeated a no end of times throughout the kohana core. This removal will make classes that little more cleaner and be another step towards removing all global state.

To still prevent people from accessing class files directly; index.php (or a symbolic link) could be placed in a public folder, or a htaccess file could be used.

shadowhand:
+1

neo22s:
+1

kemo:
:+1:

mintbridge:
:thumbsup:

samwilson:
Good idea.

samwilson:
Will `Kohana::FILE_SECURITY` remain, and still be added to log files?

kemo:
I'd drop it as well, log files are .php just like any other files

sergeyklay:
Most coders this causes at least a smile.
I believe that almost any server can be configured to use a single entry point index.php.

We will not be difficult to little change a sample configuration for Apache and Nginx to prove it.
Users of other servers will be able to customize server is similar

Best regards,

acoulton:
@lenton I think we can close this now as the PR's are merged? Reopen if not.

pdt256:
Thanks for removing these access checks. They were unnecessary in my setup. I place my index.php inside a www directory at the same level as application, modules, etc.

Below is a sample of my Nginx config to only allow index.php to be executed. I also have a rewrite rule for module asset directories. i.e. exposing: /var/www/www.example.com/modules/mymodule/assets/img/logo.png -> http://www.example.com/assets/mymodule/img/logo.png

<pre>
    root /var/www/www.example.com/www;
    index index.php index.html;
    location ^~ /assets/ {
        root /var/www/www.example.com/modules;
        rewrite "^/assets/([a-zA-Z0-9_-]*)/(.*)$" /$1/assets/$2 break;
    }
    location / {
        try_files $uri /index.php?$query_string;
    }
        
    location = /index.php {
        include fastcgi.conf;
        fastcgi_param KOHANA_ENV production;
        fastcgi_pass 127.0.0.1:9000;
        fastcgi_index index.php;
    }
</pre>

PR #509, opened 2014-06-26
3.4/develop

lenton:
This pull request is in relation to my other pull request: https://github.com/kohana/kohana/pull/27
This pull request passes all unit tests.

---

APPPATH and SYSPATH has been completely removed from the core, it's now not dependant on the two globals.

Core and application modules are now explicitly enabled using Kohana::modules() instead of magically being added in the on initiation. This gives the user greater control, and gives them a better understanding on what's really going on instead of lying to them. One benefit is you can now run Kohana without the application module if you wanted.

Kohana::$_paths was removed and replaced with Kohana::$_modules. The only difference between the two was that $_paths had numeric keys instead of the module names therefore it didn't need to exist.

Magic quotes, safe mode and register globals support has been removed since they are removed in PHP 5.4.

Debug::path() now also uses the paths of enabled modules when shortening a path. For example: "/var/www/kohana/vendor/kohana/database/classes/Database.php" becomes "{module:database}/classes/Database.php".

All direct access checks "<?php defined('SYSPATH') OR die('No direct script access.');" found at the start of classes and generated logs have been removed.

### DIFF
--- a/classes/Arr.php
+++ b/classes/Arr.php
@@ -1,3 +1,3 @@
-<?php defined('SYSPATH') OR die('No direct script access.');
+<?php
 
 class Arr extends Kohana_Arr {}
--- a/classes/Config.php
+++ b/classes/Config.php
@@ -1,3 +1,3 @@
-<?php defined('SYSPATH') OR die('No direct script access.');
+<?php
 
 class Config extends Kohana_Config {}
--- a/classes/Config/File.php
+++ b/classes/Config/File.php
@@ -1,3 +1,3 @@
-<?php defined('SYSPATH') OR die('No direct script access.');
+<?php
 
 class Config_File extends Kohana_Config_File {}
--- a/classes/Config/Group.php
+++ b/classes/Config/Group.php
@@ -1,3 +1,3 @@
-<?php defined('SYSPATH') OR die('No direct script access.');
+<?php
 
 class Config_Group extends Kohana_Config_Group {}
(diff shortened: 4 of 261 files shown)


serieznyi:
Why you remove `<?php defined('SYSPATH') OR die('No direct script access.');` from files?

lenton:
@serieznyi https://github.com/kohana/core/issues/508

shadowhand:
This is doing way too much in a single PR. Each PR should consist of a single coherent change in functionality, ideally with supporting unit tests.

And the CFS changes are a definite "no", for the same reasons as #510.

lenton:
Sorry, next time I'll split up my changes into different branches (I assume that's how you can achieve multiple pull requests to the same repository?) If I had to sum up this PR it'd be removing APPPATH and SYSPATH dependency from the core, to do that I had to make a lot of changes in a lot of places.

I can understand if when you saw this big wall of commits (and since this is my first real PR) you assumed I went on a massive changing spree and broke Kohana. I definitely didn't do that. I took a lot of time making sure that every commit was well coded (not just hacked into existing code), that everything else was accounted for, nothing broke in response to the changes, and I made the most minimal changes to the API as possible.

I've already made the appropriate changes to the unit tests, there wasn't a need to create any new ones. All of the unit tests pass and I've even tested the welcome controller on a webserver.

I've reverted Kohana::modules() back to its original order. Is there anything else you specifically disagree with about this PR? I would appreciate it if you took a really detailed look into my changes, I'm happy to justify any of my changes to you in more detail.

PR #510, opened 2014-07-01
feature: change kohana_find priority

serieznyi:
Kohana::$_paths not correct ordered by priority and Kohana::find_file returns wrong files

Problem:

Step 1) Init modules

```
Kohana::modules(array(
...
'database' => MODPATH.'database', // PRIOR 0
'module_core' => MODPATH.'module_core', PRIOR 1
'module_app' => MODPATH.'module_app', // PRIOR 2, module_app dependense from module_core
...
));
```

P.S. module_core != SYSPATH, module_app != APPPATH

Structure of files:

```
module_core/
-----views
----------template.php
module_app/
-----views
----------template.php // extend view file of module_core/views/template.php
```

Step 2: Get view

```
Kohana::find_file('views', 'template');
```

Result:

```
module_core/views/template.php
```

But we need module_app/views/template.php

Why we get this result?

Step 3: Algoritm find view files

```
echo Kohana::$_paths;

// Result
```

APPPATH,
module_core,
module_app
SYSPATH

```

Step 4: Show order for view location (order than now using in core (K3.3.2)).
```

APPPATH, // 1 
...,
database, // 2
module_core, // 3
module_app // 4
..., 
SYSPATH // 5

```

Step 5: The order what we need for get correct view (current priority)
```

APPPATH, // 1
...,
database, // 4
module_core, // 3
module_app // 2
..., 
SYSPATH // 5

```
Am I wrong?


```

### DIFF
--- a/classes/Kohana/Core.php
+++ b/classes/Kohana/Core.php
@@ -679,6 +679,12 @@ class Kohana_Core {
 			$benchmark = Profiler::start('Kohana', __FUNCTION__);
 		}
 
+		$paths = Kohana::$_paths;
+	        $paths = array_slice(Kohana::$_paths, 1, -1);
+	        $paths = array_reverse($paths);
+	        array_unshift($paths, APPPATH);
+	        $paths[] = SYSPATH;
+
 		if ($array OR $dir === 'config' OR $dir === 'i18n' OR $dir === 'messages')
 		{
 			// Include paths must be searched in reverse
@@ -701,7 +707,7 @@ class Kohana_Core {
 			// The file has not been found yet
 			$found = FALSE;
 
-			foreach (Kohana::$_paths as $dir)
+			foreach ($paths as $dir)
 			{
 				if (is_file($dir.$path))
 				{


serieznyi:
Ops ... I wrong in commit description :) 
Correct description is "change Kohana::find_file() find file priority"
P.S. And sorry for my English ))

shadowhand:
Module paths are checked in the order they are defined. That is how it always has been.

serieznyi:
``` php
$modules = array(
'module_core' => MODPATH . 'module_core',
'module_app' => MODPATH . 'module_app'
);
Kohana::modules($modules);
```

Tell me, in this situation files from module_app should extends files from module_core?

shadowhand:
That's no problem. Files are not loaded until requested. Classes can depend on lower modules, just as application files can depend on system files.

serieznyi:
But then I call Kohana::find_file('view', 'template'), I get module_core/tempalte.php. But I should get module_app/tempalte.php.
And I think this is wrong.
In Kohana::modules() you create that Kohana:$_path list:

```
APPATH
module_core,
module_app,
SYSPATH
```

I think you should crete that Kohana:$_path list:

```
APPATH
module_app,
module_core,
SYSPATH
```

shadowhand:
The order of the list is determined by the user, it follows the same order as the `Kohana::modules()` call. What's the problem here?

serieznyi:
In my case not problem.I always may create my own kohana/core fork and use him.
But I think Kohana::find_file work not correct.
About "he order of the list is determined by the user"
I can call (use my own reverse):

```
$modules = array(
'module_app' => MODPATH . 'module_app',
'module_core' => MODPATH . 'module_core',
);
Kohana::modules($modules);
```

But in this case init.php files be included in wrong order.

shadowhand:
Ah, `init.php` is the problem! That makes more sense, but init files are not meant to talk to other modules.

serieznyi:
In principle it is possible to abandon from some logic in init. But all the same I think this is mistake ))
Ok, I gone to use my own core fork. Bye

samwilson:
> but init files are not meant to talk to other modules.

The userguide doesn't mention anything about that. Surely there's nothing wrong with specifying that a certain module must be loaded after some other module?

I've always liked the fact that one can specify the order in which modules are loaded.

lenton:
@shadowhand He's talking about the modules being loaded in reverse order when they should logically be loaded from top to bottom. I actually have already made this change in my pull request.

shadowhand:
Modules are not loaded in reverse order, nor have they ever been. Modules are added and searched in the same order, the way the user defines them. His issues is that he wants "app" to be searched before "core", but at the same time "app" depends on "core", so "core" must be listed first.

Kohana has always worked this way, and I think it is silly to change the behavior now.

lenton:
@shadowhand Modules do get loaded in reverse order.

```
Kohana::modules([
    'foo' => 'path',
    'bar' => 'path',
    'baz' => 'path',
]);
```

This is the order that the modules get loaded in:

```
baz
bar
foo
```

You expect that the latter modules would overwrite the former ones, not the other way around.

zombor:
No. Modules always apply top to bottom. Things at the top of the list "over-rule" things below them. To do this, the top needs to be loaded last.

This is all expected behavior.

lenton:
@zombor It's much more logical to enable the modules in the order which they are going to be loaded in, it's intuitive to a programmer that modules loaded later will override previous modules.

This is coming from someone who started Kohana and got really confused when my module wasn't overwriting a file because it was placed after, and I bet there's many more people who scratched their heads.

zombor:
The whole point of the CFS is that things on the top override the things on the bottom.

lenton:
@zombor Anyone who's experienced procedural programming doesn't think that way, the CFS should be thought about as modules loaded later take precedence.

zombor:
Kohana has almost 10 years of CFS precedence here. That's how the CFS works. Top down. It's diagramed very clearly in the documentation. It's not changing.

lenton:
@zombor "because it's always been like that" isn't a valid reason, and the diagram wouldn't need to be changed. Every programmer would assume that Kohana will load modules in the exact order you specify them, not secretly use array_reverse() in Core.php. It's not programmer friendly.

zombor:
I gave you a valid reason: The CFS works top to bottom. The things at the top over-ride the things at the bottom. This is how it was designed, it's not a bug. "because it's always been like that" is also a valid reason, because that's how it was designed when it was first developed.

lenton:
@zombor The CFS would still work "top to bottom", the only difference being would be an API change where Kohana::modules() loads modules in order. Saying that something can't be changed because that's how it was first developed is silly and would make progression impossible.

zombor:
I'm saying it shouldn't be changed because thats the way it was designed. That's how we wanted it to work. Top to bottom. Things on top over-riding the things on the bottom. This is literally the first time anyone has ever said it didn't work as expected. Ever.

cbandy:
Maybe "override" isn't clear... The modules listed first have the highest precedence (i.e. priority); modules listed last have the lowest.

Later modules do not _replace_ earlier ones.

serieznyi:
Why "the modules listed first have the highest precedence (i.e. priority)" then in Kohana::$_path first element of list is SYSPATH?

In standard core the modules listed first have the lower precedence (i.e. priority) because paths structure is:

```
SYSPATH
MODULES
APPPATH
```

This is my vision of solutions
https://github.com/serieznyi/kohana-core/commit/43dd3a3a622dbfe852a8c32528799295ed366477

lenton:
@serieznyi In my pull request https://github.com/kohana/core/pull/509 I've done that and more. I've made it so you explicitly enable the core and application modules in Kohana::modules() too. Can someone please take a look at it, it's been a week now with no comments.

serieznyi:
@lenton, yours pull request add many changes and have many controversial moments. My pull request regarding only Kohana::find_file functionality

PR #512, opened 2014-07-04
don't empty the error message with special encode

Yahasana:
refer from the manual: [htmlspecialchars](http://cn2.php.net/htmlspecialchars#refsect1-function.htmlspecialchars-returnvalues)

> If the input string contains an invalid code unit sequence within the given encoding an empty string will be returned


enov:
Excellent. Thanks!

PR #514, opened 2014-07-12
Fixed method type from xdebug_get_function_stack()

paprykarz:
Without this patch I see `Kohana_Corestaticerror_handler(arguments)` instead of `Kohana_Core::error_handler(arguments)`.


enov:
Ooops... I fixed this in #555 not knowing that there was already a PR for this. Thanks anyways @paprykarz! And sorry we could not come back to you any earlier...

PR #515, opened 2014-07-21
Require PHP version 5.4.0 or greater

lenton:


shadowhand:
I like this one. :) But it's missing the install.php changes.

lenton:
This is kohana/core, already created a PR to require PHP 5.4 in kohana/kohana.

shadowhand:
Oh right. :)

PR #516, opened 2014-07-21
Remove magic quotes, safe mode and register globals support

lenton:


shadowhand:
Makes sense, despite the lack of an Oxford comma. ;)

PR #517, opened 2014-07-22
Bugfix (3.4): Text::censor

slik:
Since php 5.5.0 /e modifier for `preg_replace` is deprecated. So I've replaced `preg_replace` with `preg_replace_callback`. Now it passes `Text::censor` tests on 5.5.


lenton:
:+1: I've been getting these three errors in phpunit using PHP 5.5 too, glad someone's fixed them.

EDIT: I tried the fix and now I get no errors when running the unit tests.

PR #518, opened 2014-07-22
Bugfix (3.4): Text::readable_list

slik:
Fixed the data for `test_readable_list` and `test_readable_list` itself.
Also replaced `InvalidArgumentException` with `Kohana_Exception` because `Kohana_Exception` is the enter point of all Kohana exceptions. Tell me if I'm wrong.

### DIFF
--- a/classes/Kohana/Text.php
+++ b/classes/Kohana/Text.php
@@ -705,12 +705,12 @@ class Kohana_Text {
 			// Check that the word isn't an array itself.
 			if (is_array($word))
 			{
-				throw new InvalidArgumentException('The array must only have one dimension.');
+				throw new Kohana_Exception('The array must only have one dimension.');
 			}
 			// Check that the value of the word is appropriate.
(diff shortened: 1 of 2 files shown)


lenton:
I've tested this PR and it fixes the existing four failures caused by TextTest, +1 for merge!

classes/Kohana/Text.php
```diff
@@ -705,12 +705,12 @@ public static function readable_list(array $words, $conjunction = 'and', $serial
 			// Check that the word isn't an array itself.
 			if (is_array($word))
 			{
-				throw new InvalidArgumentException('The array must only have one dimension.');
+				throw new Kohana_Exception('The array must only have one dimension.');
```
shadowhand, line 708: why the change in exception type?

slik, line 708: @shadowhand 
The first reason is that `Kohana_Exception` is expected in the test, the second reason is in my comment to the PR: because Kohana_Exception is the enter point of all Kohana exceptions.

PR #519, opened 2014-07-23
Explicitly enable application and core modules

lenton:
The application and core modules must now be explicitly set in the bootstrap.php instead of magically being added in the core behind the scenes. This gives the user total control and flexibility over modules and makes all modules optional.

Kohana::include_paths() method has been removed, the only difference between this and Kohana::modules() is that it included the application and core module paths. Now that these two modules are included in Kohana::$_modules there's no need for the method to exist, likewise with Kohana::$_paths.

In the refactor of Kohana::modules(), array_reverse() is used before setting and getting modules. This makes it so Kohana::$_modules internally stores modules in the correct order, however externally it works in the same "top to bottom" fashion. As array_reverse() is now in one place, other methods such as find_file() don't have to worry about reversing the modules array before use.

### DIFF
--- a/classes/Kohana/Core.php
+++ b/classes/Kohana/Core.php
@@ -552,44 +552,46 @@ class Kohana_Core {
 	 */
 	public static function modules(array $modules = NULL)
 	{
-		if ($modules === NULL)
+		// If modules array has been passed
+		if ($modules !== NULL)
 		{
-			// Not changing modules, just return the current set
-			return Kohana::$_modules;
+			// Reverse modules before loading
+			$modules = array_reverse($modules);
+			
+			// Reset enabled modules
+			Kohana::$_modules = [];
+			
+			foreach ($modules as $name => $path)
+ 			{
+				// If module directory doesn't exist
+				if ( ! is_dir($path))
+				{
+					// Throw an error
+					throw new Kohana_Exception('Attempted to load an invalid or missing module \':module\' at \':path\'', [
+						':module' => $name,
+						':path'   => Debug::path($path),
+					]);
+				}
+				
+				// Get resolved, absolute path
+				$path = realpath($path).DIRECTORY_SEPARATOR;
+				
+				// Enable module
+				Kohana::$_modules[$name] = $path;
+				
+				// If module init file exists
+				$init_path = $path.'init'.EXT;
+				
+				if (is_file($init_path))
+				{
+					// Include the module initialization file
+					require_once $init_path;
+				}
+			}
 		}
 
-		foreach ($modules as $name => $path)
-		{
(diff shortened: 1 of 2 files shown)


classes/Kohana/Core.php
```diff
@@ -134,12 +134,7 @@ class Kohana_Core {
 	/**
 	 * @var  array   Currently active modules
 	 */
-	protected static $_modules = array();
-
-	/**
-	 * @var  array   Include paths that are used to find files
-	 */
-	protected static $_paths = array(APPPATH, SYSPATH);
+	protected static $_modules = [];
```
shadowhand, line 137: This makes this diff depend on #29, and is basically the exact reason I said that it doesn't make sense to replace everything at once.

classes/Kohana/Core.php
```diff
@@ -557,64 +552,46 @@ public static function auto_load_lowercase($class, $directory = 'classes')
...
-	/**
-	 * Returns the the currently active include paths, including the
-	 * application, system, and each module's path.
-	 *
-	 * @return  array
-	 */
-	public static function include_paths()
-	{
-		return Kohana::$_paths;
+		// Return enabled modules in reverse order
```
shadowhand, line 593: why is this being changed?

lenton, line 593: array_values(Kohana::include_paths()) === Kohana::modules()

As I explain in my description, it's not needed since the only difference is include_paths() isn't an associative array.

classes/Kohana/Core.php
lenton, line 137: Huh? These changes don't rely on any other PR.

acoulton, line 137: Yes they do - I think @shadowhand meant to link to kohana/kohana#29 - the array access syntax here requires PHP5.4, but aren't based on a branch where that dependency has been updated.

lenton, line 137: Oh right, yeah it does depend on this little change to core https://github.com/kohana/core/pull/515/files but updating to 5.4 has already been discussed and accepted so there's no chance it would get rejected. However I do accept that I should have used array() or waited for 515 to be merged, I'm still getting used to making small, non-dependant PRs.

classes/Kohana/Core.php
WinterSilence, line 593: Kohana::include_paths() != Kohana::modules() , paths also includes APPPATH, SYSPATH

lenton, line 593: This PR makes it so the user has to explicitly enable the application and system modules in their bootstrap, so APPPATH and SYSPATH end up in Kohana::modules() now too.

PR #520, opened 2014-07-25
Wrong request content length calculation, rewriting it on render.

functionundefined:
See: request::render() line: 1475

Wrong content length calculation. And not sure if it's ok rewriting length on render.
2853f71589e33788b109250d40df3293

If request has no body and only INFILE/PUT it returns 0. App fails when I debug request object with __string.


Ikke:
Hello Evgeny,

You have applied this pull request to an incorrect branch. This causes the pull request to contains a lot of incorrect commits.

In any case, always apply your changes against develop branches (for example, 3.2/develop), and also make sure you changes are based on that branch (rebase if necessary).

PR #521, opened 2014-07-26
Added encoding support to ucfirst method

Come2Daddy:
Uses mb_strtoupper and mb_substr to achieve Upper Case on first charater of each word along with encoding, following Text::ucfirst() behavior. Default encoding is UTF-8.

Exemple : Text::ucfirst('вы-рок', '-', 'UTF-8'); => 'Вы-Рок'

### DIFF
--- a/classes/Kohana/Text.php
+++ b/classes/Kohana/Text.php
@@ -251,10 +251,10 @@ class Kohana_Text {
 		// Put the keys back the Case-Convention expected
 		return implode($delimiter,
 			array_map(
-				create_function(
-					'$word, $encoding',
-					'return mb_strtoupper(mb_substr($word, 0, 1, $encoding), $encoding) . mb_strtolower(mb_substr($word, 1, mb_strlen($word), $encoding), $encoding);'
-					),
+				function($word, $encoding)
+				{
+				  return mb_strtoupper(mb_substr($word, 0, 1, $encoding), $encoding) . mb_strtolower(mb_substr($word, 1, mb_strlen($word), $encoding), $encoding);
+				},
 			$array,
 			array_fill(0, count($array), $encoding)
 			)
--- a/classes/Kohana/Text.php
+++ b/classes/Kohana/Text.php
@@ -243,7 +243,7 @@ class Kohana_Text {
 	* @param   string  $encoding   character encoding
 	* @return  string 
 	*/
-	static function ucfirst($string, $delimiter = '-', $encoding = 'UTF-8')
+	public static function ucfirst($string, $delimiter = '-', $encoding = 'UTF-8')
 	{
 		// Explode string
 		$array = explode($delimiter, $string);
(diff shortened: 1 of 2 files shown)


classes/Kohana/Text.php
```diff
@@ -233,19 +233,32 @@ public static function random($type = NULL, $length = 8)
...
-	 * delimiter or the default.
-	 *
-	 *      $str = Text::ucfirst('content-type'); // returns "Content-Type"
-	 *
-	 * @param   string  $string     string to transform
-	 * @param   string  $delimiter  delimiter to use
-	 * @return  string
-	 */
-	public static function ucfirst($string, $delimiter = '-')
+   	* Uppercase words that are not separated by spaces, using a custom
```
shadowhand: the alignment of the documentation go screwed up... please revert

classes/Kohana/Text.php
```diff
@@ -233,19 +233,32 @@ public static function random($type = NULL, $length = 8)
...
+	public static function ucfirst($string, $delimiter = '-', $encoding = 'UTF-8')
 	{
+		// Explode string
+		$array = explode($delimiter, $string);
+		
 		// Put the keys back the Case-Convention expected
-		return implode($delimiter, array_map('ucfirst', explode($delimiter, $string)));
+		return implode($delimiter,
+			array_map(
+				create_function(
```
shadowhand: please do not use create function... either use a private method or a closure.

shadowhand:
Love the concept of this PR. Excellent stuff!

Come2Daddy:
I'm not yet familiar with github and its tools, sorry for messing up doc.

classes/Kohana/Text.php
```diff
@@ -233,19 +233,30 @@ public static function random($type = NULL, $length = 8)
...
+	*/
 	public static function ucfirst($string, $delimiter = '-')
 	{
 		// Put the keys back the Case-Convention expected
-		return implode($delimiter, array_map('ucfirst', explode($delimiter, $string)));
+		return implode($delimiter,
+			array_map(
+				function($word)
+				{
+				  return UTF8::strtoupper(UTF8::substr($word, 0, 1)) . UTF8::strtolower(UTF8::substr($word, 1, UTF8::strlen($word)));
```
rjd22: I'm not sure if it is good practice to use other library functions directly like this in other libraries. I will cause errors later on when changes are made to the other libraries. @shadowhand what do you think?

Ikke: These are 'simple' functions with a stable interface. I see no problems in using them. Otherwise, you'd need to duplicate the work done in those libraries.

enov:
Should this be fixed in 3.3? As the way we have `Text::ucfirst` right now, not supporting UTF-8, feels like a bug.

enov:
cc @acoulton

acoulton:
@enov yeah, it does feel like a 3.3 bug. Looks like it would be easy enough to cherry-pick this fix back into 3.3 - note there's also an unrelated Arr::get change crept in here.

PR #522, opened 2014-07-29
Remove automatic enablement of application and core modules

lenton:
This PR makes it so the application module and core module must be explicitly enabled in bootstrap.php alongside other modules. This gives the user total control and flexibility when enabling modules.

### DIFF
--- a/classes/Kohana/Core.php
+++ b/classes/Kohana/Core.php
@@ -129,7 +129,7 @@ class Kohana_Core {
 	/**
 	 * @var  array   Include paths that are used to find files
 	 */
-	protected static $_paths = array(APPPATH, SYSPATH);
+	protected static $_paths = [];
 
 	/**
 	 * @var  array   File path cache, used when caching is true in [Kohana::init]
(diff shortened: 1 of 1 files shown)


classes/Kohana/Core.php
```diff
@@ -505,11 +502,8 @@ public static function modules(array $modules = NULL)
 			}
 		}
 
-		// Finish the include paths by adding SYSPATH
-		$paths[] = SYSPATH;
-
 		// Set the new include paths
-		Kohana::$_paths = $paths;
+		Kohana::$_paths = array_values($modules);
```
shadowhand, line 506: doesn't this PR make it possible to dump `$_paths` completely and always refer to `$_modules`?

lenton, line 506: Yes, I did remove it in my previous PR but I thought there was too much going on making it hard to understand, so I decided to split it up. If this PR gets merged then the replacement of $_paths with $_modules would be the next PR.

shadowhand, line 506: And what about an accompanying PR for kohana/kohana to add app and sys to the `Kohana::modules([])` call?

acoulton:
@lenton this does make it a bit harder to dynamically add modules after the initial call, because they now have to be inserted manually in position above SYSPATH rather than Kohana always ensuring that SYSPATH is at the bottom.

I think this needs some changelog documentation, as there may well be end-user code out there that is doing the same as the unittest module (eg `Kohana::modules(Kohana::modules() + array('unittest' => MODPATH.'unittest'));`) which may not work as expected following an upgrade.

PR #525, opened 2014-08-04
Don't use classes from modules until all modules are successfully enabled

lenton:
Kohana_Exception and Debug::path can not be used until all modules are successfully enabled.

### DIFF
--- a/classes/Kohana/Core.php
+++ b/classes/Kohana/Core.php
@@ -489,10 +489,7 @@ class Kohana_Core {
 			else
 			{
 				// This module is invalid, remove it
-				throw new Kohana_Exception('Attempted to load an invalid or missing module \':module\' at \':path\'', array(
-					':module' => $name,
-					':path'   => Debug::path($path),
-				));
+				throw new Exception('Attempted to load an invalid or missing module "'.$name.'" at "'.$path.'"');
 			}
 		}
 
--- a/tests/kohana/CoreTest.php
+++ b/tests/kohana/CoreTest.php
@@ -241,7 +241,7 @@ class Kohana_CoreTest extends Unittest_TestCase
 	 *
 	 * @test
 	 * @dataProvider provider_modules_detects_invalid_modules
-	 * @expectedException Kohana_Exception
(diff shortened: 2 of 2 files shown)


classes/Kohana/Core.php
```diff
@@ -489,10 +489,7 @@ public static function modules(array $modules = NULL)
 			else
 			{
 				// This module is invalid, remove it
-				throw new Kohana_Exception('Attempted to load an invalid or missing module \':module\' at \':path\'', array(
-					':module' => $name,
-					':path'   => Debug::path($path),
-				));
+				throw new Exception('Attempted to load an invalid or missing module "'.$name.'" at "'.$path.'"');
```
sergeyklay, line 492: Okay, but it is impossible to localize

lenton, line 492: Yes, but that's unfortunately something we have to live with, we can not use what we haven't loaded yet.

sergeyklay, line 492: however u can use `sprintf` with `%s`

enov:
You are not solving anything here. What about userland not being able to throw `Kohana_Exception` in `init.php` of their modules?

You still need to come up with a solution for function `__` to be available in `Kohana_Exception`, before we think what to do here.

lenton:
All modules are enabled before init.php files are included so Kohana_Exception will be available. As for the __ function not being available, that's a separate issue. This PR solves a different issue of Kohana_Exception not being available before the modules are enabled.

classes/Kohana/Core.php
enov, line 492: please use `strtr` and try to keep the format of the code

lenton, line 492: Is that really necessary? This is a standard php exception, it shouldn't have to follow the same format as Kohana_Exception.

enov, line 492: No, not really necessary, if you don't have the time for that, I'll merge this anyways.

lenton, line 492: I think it's best to keep it a simple string, there's really no benefit to using strtr.

Issue #526, opened 2014-08-06
Deprecate `__` in favor of `I18n::__`


enov:
I am not sure if this is a good idea, so I want your input here.

kohana/kohana#31 has made the core to be loaded in `Kohana::modules()`, which makes the function `__` unavailable for `Kohana_Exception`.

We can either test its availability in `Kohana_Exception` or deprecate/replace `__` with `I18n::__`.

We can still leave `__` available for standard views, maybe it will be just an alias for `I18n::__`. `__` can be defined in bootstrap or we can keep it where it is now.

I don't know if I am making sense. You input is appreciated. Thanks!

lenton:
I realise now what you've been trying to say about the __ function not being included, I thought you was talking about a different problem.

I'm in favour of moving the logic into a method (although I think a better name would be I18n::trans()) and then define the __ function in the bootstrap above the Kohana::methods() call.

enov:
> I realise now what you've been trying to say about the __ function not being included, I thought you was talking about a different problem.

It's OK. But I think it's better to be a friendly developer than a good developer ;)

And btw, you need to define the function `__` below `Kohana::modules` call, as it uses the `I18n` class.

lenton:
The function can be defined before Kohana::modules() without any problems, that way it can be used inside init.php files too.

enov:
Alright then.

enov:
But we need to replace occurrences of `__` with `I18n::translate` so that the `kohana/core` would be self contained and not relying on bootstrap. We can still keep `__` for Kohana users for their standard views.

sergeyklay:
just `__` mb as alias for `I18n::translate`

lenton:
I was thinking, maybe it's best to move the __() function back into the I18n class (still deprecated of course). It would make the upgrading process easier since users wouldn't have to add the function to their bootstrap; they wouldn't need to do anything.

rjd22:
@lenton I don't agree because of the global scope pollution because of this feature. I have collisions because of this function with other libraries before.

lenton:
@rjd22 I don't think that's much of a problem since it's rare and there are multiple ways for users to easily solve the problem:
- Kohana uses `if ( ! function_exists('__'))` so you could declare the library's __() function before Kohana does.
- Extend the Kohana_I18n class via the cascading filesystem to hide the __() function declaration.
- Remove the __()  function declaration directly from the core's Kohana_I18n class. It's deprecated and is due to be removed in a future version so it's ok.

[...]

PR #527, opened 2014-08-06
Move __ function to I18n::translate()

lenton:
I will define the __ function in the bootstrap once enov's revert gets reverted in kohana/kohana.


enov:
We need to mention in the docs that the `__` function is now _half_ deprecated: it can not be used in the `init.php` of the modules, and should be properly included in the bootstrap to be used in standard views.

enov:
related #526

PR #528, opened 2014-08-07
Update Encrypt.php

sarciszewski:
`mt_srand()` doesn't touch `rand()` it only affects `mt_rand()`, which is not called when `MCRYPT_RAND` is passed to `mcrypt_create_iv()`.

So instead, let's pass it on to a userland RNG.

### DIFF
--- a/classes/Kohana/Encrypt.php
+++ b/classes/Kohana/Encrypt.php
@@ -151,7 +151,7 @@ class Kohana_Encrypt {
 				}
 				else
 				{
-					// Use the system random number generator
+					// Use an insecure random number generator
 					Encrypt::$_rand = MCRYPT_RAND;
 				}
 			}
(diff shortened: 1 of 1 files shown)


cs278:
I raised an issue about the seeding ages ago. Personally I'd just say if the user does not have a CSPRNG available the Encrypt class should just abort, why offer the false sense of security of broken encryption?

sarciszewski:
@cs278 :+1: I fully agree. The inability to obtain reliable entropy from the kernel is in and of itself a security-affecting vulnerability, even if it's not immediately exploitable. I'm willing to commit+rebase this without a non-CS PRNG if the Kohana team agrees as well.

acoulton:
:thumbsup:

rjd22:
@sarciszewski Could you make a new pull request of what you are proposing?

sarciszewski:
Sure. Will do. :)

PR #529, opened 2014-08-07
Remove direct access checks

lenton:
Issue: https://github.com/kohana/core/issues/508

### DIFF
--- a/classes/Kohana/Core.php
+++ b/classes/Kohana/Core.php
@@ -1,4 +1,4 @@
-<?php defined('SYSPATH') OR die('No direct script access.');
+<?php
 /**
  * Contains the most low-level helpers methods in Kohana:
  *
@@ -25,7 +25,11 @@ class Kohana_Core {
 	const TESTING     = 30;
 	const DEVELOPMENT = 40;
 
-	// Security check that is added to all generated PHP files
+	/**
+	 * Security check that is added to all generated PHP files
+	 * 
+	 * @deprecated 3.4 Direct access checks are no longer encouraged
+	 */
 	const FILE_SECURITY = '<?php defined(\'SYSPATH\') OR die(\'No direct script access.\');';
 
 	// Format of cache files: header, cache name, and data
--- a/classes/Kohana/Log/File.php
+++ b/classes/Kohana/Log/File.php
@@ -78,7 +78,7 @@ class Kohana_Log_File extends Log_Writer {
 		if ( ! file_exists($filename))
 		{
 			// Create the log file
-			file_put_contents($filename, '');
+			file_put_contents($filename, '<?php exit; ?>'.PHP_EOL.PHP_EOL);
 
 			// Allow anyone to write to log files
 			chmod($filename, 0666);
--- a/classes/Kohana/Log/File.php
+++ b/classes/Kohana/Log/File.php
@@ -1,4 +1,4 @@
-<?php defined('SYSPATH') OR die('No direct script access.');
+<?php
 /**
  * File log writer. Writes out messages and stores them in a YYYY/MM directory.
  *
@@ -78,7 +78,7 @@ class Kohana_Log_File extends Log_Writer {
 		if ( ! file_exists($filename))
 		{
 			// Create the log file
-			file_put_contents($filename, Kohana::FILE_SECURITY.' ?>'.PHP_EOL);
+			file_put_contents($filename, '');
 
 			// Allow anyone to write to log files
 			chmod($filename, 0666);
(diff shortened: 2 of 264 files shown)


classes/Kohana/Core.php
```diff
@@ -25,7 +25,11 @@ class Kohana_Core {
 	const TESTING     = 30;
 	const DEVELOPMENT = 40;
 
-	// Security check that is added to all generated PHP files
+	/**
+	 * Security check that is added to all generated PHP files
+	 * 
+	 * @deprecated 3.4 Direct access checks are no longer encouraged
+	 */
 	const FILE_SECURITY = '<?php defined(\'SYSPATH\') OR die(\'No direct script access.\');';
```
shadowhand, line 33: the usage of this constant in the logs should also be removed, right?

classes/Kohana/Log/File.php
```diff
@@ -78,7 +78,7 @@ public function write(array $messages)
 		if ( ! file_exists($filename))
 		{
 			// Create the log file
-			file_put_contents($filename, Kohana::FILE_SECURITY.' ?>'.PHP_EOL);
```
shadowhand, line 81: nevermind, here it is...

classes/Kohana/Log/File.php
```diff
@@ -78,7 +78,7 @@ public function write(array $messages)
...
+			file_put_contents($filename, '');
```
shadowhand: this change means that we will end up with a .php file that does not contain any PHP code

lenton: That's true, but it causes no problems. Maybe the logs file type can be changed to a more suitable one in a future PR.

kemo: `.log.php` will have to either be ignored by the server or `die` should be appended to each files' body, I don't see a valid reason to have it available or why someone would include it elsewhere in the app codebase

acoulton: Presumably this was just to prevent the logs being accessible over HTTP with the default docroot. Since the log files aren't actually PHP, and given users will now need to be sure all their /application etc files are not directly accessible, I'd say makes most sense to make the log files .log in future.

kemo: They should really be logged in /var/log/appname but there, they're not ;)

rjd22: This change will make the log file available to the public no? So it is better to disable direct script access there or just simply `<?php die();` on top of them ;)

lenton: @rjd22 This PR makes all classes and logs directly accessible to the public, that's the point. It will be followed by (or something similar) adding a 'web' directory in kohana/kohana which contains index.php and other public files. This way, all private files will be inaccessible by the web server if set up correctly.

rjd22: @lenton this can create a big security risk for the time being. I don't think basing a PR on the future is a good idea. Could you at least secure them for now and in the next pull request where the web directory is implemented remove it again.

That way we won't have any temporary security issues.

lenton: @rjd22 This PR doesn't have to be merged right away, I too think its a good idea to secure kohana/kohana first before merging.

acoulton: @lenton disagree. This PR should be merged ASAP or we're just storing up merge conflicts. 

Moving the kohana/kohana document root down to 'web' by default needs to be a separate discussion. I'm in favour, but it's been discussed before and there were valid reasons not to (essentially that we wanted to cater to people who just drop the whole project on a shared host, who may not be able/aware to make it so "all private files will be inaccessible by the web server if set up correctly").

shadowhand: i agree with @acoulton... this should be merged quickly. leaving the security message in logs, or replacing it with `<?php die(); ?>` is a good solution.

lenton: @shadowhand Why add die() to the top of logs, no one is going to be actually using 3.4 in production.

shadowhand: @lenton to prevent direct access via `http://example.com/application/logs/date.log.php` which could leak sensitive data like tokens or passwords.

acoulton: @lenton even if nobody is using 3.4 in production yet, if you take out the check now somebody has to be sure to reorganise the project structure in future. If for whatever reason that doesn't happen before someone deploys a dev/beta branch somewhere that matters then that's a problem.

If you leave the check or switch to simple die() now it can always be changed later, but we'll not have lost any security in the interim.

cyrrill: > @rjd22 https://github.com/rjd22 This PR doesn't have to be merged right
> away, I too think its a good idea to secure kohana/kohana first before
> merging.

Every commit needs to be atomic, leaving the branch in a consistent state.
Foregoing consistency until a future commit is inadmissible.

kemo: :+1:  for `<?php die; ?>` "security headers" in logs

lenton:
I've added the exit to the top of log files, it's now ready to be merged.

Issue #524, opened 2014-07-29
Set routes in a 'latter overwrite' manner


lenton:
I propose that routes be set in a 'later routes overwrite' manner instead of the current 'first come, first serve'.

It's a problem for two reasons:
- It's not logically intuitive to programmers, we are used to later defined things overwriting.
- It forces modules to be loaded in the wrong (reverse) order.

When modules can be loaded in the correct order, later initialisation files (such as in `application`) can manipulate objects created by previous initialisation files.

enov:
maybe separate Routes into `routes.php` and have them loaded top to bottom?

lenton:
@enov That would still mean you can't interact with routes which have been set in a lower module and  it would break all existing modules which use routes.

enov:
You mean we can't interact with routes from `init.php`?

lenton:
Nope, from a route.php which is in a higher module.

I think it's best to fix this one at the source of the problem.

acoulton:
Just to be clear, are you proposing that as well as making `Route::set` overwrite a previously defined route with the same name, we also reverse the sequence of calling each module's init.php?

PR #530, opened 2014-08-07
Create new method Kohana::init_modules()

lenton:
This PR splits the module initialisation functionality from Kohana::modules() and places it into its own method called Kohana::init_modules().

Previous PRs to the 3.4/develop branch has forced Kohana::modules() to be called in the bootstrap before any classes are used. This causes a problem because a few things need to be done before the modules are initialised such as attaching a config object to Kohana::$config (the userguide uses this for example https://github.com/kohana/userguide/blob/3.3/master/init.php#L12).

To fix this problem, the method Kohana::init_modules() can be called after everything has been set up for init files in the bootstrap.

### DIFF
--- a/classes/Kohana/Core.php
+++ b/classes/Kohana/Core.php
@@ -499,18 +499,29 @@ class Kohana_Core {
 		// Set the current module list
 		Kohana::$_modules = $modules;
 
+		return Kohana::$_modules;
+	}
+	
+	/**
+	 * Initializes the enabled modules.
+	 * 
+	 * @return void
+	 */
+	public static function init_modules()
+	{
+		// For each enabled module
 		foreach (Kohana::$_modules as $path)
 		{
-			$init = $path.'init'.EXT;
+			// Get init file path
+			$init_path = $path.'init'.EXT;
 
-			if (is_file($init))
+			// If init file exists in module
+			if (is_file($init_path))
 			{
-				// Include the module initialization file once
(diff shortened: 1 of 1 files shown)


enov:
@lenton adding the core as a simple module is making things slightly complicated.

Theoretically speaking, using classes of the modules (and core) should be done after proper initialization of modules (and core).

acoulton:
@lenton I'm also not sure about this - some modules add autoloaders or otherwise register additional classes outside the CFS. It has potential to be slightly confusing if the system is part usable (and some classes have to be used) between Kohana::modules and Kohana::init_modules.

This and the issues over __/I18N makes me think that maybe we should just revert to core being available by default.

lenton:
This makes the module's init files load at the same time as they do in 3.3 so there's going to be absolutely no problems for existing init files out there. It also completely solves the __() issue and makes it available to init files too.

enov:
I'll merge this soon if there are no other objection on this.

enov:
@lenton , what's the order of running the `init.php` of the modules?

I suggest, if it is not already the case, that lower modules (those defined later) should have their `init.php` called/required before upper (those defined first) modules.

lenton:
@enov Yes, that's much more logical but unfortunately Kohana routes have to be set in a top down fashion which means init files can't be loaded in the correct order yet: https://github.com/kohana/core/issues/524.

PR #532, opened 2014-08-18
// script urls supported in HTML::script

danielvalentin:
support script URLs that that do not implicitly specify the protocol, i.e. starting with //. Ref: https://github.com/kohana/kohana/issues/40

### DIFF
--- a/classes/Kohana/HTML.php
+++ b/classes/Kohana/HTML.php
@@ -206,7 +206,7 @@ class Kohana_HTML {
 	 */
 	public static function style($file, array $attributes = NULL, $protocol = NULL, $index = FALSE)
 	{
-		if (strpos($file, '://') === FALSE)
+		if (strpos($file, '://') === FALSE AND strpos($file, '//') !== 0)
 		{
 			// Add the base URL
 			$file = URL::site($file, $protocol, $index);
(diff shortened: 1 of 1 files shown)


classes/Kohana/HTML.php
```diff
@@ -239,7 +239,7 @@ public static function style($file, array $attributes = NULL, $protocol = NULL,
 	 */
 	public static function script($file, array $attributes = NULL, $protocol = NULL, $index = FALSE)
 	{
-		if (strpos($file, '://') === FALSE)
+		if (strpos($file, '://') === FALSE && strpos($file, '//') !== 0)
```
shadowhand: Kohana uses `AND` instead of `&&`.

classes/Kohana/HTML.php
```diff
@@ -239,7 +239,7 @@ public static function style($file, array $attributes = NULL, $protocol = NULL,
 	 */
 	public static function script($file, array $attributes = NULL, $protocol = NULL, $index = FALSE)
```
shadowhand, line 240: can you also apply this to `HTML::style` please?

danielvalentin:
You bet'cha

rjd22:
@Sakto I think we have unit tests for this part of kohana. Could you also add the appropiate failing unit tests and make them pass with your commit?

enov:
BTW, this is targeting `3.3/master`, it should target `3.3/develop`

danielvalentin:
I still have a lot to learn :] Thanks for the heads ups, I'll get right on it.

rjd22:
Pointing to wrong branch. :)

enov:
@Sakto could we have a PR on the right branch with failing unit tests included? :smile:

arkagod:
Faster -  AND strncmp($file, '//', 2)

danielvalentin:
Hey @arkagod - This request is deprecated in favour of this one: https://github.com/kohana/core/pull/549

Good call btw. you should open a pull request!

PR #533, opened 2014-08-24
Allowing Config to be extended and set before Kohana::init

rafi:
`Kohana::$config` is allowed to be set before `Kohana::init()`, but it checks if it's a `Config` instance. I've changed the check to `Kohana_Config` to eliminate the need to extend `Config` and instead extend `Kohana_Config` directly.

Makes sense?

### DIFF
--- a/classes/Kohana/Core.php
+++ b/classes/Kohana/Core.php
@@ -307,7 +307,7 @@ class Kohana_Core {
 		}
 
 		// Load the config if one doesn't already exist
-		if ( ! Kohana::$config instanceof Config)
+		if ( ! Kohana::$config instanceof Kohana_Config)
 		{
 			Kohana::$config = new Config;
 		}


kemo:
Maybe `Kohana_Config` (and the rest of `Kohana_` classes) should therefore be declared abstract?

rafi:
@kemo That's an option. But for a different issue :)

enov:
Hey @rafi !

I believe `Config` is already allowed to be extended. The typical Kohana way of extending things, is that we extend the `Kohana_Config` class and name it `Config` by placing it somewhere above in the CFS.

I am not sure what I am missing here. This change _forces_ us to extend `Kohana_Config` in the typical way described above, but for one additional benefit is that it allows us to name the newly extended class something different than `Config`. I don't know if that's a huge benefit, and I am not sure about the repercussions.

Please clarify if I am missing something.

rafi:
Hi @enov - indeed the benefit would be naming the extension differently and placing it in a namespace. `Kohana::init` allows a dev to instantiate a different Config or Log object into the statics $config and $log [here](https://github.com/kohana/core/blob/3.3/master/classes/Kohana/Core.php#L332-L342). So might as well change `Log` too.

What d'you think?

enov:
Considering the fact that we are only eliminating the need to extend `Config`, in order to extend `Kohana_Config` directly, as you have rightfully stated at the beginning of this discussion, I am not seeing huge benefits.

And still, you can extend `Config` and have your beautifully namespaced class instantiated before `Kohana::init` - but, yeah, it's an additional empty class in between...

The downside of this is that it breaks any code that implemented `Config` without ever extending `Kohana_Config`. That might be downright weird, and practically unexistant. But, I can not in clear conscience assume that it does not exist.

Allow me to politely decline this :) I'll just wait here a little bit for any additional comments before closing.

Thanks!

rjd22:
Doesn't this break the class overloading for the CFS? Normally you would do `class Config extends Kohana_Config` to seamlessly overload the Config class and add your own logic to it.

enov:
@rjd22 , I think this does not break if you do `class Config extends Kohana_Config`, but only if you do `class Config` without extending `Kohana_Config`, because this is changing it to check the latter.

rafi:
@enov my whole point is that you **must** extend `Config`. You wouldn't be able say, to declare `class Config_Engine extends Kohana_Config` -- because `Kohana::init()` verifies it's an instance of `Config`. It's rare for someone to set `::$config` or `::$log` before init, but the possibility is there for a reason. I think it's such a small change with a big benefit.

enov:
@rafi , what is the downside of extending `Config`? Wouldn't `class Config_Engine extends Config` work for you?

lenton:
I don't see any downsides to this change, so why not let users extend Kohana_Config directly before Kohana::init().

I think it should also be done for the log check too if this is to be accepted.

enov:
@rafi , @lenton , I closed because I did not see a huge benefit to extend `Kohana_Config` directly vs. being forced to extend `Config`. And there is a risk to break code, in case it defined some `Config` through the CFS without extending `Kohana_Config`.

The optimal OOP way is to have an interface defined that `Kohana_Config` implements. In this case, the framework checks if the object is of a class that implements that interface, similar to a DI pattern.

PR #535, opened 2014-08-28
3.4/bugfix/use i18n translate

enov:
I think `kohana/core` should be a self sufficient module, not relying on `__()` function which is defined outside the module, in `bootstrap.php`.

I replaced occurrences of `__()` with `I18n::translate`. Also made some doc adjustments.

Kohana userland can still use `__()` if they leave the declaration in bootstrap. Otherwise they should use `I18n::translate` instead.

Please review thoroughly.

### DIFF
--- a/classes/Kohana/I18n.php
+++ b/classes/Kohana/I18n.php
@@ -6,6 +6,8 @@
  * Typically this class would never be used directly, but used via the __()
  * function, which loads the message and replaces parameters:
  *
+ * the __() function is declared in the Kohana official bootstrap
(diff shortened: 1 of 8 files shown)


lenton:
+1

mintbridge:
+1

enov:
Should I continue merging my own PRs? I wonder maybe it's healthier to have another person - not the author - to review and merge?

shadowhand:
:+1: reviewed and okay to merge.

PR #444, opened 2014-01-24
Fixes #4813 by Using Kohana_UTF8 from version 3.3 to apply all the patches

enov:
Fixes Kohana_UTF8Test::test_clean with data set no.3
and Kohana_DebugTest::test_dump with data set no.6

PR #537, opened 2014-08-29
Set and reset mb_substitute_character for mb_convert_encoding

enov:
This was accepted as part of 3.2 bugfix here #444 

I am not sure if we want to do the same for 3.3. This makes the `mb_substitute_character('none');` line in `bootstrap.php` redundant as it sets and resets back the original value within the function, when needed. So, I will make another PR for `kohana/kohana` on 3.4/develop and remove that line from `bootstrap.php`.

Regarding Kohana 3.3, I am not sure if we consider the settings of `bootstrap.php` file as something we could not change on minor releases. Probably removing `mb_substitute_character('none');` line from `bootstrap.php` might break some code relying on `mb_` functions elsewhere in the application. So let me know if you think this should be in Kohana 3.3 or not.

Not very elegant solution, but at least better than disabling and enabling the error reporting that we had prior to v. 3.2.3.

Thanks for reviewing :)

### DIFF
--- a/classes/Kohana/UTF8.php
+++ b/classes/Kohana/UTF8.php
@@ -85,13 +85,17 @@ abstract class Kohana_UTF8 {
 
 			if ( ! UTF8::is_ascii($var) AND UTF8::$server_utf8)
 			{
-				// Disable notices
-				$error_reporting = error_reporting(~E_NOTICE);
+				// Temporarily save the mb_substitute_character() value into a variable
+				$mb_substitute_character = mb_substitute_character();
 
+				// Disable substituting illegal characters with the default '?' character
+				mb_substitute_character('none');
+
+				// convert encoding, this is expensive, used when $var is not ASCII
 				$var = mb_convert_encoding($var, $charset, $charset);
 
-				// Turn notices back on
-				error_reporting($error_reporting);
+				// Reset mb_substitute_character() value back to the original setting
+				mb_substitute_character($mb_substitute_character);
 			}
 		}
 


enov:
no love for this one?

classes/Kohana/UTF8.php
```diff
@@ -85,13 +85,17 @@ public static function clean($var, $charset = NULL)
 
 			if ( ! UTF8::is_ascii($var) AND UTF8::$server_utf8)
 			{
-				// Disable notices
-				$error_reporting = error_reporting(~E_NOTICE);
```
shadowhand, line 89: this is being lost... maybe that's not ideal? there must have been a reason to disable notices...

enov, line 89: We used to have `iconv` instead of `mb_substitute_character`, the former used to generate `E_NOTICE` for illegal characters.

@rjd22 found that `mb_substitute_character` is a better solution to handle such conversions, but he did not remove those notice suppressions:

https://github.com/kohana/core/commit/2c70e8b66c7b24445480314f9f67987d4583229b#diff-7164730aa49bcfa52c92ab1167d7d051

rjd22, line 89: Ah sorry about that @enov . Indeed I replaced it with `mb_substitute_character` the notice suppression is not needed since the documentation tells it doesn't throw notices: http://php.net/manual/en/function.mb-convert-encoding.php

shadowhand:
:shipit:

classes/Kohana/UTF8.php
rjd22, line 89: My thought at that time was that `mb_substitute_character('none');` should be set in the bootstrap as it is a global setting. I don't really agree with setting it here to none and switching it back later.

rjd22:
@shadowhand Can you check my comments? What is your opinion? @Zeelot and I had a long discussion about how it should work including setting the `mb_substitute_character('none');` in bootstrap and not in the module itself.

classes/Kohana/UTF8.php
shadowhand, line 89: @rjd22 okay that's probably a reasonable thing to do.

enov:
Thanks for your feedback @rjd22.

Yes, you are right that global settings should go into the bootstrap and not into the module. But, note that I am not using `mb_substitute_character('none');` as a setting.

I am linking here the discussion that happened about this here:
https://github.com/kohana/kohana/pull/12
Also, my notes in Redmine for v3.2.3 PR here:
http://dev.kohanaframework.org/issues/4813#note-3

Basically, I agree with @EvanPurkhiser [here](https://github.com/kohana/kohana/pull/12#discussion_r4832287) and [here](https://github.com/kohana/kohana/pull/12#discussion_r4835030) that this is the best solution, if we want to keep `UTF8::clean()` to silently discard invalid characters.

Otherwise, if we want to keep `mb_substitute_character('none');` as a setting in bootstrap, we should change the goal/description of `UTF8::clean()` (the "silently discarding" part) and add to the comment above the bootstrap setting that this affects `UTF8::clean()`.

What if `mb_convert_encoding` had an additional argument that would override the global `mb_substitute_character` setting... we probably would not have this argument :)

Cheers!

enov:
> we probably would not have this argument

Quoting myself. I meant _conversation_. Not a native English speaker, chose the wrong word.

So @shadowhand , @rjd22 and also please @EvanPurkhiser; Could we agree:

1- If we want to keep `mb_substitute_character('none');` as a setting in bootstrap; update the comment above the setting to show that it affects `UTF8::clean()`; and update the docs for `UTF8::clean()`

OR

2- If we want to accept this PR and remove `mb_substitute_character('none');` from bootstrap?

Thanks

shadowhand:
Without digging in too deeply, I think option 2 makes the most sense, as it will be less error prone than someone changing the value in bootstrap and then getting unexpected results from `UTF8::clean`.

classes/Kohana/UTF8.php
sergeyklay, line 89: I suggest using `mbstring` purely  instead `icov` at all for it's widespread functionalities and ease of use

rjd22:
@shadowhand Okay, I think we can merge it in then. @enov can you also make a pull request at the 3.4/develop branch of kohana/kohana that removes the `mb_substitute_character('none');` from bootstrap?

enov:
Thanks @rjd22 . Here's the PR for bootstrap :)

https://github.com/kohana/kohana/issues/48

PR #540, opened 2014-09-04
Update Encrypt.php

sarciszewski:
- Not too long ago, @defuse informed me that `MCRYPT_DEV_URANDOM` on Windows works as well as it does on BSD/Linux.
- Don't allow insecure, predictable random numbers for IVs

### DIFF
--- a/classes/Kohana/Encrypt.php
+++ b/classes/Kohana/Encrypt.php
@@ -129,29 +129,15 @@ class Kohana_Encrypt {
 	 */
 	public function encode($data)
 	{
-		// Set the rand type if it has not already been set
+		// Set the rand type if it has not already been set// Set the rand type if it has not already been set
 		if (Encrypt::$_rand === NULL)
 		{
-			if (defined('MCRYPT_DEV_URANDOM'))
-			{
-				// Use /dev/urandom
-				Encrypt::$_rand = MCRYPT_DEV_URANDOM;
-			}
-			elseif (defined('MCRYPT_DEV_RANDOM'))
-			{
-				// Use /dev/random
-				Encrypt::$_rand = MCRYPT_DEV_RANDOM;
-			}
-			else
-			{
-				// Use the system random number generator
-				Encrypt::$_rand = MCRYPT_RAND;
-			}
+		    Encrypt::$_rand = MCRYPT_DEV_URANDOM;
 		}
-
-		if (Encrypt::$_rand === MCRYPT_RAND)
+		// Do not allow insecure random number generator
+		if ((Encrypt::$_rand !== MCRYPT_DEV_URANDOM) AND (Encrypt::$_rand !== MCRYPT_DEV_RANDOM))
 		{
-			throw new Kohana_Exception("Insecure random number generator chosen")
+		    throw new Kohana_Exception("Insecure random number generator chosen");
 		}
 
 		// Create a random initialization vector of the proper size for the current cipher
--- a/classes/Kohana/Encrypt.php
+++ b/classes/Kohana/Encrypt.php
@@ -132,36 +132,26 @@ class Kohana_Encrypt {
 		// Set the rand type if it has not already been set
 		if (Encrypt::$_rand === NULL)
 		{
-			if (Kohana::$is_windows)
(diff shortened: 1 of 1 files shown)


enov:
Thanks @sarciszewski , few questions:
- Should we add to our "installer" that we are requiring `MCRYPT_DEV_URANDOM` or `MCRYPT_DEV_RANDOM`? see https://github.com/kohana/kohana/blob/3.4/develop/install.php#L200-L207
- Practically speaking, are there any systems that do not have `MCRYPT_DEV_URANDOM` or `MCRYPT_DEV_RANDOM`? From what I read, there is not a *nix distro without, and PHP seems to bundle it for Windows.
- Why `Encrypt::$_rand = MCRYPT_RAND;` and then `throw Exception`? Why not `throw Exception` on `else`?
- Can the developer bypass the `Exception` by deliberately `Encrypt::$_rand = MCRYPT_RAND;` in `bootstrap.php`?

Thanks for the hard work!

enov:
Hmm.. Probably my 3rd and 4th questions answer each other... It seems you are not allowing the developer to bypass the `Exception` with `Encrypt::$_rand = MCRYPT_RAND;` in bootstrap...

sarciszewski:
- Yes, that would prevent unexpected surprises down the line.
- PHP >= 5.3.0 will have them: http://php.net/manual/en/function.mcrypt-create-iv.php#refsect1-function.mcrypt-create-iv-changelog

`MCRYPT_RAND` is unsafe. In any instance that they are selected, it will throw an exception and abort. `MCRYPT_RAND` is simply not acceptable to use: https://wiki.php.net/rfc/deprecate_mcrypt_rand

enov:
Thank you for your reply.

> PHP >= 5.3.0 will have them

We are requiring PHP >= 5.4 for Kohana v.3.4. If both _random number generation_ [edited] sources are available, wouldn't the code be reduced to:

<pre>
// Set the rand type if it has not already been set
if (Encrypt::$_rand === NULL)
{
    Encrypt::$_rand = MCRYPT_DEV_URANDOM;
}
// Do not allow insecure random number generator
if ((Encrypt::$_rand !== MCRYPT_DEV_URANDOM) AND (Encrypt::$_rand !== MCRYPT_DEV_RANDOM))
{
    throw new Kohana_Exception("Insecure random number generator chosen");
}
</pre>

Thanks!

sarciszewski:
Yep. That would certainly resolve the issue.

enov:
Thank you @sarciszewski .

Just a few more issues that are coming to my mind:
1. If we want to have this in Kohana 3.3.3
   - This theoretically changes the API (the method now throws a `Kohana_Exception`), usually we don't change API in minor releases. @shadowhand , should we have this in 3.3.3?
   - Does this break an existing application? Well it will throw the exception if the app is using `MCRYPT_RAND`. But I am asking if it breaks an app after changing `Encrypt::$_rand` to start using `MCRYPT_DEV_URANDOM`. My guess is that, since both are just different sources for random numbers, the app will continue to operate. Maybe @sarciszewski , if you could give us a better insight here?
2. Another issue is that there are no unit tests covering the Encrypt class

Thanks again :)

sarciszewski:
We could silently rewrite `MCRYPT_RAND` to `MCRYPT_DEV_URANDOM` too, if it exists, as a BC failsafe. :)

rjd22:
@enov This can't be merged into 3.3.3 since it is changing the API.

@sarciszewski would it be possible to write unit tests for it? We want to be sure it is working.

enov:
@rjd22 

There are no unit tests for the `Encrypt` class at all. I manually tested it, locally, before merging.

Also, @sarciszewski is proposing to silently switch to `MCRYPT_DEV_URANDOM` in Kohana 3.3. We are requiring PHP >= 5.3 for Kohana 3.3, so `MCRYPT_DEV_URANDOM` will be available. and I think that's a cool idea.

@sarciszewski would you mind a PR towards `3.3/develop` branch? Thanks!

enov:
@rjd22 I created issue #543 so that we do not forget to add tests for this class.

sarciszewski:
@enov Sure. If I have time today, I'll also throw some unit tests together :smile:

PR #541, opened 2014-09-05
Update Text.php

zogame:
$agent has been already defined


enov:
Thanks @Zogame

enov:
Created issue #544 for tests.

PR #542, opened 2014-09-06
Fix modules test to work when composer is fully implemented

lenton:
This needs to be merged to get the travis build of https://github.com/kohana/kohana/pull/41 to pass.

### DIFF
--- a/tests/kohana/CoreTest.php
+++ b/tests/kohana/CoreTest.php
@@ -274,7 +274,7 @@ class Kohana_CoreTest extends Unittest_TestCase
 	{
 		return array(
 			array(array(), array()),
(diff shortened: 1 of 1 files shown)


acoulton:
@lenton changing it like this will break the test for any project/module that is running the tests on an old directory structure.

Could you instead change it to return an array with any known directory from this module? Eg (**DIR** would possibly be enough). The only thing that matters is it's a directory we know exists and it doesn't (and won't ever) contain an init.php

lenton:
@acoulton That's a better idea.

Ok, I've now mocked the module path by using **DIR**, it's ready to be reviewed and merged.

lenton:
@enov would you mind merging this? @acoulton seems to be away

PR #545, opened 2014-09-09
Update Encrypt.php

sarciszewski:
Parallel with #540; will silently replace the insecure `MCRYPT_RAND` with `MCRYPT_DEV_URANDOM`

### DIFF
--- a/classes/Kohana/Encrypt.php
+++ b/classes/Kohana/Encrypt.php
@@ -134,11 +134,6 @@ class Kohana_Encrypt {
 		{
 			Encrypt::$_rand = MCRYPT_DEV_URANDOM;
 		}
-		// Do not allow insecure random number generator
-		if ((Encrypt::$_rand !== MCRYPT_DEV_URANDOM) AND (Encrypt::$_rand !== MCRYPT_DEV_RANDOM))
-		{
-			throw new Kohana_Exception("Insecure random number generator chosen");
-		}
 
 		// Create a random initialization vector of the proper size for the current cipher
 		$iv = mcrypt_create_iv($this->_iv_size, Encrypt::$_rand);
--- a/classes/Kohana/Encrypt.php
+++ b/classes/Kohana/Encrypt.php
@@ -130,38 +130,14 @@ class Kohana_Encrypt {
 	public function encode($data)
 	{
 		// Set the rand type if it has not already been set
-		if (Encrypt::$_rand === NULL)
+		if ((Encrypt::$_rand === NULL) OR (Encrypt::$_rand === MCRYPT_RAND))
 		{
-			if (Kohana::$is_windows)
-			{
-				// Windows only supports the system random number generator
-				Encrypt::$_rand = MCRYPT_RAND;
-			}
-			else
-			{
-				if (defined('MCRYPT_DEV_URANDOM'))
-				{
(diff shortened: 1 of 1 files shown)


enov:
Hmmm... Even though, practically, the throwing of the exception is never reachable, it changes the method's API, which we are not allowed to do in a minor release.

Instead, could we just replace with `MCRYPT_DEV_URANDOM`? I think this also takes care of the `NULL` and `MCRYPT_RAND` of the first `if`:

<pre>
// Silently revert to MCRYPT_DEV_URANDOM when an insecure random number generator is chosen
if ((Encrypt::$_rand !== MCRYPT_DEV_URANDOM) AND (Encrypt::$_rand !== MCRYPT_DEV_RANDOM))
{
    Encrypt::$_rand = MCRYPT_DEV_URANDOM;
}
</pre>

Thank you for your time :)

sarciszewski:
How does that look? :)

shadowhand:
Can't we just drop the `Encrypt::$_rand` parameter entirely and always use `MCRYPT_DEV_URANDOM`?

sarciszewski:
Sure. Performance- and security-wise, `MCRYPT_DEV_URANDOM` is optimal. 
http://www.sockpuppet.org/blog/2014/02/25/safely-generate-random-numbers

Anyone object to this change?

shadowhand:
The requirement about API changes in minor versions is that the **public** API cannot change. Since the only change here is internal, there's nothing wrong with doing this within 3.3 branch.

sarciszewski:
I wonder if @tqbf would approve? :smile:

enov:
@shadowhand 

> The requirement about API changes in minor versions is that the public API cannot change

Unfortunately, due to the CFS nature of Kohana, even the internal API should be left intact. I remember this is what I was always advised by @Zeelot. What happens if, for example, an extending class made use of an internal method, any API change to that method in a minor release, even to the set of the exceptions that it throws, will break the application.

@sarciszewski

I see you have left only `MCRYPT_DEV_URANDOM`. Personally I don't mind, but a dev can not choose  `MCRYPT_DEV_RANDOM`. I'll merge if all are OK with this change. Thank you for your hard work.

enov:
@shadowhand 

Speaking of a minor release, I am willing to invest some time for Kohana 3.3.3 to be released. We have still in 3.3, bugs that were fixed in 3.2.3. I need our old issue tracker to review issues, but I can not connect to http://dev.kohanaframework.org . Could you kindly check what's happening?

Thanks!

enov:
@sarciszewski I originally did not mind the d4d4bba change, but honestly I am not comfortable with it anymore, because `Encrypt::$_rand` looks silly, now that it is never used.

Could you please either revert to your earlier commit, or use my suggestion above?

Also you were saying that you'll throw some unit tests if you had time? :smile:

sarciszewski:
Yikes, how did I overlook this? (I need to make a list of outstanding github issues that need addressing for my own uses.)

Will try to tackle this weekend.

enov:
@sarciszewski I did the unit tests here in #560 if you have time to review it.

enov:
This has been fixed in #565. Thanks @sarciszewski for bringing this forward and contributing to the security level of the framework.

Issue #547, opened 2014-09-17
Issues reported for version 3.2.3 in Redmine that were not fixed in version 3.3.3


enov:
This is a placeholder issue for all Issues reported for version 3.2.3 in Redmine that were not fixed in version 3.3.3.

A new issue will be created and linked here, for each item below, with its title and description from Redmine.

The letter **R** is prefixed to the Redmine's issue number to avoid confusion between the issues of Github.

If you want to take an issue, drop me a line here, so that I assign it to you.

Hopefully we get through this and have a good 3.3.3 release :)

<hr>
- [x] [R4791](http://dev.kohanaframework.org/issues/4791)   Request_Client_Curl adds extra content-type header to all requests
- [x] [R4778](http://dev.kohanaframework.org/issues/4778)   PHPDoc comments - license location wrong
- [x] [R4678](http://dev.kohanaframework.org/issues/4678)   send_file() error on mime type
- [x] [R4651](http://dev.kohanaframework.org/issues/4651)   Request::uri() throws exception instead of returning a valid uri (docs obsolete)
- [x] [R4627](http://dev.kohanaframework.org/issues/4627)   shutdown_handler always uses 'Kohana_Exception::handler' instead of what's set in exception_handler()
- [x] [R4625](http://dev.kohanaframework.org/issues/4625)   Miscellaneous spelling errors
- [x] [R4622](http://dev.kohanaframework.org/issues/4622)   StdOut and StdErr Log Writers invalid log line format causes them not to output levels
- [x] [R4613](http://dev.kohanaframework.org/issues/4613)   Request class converts . to _ in url query string
- [x] [R4607](http://dev.kohanaframework.org/issues/4607)   Innacurate secure connection detection behind a load balancer
- [x] [R4592](http://dev.kohanaframework.org/issues/4592)   Response::send_file broken when no mime type specified
- [x] [R4587](http://dev.kohanaframework.org/issues/4587)   Request::redirect calls URL::site using the wrong parameters
- [x] [R4568](http://dev.kohanaframework.org/issues/4568)   DebugTest::test_dump fails because file has crlf line endings
- [x] [R4522](http://dev.kohanaframework.org/issues/4522)   Response does not respect Request's protocol
- [x] [R4482](http://dev.kohanaframework.org/issues/4482)   Array to string conversion in Arr::merge
- [x] [R4287](http://dev.kohanaframework.org/issues/4287)   Validation memory leak
[...]

PR #459, opened 2014-02-14
Fixes #3931: allows Kohana to shutdown properly and exit(1) in case of an exception

enov:

PR #553, opened 2014-09-19
R3931 - Register exit(1) as a shutdown function, to allow Kohana shutdown properly

enov:
Originally submitted by @zombor in [R3931](http://dev.kohanaframework.org/issues/3931) that `Kohana_Exception::handler` should `exit(1)`.

Also related to [R4794](http://dev.kohanaframework.org/issues/4794) and to its duplicate [R4275](http://dev.kohanaframework.org/issues/4275) in version 3.2, however these are not affected in version 3.3 because of the availability and use of `Kohana_Exception::_handler` (with underscore).

Through a shutdown function, we will postpone the `exit(1)` after Kohana shutdowns properly, i.e. after `Kohana::shutdown_handler`.

Ref #547. Please review and merge. Thanks!

### DIFF
--- a/classes/Kohana/Kohana/Exception.php
+++ b/classes/Kohana/Kohana/Exception.php
@@ -88,13 +88,8 @@ class Kohana_Kohana_Exception extends Exception {
 		// Send the response to the browser
 		echo $response->send_headers()->body();
 
-		// Shutdown closure
-		$shutdown_exit1 = function() {
-			exit(1);
-		};
-
-		// Register the closure
-		register_shutdown_function($shutdown_exit1);
+		// Register shutdown function exit(1)
+		register_shutdown_function('exit', 1);
 
 		// Return
 		return TRUE;
--- a/classes/Kohana/Kohana/Exception.php
+++ b/classes/Kohana/Kohana/Exception.php
@@ -88,7 +88,13 @@ class Kohana_Kohana_Exception extends Exception {
 		// Send the response to the browser
 		echo $response->send_headers()->body();
 
-		exit(1);
+		// Shutdown closure
+		$shutdown_exit1 = function() {
+			exit(1);
+		};
+
+		// Register the closure
+		register_shutdown_function($shutdown_exit1);
 	}
 
 	/**
--- a/classes/Kohana/Kohana/Exception.php
+++ b/classes/Kohana/Kohana/Exception.php
@@ -95,6 +95,9 @@ class Kohana_Kohana_Exception extends Exception {
 
 		// Register the closure
 		register_shutdown_function($shutdown_exit1);
+
+		// Return
+		return TRUE;
 	}
 
 	/**


kemo:
How about `register_shutdown_function('exit', 1)`?

enov:
Haha! @kemo that's good and straightforward, never thought of that. Where were you when I did an ugly hack in #460 for 3.2?

kemo:
I don't pay too much attention to 3.2 development anymore :blush:

acoulton:
@enov in the discussion on [R3931](http://dev.kohanaframework.org/issues/3931) you suggested a flag so that an outer exception handler could clear it if it wanted to. I don't know if there's much of a use case for that, but I guess there might be...

What happens if you want to set a specific exit code, for example, and so you call `exit(199)` - would that carry through or does the shutdown function override it and set it back to 1?

kemo:
@acoulton calling `exit` inside of a registered shutdown function should prevent the remaining registered shutdown handlers to be called, I suppose the same goes if you set `exit` as a shutdown handler but that should be tested

enov:
@acoulton I suggested a flag because we already have a shutdown handler, so instead of adding a new one (at that time,I did not even know we can add one), we test for that flag at the end of Kohana::shutdown_handler. The idea was implemented, but rejected, here: #459.

enov:
It would be a good idea for version 3.4 to have a `$exit_code` property in Kohana_Exception. It could be overridden by any exception class and set to `199` if you want :)

enov:
To add to the discussion, exit codes are usually used by the CLI but now we have Minion for that.

acoulton:
@kemo I was more meaning if you end up with:

``` php
// Error handler
register_shutdown_function('exit', 1);
// Your own code that runs later
register_shutdown_function('exit', 12);
exit(199);
```

HHVM gives an exit code of 1 - presume PHP will too. So with this change there's no way to return anything other than 1 once Kohana's handled the exception.

@enov an `$exit_code` on Kohana_Exception is a good idea for future. I'm not sure how best we could fix this for 3.3 though - there really needs to be a way to have the shutdown function exit 1 only if no other exit code has been set, but I don't know if that information is available anywhere?

enov:
@acoulton from reading your example, PHP should exit with 199 error code, shutdown functions will not be called. I did not test though.

kemo:
@acoulton in your example `exit(199)` will call the shutdown sequence in the order handlers were added in - thus the first shutdown handler `exit(1)` will be the final one as it will prevent invoking the remaining shutdown handlers (hence my previous comment)

enov:
I guess @kemo is right. From the PHP manual (http://php.net/manual/en/function.register-shutdown-function.php)

> Registers a callback to be executed after script execution finishes or exit() is called.

enov:
Please do not merge this, as it seems this is redundant. It seems that Kohana will be able to shutdown properly anyways.

enov:
I will close this now.

@acoulton , for 3.3, maybe you want to extend `Kohana_Exception::handler` to exit with the code number you want? Also, `exit(199)` anywhere from your code might work too.

@kemo thanks for clarifying up :+1: . This seems redundant for 3.3.

acoulton:
Done a bit more testing.
- PHP5.3, `exit(199)` triggers the shutdown callbacks in order. When it hits `exit(1)` it exits - no further callbacks run. The process exits with code 199.
- HHVM the execution flow is the same, but the process exits with code 1. 

So it looks like PHP5 always exits with the **first** value passed to exit, while HHVM returns the last one.

@enov I don't think this is redundant - the regression (accepted and fixed as a 3.2 bug, only still present because that fix has not been merged up) that we shouldn't be exiting from inside the handler still exists.

But I don't think we can just defer it with `register_shutdown_function('exit', 1);` because of the inconsistent final exit code behaviour. Also extending Kohana_Exception::handler won't work as far as I can see because exit will already be called in the parent method - unless you mean duplicating and never calling the inherited method?

Why can't we use the solution already accepted and released for 3.2?

enov:
@acoulton ,

The regression, `exit`-ing inside the handler, was shadowed away with the presence of the `_handler` method.

Yes, Custom error codes can be emitted by either duplicating the `handler` and never calling the inherited method, or we need introduce another class property.

It is better not to use the solutions from 3.2 because it does the same thing, but in a hacky manner. It registers the standard Kohana::shutdown_handler again. The standard Kohana::shutdown_handler is tweaked in a way that it acts differently the second time it is called.

I am open to suggestions. The branch is still there, and I can reopen.

acoulton:
@enov Sorry I'd not quite followed all the related tickets. 

So the exit inside `handler` remains as a code smell for now, but the specific regression (that userguide was broken) is resolved because it now calls `_handler` which doesn't exit.

In that case I agree that this does not need to be fixed for 3.3, and we don't need to add the hacky 3.2 fix just to move the code smell somewhere else.

PR #555, opened 2014-09-22
Xdebug returns `dynamic` and `static` instead of using `->` and `::`

enov:
When debugging a Kohana application through error view, things starts to be confusing as the names of the classes are concatenated to the method names with words `dynamic` and `static`. See the images:

Before:
before

After:
after

So I borrowed (back) some code from here:
https://github.com/symfony/symfony/pull/5863/files#diff-230e5606e6f592c8b8973fc758d112c8R185

As it seems to me that that code itself was borrowing some code from here:
https://github.com/kohana/core/pull/194/files#diff-6e70141994f9975e43722712df1e9611R201

Please review and merge.

Cheers!

### DIFF
--- a/classes/Kohana/Kohana/Exception.php
+++ b/classes/Kohana/Kohana/Exception.php
@@ -217,6 +217,16 @@ class Kohana_Kohana_Exception extends Exception {
 							$frame['type'] = '??';
 						}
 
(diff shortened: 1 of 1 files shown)


kemo:
:+1:

rafi:
Cool! @enov notice that Kohana's code is mostly 80 characters long, especially comments which are broken down to fit a 80 width window.

Issue #544, opened 2014-09-09
Text::user_agent needs test


enov:
The `Text::user_agent` method needs test.

enov:
I feel that the new #556 tests are satisfactory. If someone thinks otherwise, please reopen.

PR #556, opened 2014-09-24
Feature/tests user agent

Ikke:
These are some initial tests. It's difficult to make these tests exhaustive, so I tried to cover the main ones, but it could be that I'm still missing some.

If anyone has comments on the way this is tested, please do tell.

Addresses #544


enov:
This is excellent, @Ikke , thank you.

Issue #543, opened 2014-09-09
The Encrypt class needs test


enov:
The Encrypt class does not have unit tests.

PR #560, opened 2014-10-07
Unit test for the Encrypt class

enov:
The test data are provided by:
Cryptographic Algorithm Validation Program
National Institute of Standards and Technology

http://csrc.nist.gov/groups/STM/cavp/index.html

Used AES Multiblock Message Test (MMT) Sample Vectors
from http://csrc.nist.gov/groups/STM/cavp/documents/aes/aesmmt.zip
(the actual file used in that zipped file is CBCMMT128.rsp)

Also adds a class Kohana_EncryptTest_TestableEncrypt that wraps the
Encrypt class to mock out the actual mcrypt_create_iv calls.

However, it also changes Kohana_Encrypt internal API by adding
a protected function _create_iv to proxy mcrypt_create_iv
and to allow mocking and testing against KAT vectors.

Addresses #543.

Ready for merge. Kindly review thoroughly.

Thanks!

### DIFF
--- a/tests/kohana/EncryptTest.php
+++ b/tests/kohana/EncryptTest.php
@@ -388,6 +388,219 @@ class Kohana_EncryptTest extends Unittest_TestCase
 		$this->assertSame($expected, $actual);
 	}
 
+	/**
+	 * Provider for test_encode_decode, test_consecutive_encode_different_results
+	 *
+	 * @return array of $key, $mode, $cipher, $txt_plain
+	 */
+	public function provider_encode_decode()
+	{
+		return array(
+			array(
+				// key
+				"Some super secret key",
+				// mode
+				MCRYPT_MODE_NOFB,
+				// cypher
+				MCRYPT_RIJNDAEL_128,
+				// txt_plain
+				"The quick brown fox jumps over the lazy dog",
+			),
+			array(
+				// key
+				"De finibus bonorum et malorum",
+				// mode
+				MCRYPT_MODE_NOFB,
+				// cypher
+				MCRYPT_RIJNDAEL_128,
+				// txt_plain
+				"Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.",
+			),
+		);
+	}
+
+	/**
+	 * @param type $key Encryption Key
+	 * @param type $mode Encryption Mode
+	 * @param type $cipher Encryption Cipher
+	 * @param type $txt_plain Plain text to encode and then decode back
+	 *
+	 * @dataProvider provider_encode_decode
+	 * @covers Encrypt::encode
+	 * @covers Encrypt::decode
+	 */
+	public function test_encode_decode($key, $mode, $cipher, $txt_plain)
+	{
+		// initialize, encode
+		$e = new Encrypt($key, $mode, $cipher);
+		$txt_encoded = $e->encode($txt_plain);
+
+		// prepare data
+		$expected = $txt_plain;
+		$actual = $e->decode($txt_encoded);
+
+		// assert
+		$this->assertSame($expected, $actual);
+	}
+
+	/**
+	 * Provider for test_decode_invalid_data
+	 *
+	 * @return array of $key, $mode, $cipher, $txt_invalid_encoded
+	 */
+	public function provider_decode_invalid_data()
+	{
+		return array(
+			array(
+				// key
+				"Some super secret key",
+				// mode
+				MCRYPT_MODE_NOFB,
+				// cypher
+				MCRYPT_RIJNDAEL_128,
+				// txt_invalid_encoded
+				".:This data is not a valid base 64 string:.",
+			),
+			array(
+				// key
+				"Some super secret key",
+				// mode
+				MCRYPT_MODE_NOFB,
+				// cypher
+				MCRYPT_RIJNDAEL_128,
+				// txt_invalid_encoded
+				base64_encode("too short"),
+			),
+		);
+	}
+
+	/**
+	 * Tests for decode when the string is not valid base64,
+	 * or is too short to contain a valid IV
+	 *
+	 * @param type $key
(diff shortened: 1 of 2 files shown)


acoulton:
@enov this looks good and thorough to me - there might be an argument that it tests too much of the internal mcrypt behaviour but I think given the security implications it's probably better to do that than to mock out the mcrypt calls as well.

A few things I'd consider would really make this comprehensive
- Explicitly verifying that value > encode > decode === value - I know that the providers essentially do that by using known values, but IMHO it's a clearer demonstration of the behaviour of the class. Maybe you could combine the encode and decode tests?
- A test without the iv stubbed to verify that consecutive calls to encode the same value produce different results
- Tests for instance - that it returns singleton, that an instance created through the instance method has the expected configuration
- Tests for decode when the string is not valid base64, or is too short to contain a valid iv

enov:
Thanks for reviewing this @acoulton . I will rework on this soon to include the tests you described.

acoulton:
@enov no problem - they're a bit nerdy, you've done great work finding and implementing the NIST tests :)

enov:
I don't know if we're affected by this?
https://bugs.php.net/bug.php?id=68238

enov:
http://php.net/manual/en/function.mcrypt-encrypt.php#refsect1-function.mcrypt-encrypt-changelog

## Changelog

**5.6.0**    Invalid key and iv sizes are no longer accepted. mcrypt_encrypt() will now throw a warning and return FALSE if the inputs are invalid. Previously keys and IVs were padded with '\0' bytes to the next valid size.

enov:
Reading @nikic comments, in https://bugs.php.net/bug.php?id=68238

> If that is the case, you should either modify it to explicitly emulate the old \0 byte padding behavior or - which I would strongly recommend - fix your encryption code to use correctly sized keys.

I suggest that we add the old \0 byte padding behavior ourselves in 3.3.

In 3.4, we could test for accepted key sizes (mcrypt_enc_get_supported_key_sizes?) and throw an exception at construction when not acceptable. Also, we probably need to provide/document some way to migrate old encrypted data to the new encryption behavior.

enov:
@acoulton this is now ready for merge, with all your nerdy suggestions taken care of :)

Actually one of your suggestion revealed that PHP 5.6 does not pad null bytes to encryption keys anymore.

Kindly review.

tests/kohana/EncryptTest.php
```diff
@@ -0,0 +1,732 @@
...
+	 * @param type $mode Encryption Mode
+	 * @param type $cipher Encryption Cipher
+	 * @param type $txt_plain Plain text to encode and then decode back
+	 *
+	 * @dataProvider provider_key_normalization
+	 * @covers Encrypt::_normalize_key
+	 */
+	public function test_key_normalization($key, $iv, $mode, $cipher, $txt_plain)
+	{
+        if (version_compare(PHP_VERSION, '5.6.0', '>='))
```
acoulton: @enov couple of whitespace issues here

tests/kohana/EncryptTest.php
```diff
@@ -0,0 +1,732 @@
...
+
+		// test if instances are well configured
+		// prepare expected variables
+		$expected_cipher = $config[$config_group]['cipher'];
+		$expected_mode = $config[$config_group]['mode'];
+		$expected_key_size = mcrypt_get_key_size($expected_cipher, $expected_mode);
+		$expected_key = substr($config[$config_group]['key'], 0, $expected_key_size);
+
+		// assert
+		$this->assertSame($expected_key, $e->_key);
```
acoulton: @enov shouldn't _key, _mode etc be protected properties? It looks like they're not actually declared on the class.

acoulton:
@enov this looks great, nice work! Should give us a lot more confidence. Just flagged a couple of minor issues.

tests/kohana/EncryptTest.php
enov: those are fixed now in f3eea76

tests/kohana/EncryptTest.php
enov: Yes, they are not declared on the class, they should have been originally declared as protected.

If I do so now, should I create the getters in `Kohana_EncryptTest_IvStubbed`? Or directly in the original `Kohana_Encrypt` class?

enov: I guess it should go directly to the original `Kohana_Encrypt` class. But getters should be `public` and that would be too much to assume we're not breaking any existing code... Maybe we fix those in version 3.4?

acoulton: hm, tricky. To test the values provided through `instance` you'll need to be checking on `Kohana_Encrypt` unless you can find a way to make `instance` create instances of the subclass.

But it doesn't feel right to expose getters for them - especially the key. They're deliberately internal to the class once it's constructed.

Sometimes for specific cases like this I assert the value of the protected property using reflection. It couples the test to the implementation of the class, and feels a bit dirty. But it limits the dirtiness to the test rather than exposing new public API to all consumers.

So eg:

``` php
$this->assertSameProtectedProperty($expected_key, $e, '_key');

protected function assertSameProtectedProperty($expect, $object, $name)
{
  $refl = new ReflectionClass($object);
  $property = $refl->getProperty($name);
  $property->setAccessible(TRUE);
  $this->assertSame($expect, $property->getValue($object));
}
```

enov: @acoulton could you check now, I implemented the `assertSameProtectedProperty` and fixed the tests to use that.

enov: Should I also test the `_iv_size`?

acoulton: @enov great. I don't think you need to test `_iv_size` - it's calculated from mcrypt inside the class based on the cipher and method arguments. It's implementation detail that's covered by the fact your tests for encrypt/decrypt would fail if it broke.

You only need to spy into the class for the args provided to the constructor, everything else is covered by your behaviour tests already.

PR #563, opened 2014-11-04
PSR-3 compliance

enov:
This is a work-in-progress for PSR-3 compliance.


enov:
Hmm, this is not based on 3.4/develop. It seems to me that I have created a branch starting from the wrong revision.

enov:
Or maybe I did not pull before starting.

enov:
@acoulton, could you kindly review this? Also, kindly guide me what types of tests I need to write to make this complete?

cyrrill:
This is freking great, I'm very glad to see Kohana move closer to full PSR :) :+1:

acoulton:
@enov this is looking great. A couple of thoughts:
- Do we need a Logger and Kohana_Logger interface? What benefit is there to someone being able to transparently extend the interface?
- Might it be worth introducing a class (perhaps with ArrayAccess for compatibility) for the log message itself, rather than the simple array that Log::log pushes onto the _messages array? Would allow better typehinting for that property and also for writer methods etc.

In terms of testing, things I'd consider covering (I've not reviewed what if anything is covered by existing tests):

#### Kohana_Core::init
- Creates default log instance if not set
- Does not change log instance if already assigned to a dummy \Psr\Log\LoggerInterface

#### Kohana_Exception::log
- Behaviour when Kohana::$log is not assigned (fails silently?)
- Logs the exception
- Writes logs if a Logger
- Does not fail (eg from attempting to call write) if the Kohana::$log is a non-kohana LoggerInterface

### Kohana_Log
- I'd cover the inherited AbstractLogger methods for completeness, eg with a self-shunt extension class that captures calls to log and verifies that the level-specific methods call it as expected.
- That add calls log with expected argument structures
- How ::log handles the various types of levels (string, int, undefined)
- That ::log casts message to string
- That ::log interpolates context to message
- The various message properties are set in the messages array as expected
- That ::log writes if write_on_add is set - again could do this with a self-shunt

### Kohana_Log_Syslog
- That it calls syslog with expected arguments, eg using a self-shunt
- Level conversion
- Exception logging

And if Kohana_Log_Writer::format_message doesn't already have coverage then I guess also tests for that.

enov:
> Do we need a Logger and Kohana_Logger interface

No we don't, interfaces should not be transparently extended like implementations, IMHO.

> Might it be worth introducing a class for the log message itself

My initial intention was to be psr-3 compliant with the least change possible in the code base. I thought can further enhance the code in later versions. What do you say?

acoulton:
> > Do we need a Logger and Kohana_Logger interface
> 
> No we don't, interfaces should not be transparently extended like implementations, IMHO.

Agreed - but you have both in this PR. My preference would be to drop `Logger` and update the type checks throughout to `Kohana_Logger` since that keeps it clearer where that has come from (eg makes obvious it's not a PSR3 interface).

> > Might it be worth introducing a class for the log message itself
> 
> My initial intention was to be psr-3 compliant with the least change possible in the code base. I thought can further enhance the code in later versions. What do you say?

Fair point, yes this pull should be limited to what's strictly necessary for PSR3.

enov:
> My preference would be to drop Logger and update the type checks throughout to Kohana_Logger

Agreed. I will create an issue for this later, targeting all other interfaces that follow this pattern.

arteymix:
@enov do you think [PSR-7 http-message](https://github.com/php-fig/http-message) should be implemented as well?

enov:
@arteymix that's still under discussion. Plus, it would require us to be PSR-1/2 coding style, because unlike this PSR-3, it has long camelCased method names.

There is also a draft RFC about HTTP interface, that _might_ go into PHP 7:
https://wiki.php.net/rfc/http-interface

arteymix:
That's what I feared when I looked into the specification. PSR 1/2 is hardly thinkable at thins point.

It's a good thing though that we get more compliance with PSR, so I second you on that PR.

enov:
@acoulton, I am struggling with writing the first test:

> Kohana_Core::init
> - Creates default log instance if not set

The problem is that PHPUnit boostraps Kohana the standard way, which calls `Kohana::init()`.

So, since  `Kohana::init()` is already called and log is set, in order for me to have an unset log, I have to unset it within the test. Now, while that seems to be easy, calling `Kohana::init()` again is not, because the protected flag `Kohana::$_init` that checks if it is initialized. That would lead us to call `Kohana::deinit()` within the test, which sets that flag to `FALSE`, but would remove also the registration of the standard Kohana autoloader, which makes other tests to fatal.

It seems to me I am hunting some kind of chicken and egg problem. I found the following 2 sub-par solutions:

## Solution 1: Replay parts of bootstrap within the test

To be able to `deinit()` and then re`init()`, I need to replay, in between these calls, parts of bootstrap required for tests to continue. I think this is sub-par solution because changing the bootstrap later on, would require us to revisit these tests and fix them.

Kindly find here-under the code for this solution:

``` php
    /**
     * Tests Kohana::init() for default log file creation
     */
    public function test_init_creates_default_log_instance_if_not_set()
    {
        // de-initialize Kohana, should set Kohana::$log to NULL
        Kohana::deinit();

        // re-register autoloader as de-initialization removes that
        spl_autoload_register(array('Kohana', 'auto_load'));

        // test if Kohana::$log is NULL at this point
        $this->assertNull(Kohana::$log);

        // re-initialize Kohana
        Kohana::init();

        // test if a default log instance is created
        $this->assertInstanceOf('Log', Kohana::$log);

        /**
         * At this point we need to replay some part from bootstrap
         * so that other tests may continue normally
         */

        // Attach the file write to logging. Multiple writers are supported.
        Kohana::$log->attach(new Log_File(APPPATH . 'logs'));

        //Attach a file reader to config. Multiple readers are supported.
        Kohana::$config->attach(new Config_File);
    }
```

## Solution 2: a TestableCore that extends Core with a public set_init_flag

This solution relies on the fact that we can run `init()` again if the flag `$_init` is set to `FALSE`, which help us avoid to `deinit()` Kohana. I do think this solution is also sub-par, because it re-registers the exception handler, error handler and shutdown handler with PHP runtime again.

Please see the code for this solution here:

``` php
    /**
     * Tests Kohana::init() for default log file creation
     */
    public function test_init_creates_default_log_instance_if_not_set()
    {
        // set $log to NULL, as it inherited the instance from the parent class
        Kohana_CoreTest_TestableCore::$log = NULL;

        // test if it is NULL (do we need this?)
        $this->assertNull(Kohana_CoreTest_TestableCore::$log);

        // set the flag Kohana::$_init to FALSE to be able to run init() again
        Kohana_CoreTest_TestableCore::set_init_flag(FALSE);

        // initialize Kohana again
        Kohana_CoreTest_TestableCore::init();

        // assert
        $this->assertInstanceOf('Log', Kohana_CoreTest_TestableCore::$log);
    }
```

Code for `Kohana_CoreTest_TestableCore`:

``` php
class Kohana_CoreTest_TestableCore extends Kohana_Core {

    /**
     * Sets the init flag to force initialization/deinitialization
     * @param bool $flag
     */
    public static function set_init_flag($flag)
    {
        static::$_init = $flag;
    }

}
```

Thanks.

acoulton:
@enov yes, that's all fairly horrid isn't it :)

Not sure what to suggest, as you say the main issue is that `init` does so much and replaying all of that in the unit test is bound to cause problems.

I suppose you could split all the separate actions in init into protected methods, then stub a class that makes all of them no-op - for example:

``` php
class Kohana_Core {
  public static function init() {
    if (Kohana::$_init)
    {
      // Do not allow execution twice
      return;
    }

    // Kohana is now initialized
    Kohana::$_init = TRUE;
    static::init_profile();
    static::init_errors();
    static::init_shutdown();
    static::init_windows();
    static::init_cache();
    // etc
    static::init_log();
    static::init_config();
  }
}

class LogInitKohana_CoreStub extends Kohana_Core {
  public static function init()
  {
    self::$_init = FALSE;
  }

  public static function init_profile() {}
  public static function init_errors() {}
  public static function init_cache() {}
  // etc for everything except init_log
}
```

That's not very nice either because of the need to be sure every `init_` method was extended with an empty stub over time.

Another option would be to break the init method up into a stack of initialisers (classes or callables of some kind) - then you could just test the LogInitialiser. That would be cleaner and give more scope for extension in future and by end-users, but is obviously a chunky refactoring and may have performance, BC and other considerations.

Maybe for now the only sensible thing in the current architecture is to assume that PHPUnit didn't itself assign anything to `Kohana::$log` and so if it has a value it must have been assigned by `init` - in other words just test the result. It might be better than trying to sort out the whole init method as part of this feature.

enov:
> assume that PHPUnit didn't itself assign anything to Kohana::$log and so if it has a value it must have been assigned by init - in other words just test the result.

If we can get away with that for the first test, we can not for the second test you suggested :smile:

> Kohana_Core::init
> -   Does not change log instance if already assigned to a dummy \Psr\Log\LoggerInterface

acoulton:
@enov indeed. I think maybe we have to accept that's almost impossible to test without introducing a lot of changes that aren't directly relevant to this PR. The only other thing I thought of was shelling out to a script in a separate PHP process but that would be pretty unpleasant too. Maybe for now these are things that just have to be verified by eye/knowing they work in end-user apps.

enov:
I added tests comparing messages resulting from the generic `log()` method, as well as the 8 specialized _level_ methods.

**Tests are currently failing**. This is because `Kohana_Log` adds _trace_ information to the message. And since there is an additional method call when one of the _level_ methods are used, we are having different _trace_ entries in those messages when we compare them to the message generated by the generic `log()` method.

PSR-3 specifies that:

> A ninth method, `log`, accepts a log level as first argument. Calling this method with one of the log level constants MUST have the same result as calling the level-specific method.

I think we either need to normalize the _trace_ entries, by carefully re-engineering the result of the call to [debug_backtrace](https://github.com/kohana/core/blob/3.4/feature/psr-3-compliance/classes/Kohana/Log.php#L274), or to forget about the _trace_ entry, as it is ignored by the abstract `Log_Writer`.

For the record, the [_file_, _line_, _class_, and _function_](https://github.com/kohana/core/blob/3.4/feature/psr-3-compliance/classes/Kohana/Log.php#L287-290) entries are also giving different results. Sometimes they are `NULL`. I am not sure why. I also did some tests without [array_slice](https://github.com/kohana/core/blob/3.4/feature/psr-3-compliance/classes/Kohana/Log.php#L274), and results seemed to be slightly better (no `NULL`s, but again different results). I am even unsure why that `array_slice` is there, to begin with!

Any insight and guidance is greatly appreciated.

enov:
Alright, played with fire and removed _trace_, _class_, and _function_ keys from log message.

Please review.

cc @acoulton, @lenton, @rjd22, @shadowhand

enov:
Do **not** merge yet.

enov:
@acoulton could you please review this? Specifically this commit: https://github.com/kohana/core/commit/32247e58cd3e43ea7e2eb6dbb56a5b8712639583

enov:
@acoulton I know you might be busy, but I just need your confirmation in case we can get rid of _trace_, _class_, and _function_ from log message (as we are not using them in the default log writer).

Those are still available in case of an exception. I'll write more tests to prove the case.

acoulton:
@enov so sorry, I've been flat out. I've just had a quick look over this and it's looking good.

I think it's fine to drop trace, class and function from ordinary log messages - as you say they're not written by any of our own writers and anyway that feels like something where the end-user should pass them as part of the context if they want them. Apart from unexpected exceptions (which are handled separately) usually you know the trace of application-specific logs.

enov:
Thanks for the review @acoulton .

rjd22:
@enov I also reviewed it and it looks good. I'm just wondering if we should still do our own logging. Wouldn't it be better to move to monolog? What do you think @acoulton 

Anyway all in good time. For now :+1:

enov:
Thanks for the review @rjd22.

rjd22:
@enov any updates on this?

enov:
@rjd22 Need to find time to write the remaining tests, and write some docs about how to plug-in a PSR-3 compatible logger, instead of using the stock Kohana logger.

rjd22:
@enov Another poke :P. It would be really nice to have this in Kohana :)

enov:
I will try to concentrate my time on this :)

enov:
I believe I have completed the writing of the requested tests, as well as implementing the official `\Psr\Log\Test\LoggerInterfaceTest`. Some tests seems to be overlapping. Please let me know if I am missing something.

Regarding the failure on PHPUnit v4.0.0, that's because tests in acd9d16ed5686207598c60823afe6d64c989bd4d are running via a separate PHPUnit process using both `@runInSeparateProcess` and `@preserveGlobalState disabled` annotations. When both are present, PHPUnit [prior to version 4.3.0](https://github.com/sebastianbergmann/phpunit/commit/7d8da4174961b3144e364990a499efe90c5b49f5) used to ignore the bootstrapping. That was fixed in https://github.com/sebastianbergmann/phpunit/pull/1364.

Kindly requesting your review and comments. Docs should be updated once everything is reviewed and accepted.

Thanks.

rjd22:
@enov on the PHPUnit problem. Would it be possible to just update the PHPUnit version to 4.3+ in `https://github.com/kohana/unittest/blob/3.3/master/composer.json` ?

classes/Kohana/Log.php
```diff
@@ -107,6 +122,29 @@ public function detach(Log_Writer $writer)
...
+		return $this->write_immediately;
+	}
+
+	/**
+	 * Set/unset immediate writing
+	 * 
+	 * @param boolean $write_immediately
+	 * @return Log
+	 */
+	public function set_immediate_write($write_immediately = TRUE)
```
rjd22: Does this need the default `TRUE` in the parameter? I would say that explicitly setting this would make this function more obvious and gives it less redundancy.

enov:
Yes, if the tests are finalized the way they are right now, updating the PHPUnit version to 4.3 might do the job.

classes/Kohana/Log.php
```diff
@@ -23,9 +23,24 @@ class Kohana_Log {
...
+		LOG_NOTICE  => \Psr\Log\LogLevel::NOTICE,
+		LOG_INFO    => \Psr\Log\LogLevel::INFO,
+		LOG_DEBUG   => \Psr\Log\LogLevel::DEBUG,
+	);
+
+	/**
 	 * @var  boolean  immediately write when logs are added
 	 */
-	public static $write_on_add = FALSE;
+	protected $write_immediately = FALSE;
```
rjd22: This can be a private right? Since it can be changed by the `set_immediate_write`

classes/Kohana/Log.php
enov: I think you're right about explicitly setting this.

classes/Kohana/Log.php
enov: `protected` gives access to classes that extends this class.

classes/Kohana/Log.php
```diff
@@ -225,4 +240,99 @@ public function write()
...
+	 * replaced using Text::interpolate
+	 *
+	 *     $log->log(\Psr\Log\LogLevel::INFO, 'You look good, today.');
+	 *
+	 *     $log->log(Log::ERROR, 'Could not locate user: {username}', array(
+	 *         'username' => $username,
+	 *     ));
+	 *
+	 * @uses   Text::interpolate Inserts the replacement values into the message
+	 * @param  mixed  $level    level of message
```
rjd22, line 273: This should only accept `int|string`

classes/Kohana/Log.php
```diff
@@ -225,4 +240,99 @@ public function write()
...
+		else if (is_string($level) AND in_array(strtolower($level), $this->_log_levels))
+		{
+			$level = strtolower($level);
+		}
+		else
+		{
+			throw new Psr\Log\InvalidArgumentException('Undefined level "' . $level . '"');
+		}
+
+		// cast $message to string in compliance with PSR-3
```
rjd22, line 283: No need for this comment. We can see what you're doing ;)

classes/Kohana/Log.php
```diff
@@ -23,9 +23,24 @@ class Kohana_Log {
 	const DEBUG     = LOG_DEBUG;    // 7
 
 	/**
+	 * Numeric log level to string lookup table.
+	 * @var array
+	 */
+	protected $_log_levels = array(
```
rjd22: This can be private? AFAIK

classes/Kohana/Log.php
```diff
@@ -225,4 +240,99 @@ public function write()
...
+	 *
+	 * @uses   Text::interpolate Inserts the replacement values into the message
+	 * @param  mixed  $level    level of message
+	 * @param  string $message  message body
+	 * @param  array  $context  values to replace in the message
+	 * @return Log
+	 */
+	public function log($level, $message, array $context = [])
+	{
+		// check if level is available
```
rjd22: // check if log level exists in the `$this->_log_levels` array.

classes/Kohana/Log.php
```diff
@@ -225,4 +240,99 @@ public function write()
...
+			if (isset($trace[0]['file']) AND  $parent_class_file === $trace[0]['file']) {
+				$trace = array_slice($trace, 1);
+			}
+
+			// set $context['exception'] to FALSE in order to not repeat
+			// the above if conditions again elsewhere
+			$context['exception'] = FALSE;
+		}
+
+		// Create a new message
```
rjd22: Redundant comment.

classes/Kohana/Log.php
```diff
@@ -225,4 +240,99 @@ public function write()
...
+		$message = array(
+			'time' => time(),
+			'level' => $level,
+			'body' => $message,
+			'file' => isset($trace[0]['file']) ? $trace[0]['file'] : NULL,
+			'line' => isset($trace[0]['line']) ? $trace[0]['line'] : NULL,
+		);
+
+		// add exception object to message, if available
```
rjd22: Redundant comment.

classes/Kohana/Log.php
enov, line 283: I usually comment excessively, but this is needed to show compliance.

classes/Kohana/Log/Syslog.php
```diff
@@ -42,16 +57,38 @@ public function write(array $messages)
...
+			$level = $this->_log_levels[$message['level']];
 
-			if (isset($message['additional']['exception']))
+			// write to syslog
+			$this->_syslog($level, $message['body']);
+
+			if (isset($message['exception']))
 			{
-				syslog(Log_Writer::$strace_level, $message['additional']['exception']->getTraceAsString());
+				// convert the level into int level
```
rjd22: // convert PSR log level into syslog log level

enov: :+1:

classes/Kohana/Log.php
enov: :+1:

classes/Kohana/Log.php
rjd22, line 283: Ok, then leave it there.

classes/Kohana/Log.php
enov: `protected` gives access to classes that extends this class.

classes/Kohana/Log.php
enov, line 283: :+1:

classes/Kohana/Log.php
enov, line 273: https://github.com/php-fig/log/blob/master/Psr/Log/LoggerInterface.php#L116

classes/Kohana/Log.php
rjd22: @enov the question is why would they need access. Normally I would agree in the sense of CFS but in this case the inner workings are already exposed.

rjd22:
@enov a few comments. You are commenting quite excessively. Please check all your comments in the review screen and see if the code is not telling the story itself.

Comments should be used as less as possible since they won't get maintained and will become obsolete when someone starts refactoring.

classes/Kohana/Log.php
rjd22, line 273: Wow, ok then leave it there ;)

classes/Kohana/Log.php
enov: Not sure why would they need, but not sure either why restrict? What do we loose?

enov:
Will review again my comments, and those that are redundant, I will remove them.

Thanks for taking time reviewing :+1:

classes/Kohana/Log.php
rjd22: Not much ;), If you think extending is important then leave them protected.

classes/Kohana/Log.php
rjd22: Ok leave it protected

classes/Kohana/Log.php
```diff
@@ -225,4 +240,99 @@ public function write()
...
+		if ($context)
+		{
+			// Insert the values into the message
+			$message = Text::interpolate($message, $context);
+		}
+
+		// Grab a copy of the trace and sanitize $context['exception']
+		if (
+		  isset($context['exception']) AND
+		  is_object($context['exception']) AND
```
Yahasana: `is_object($context['exception']) AND` no need

classes/Kohana/Response.php
```diff
@@ -556,10 +556,13 @@ public function send_file($filename, $download = NULL, array $options = NULL)
 				if (is_object(Kohana::$log))
 				{
 					// Add this exception to the log
-					Kohana::$log->add(Log::ERROR, $error);
+					Kohana::$log->error($error);
 
-					// Make sure the logs are written
-					Kohana::$log->write();
+					if (Kohana::$log instanceof Kohana_Logger)
```
Yahasana: why only write it for Kohana_logger?

``` php
Akunamutata_Log extends Psr\Log\LoggerInterface { ... }

Kohana::$log = new Akunamutata_Log;
```

classes/Kohana/Session.php
```diff
@@ -374,8 +374,14 @@ public function write()
 		catch (Exception $e)
 		{
 			// Log & ignore all errors when a write fails
-			Kohana::$log->add(Log::ERROR, Kohana_Exception::text($e))->write();
+			Kohana::$log->error(Kohana_Exception::text($e));
 
+			if (Kohana::$log instanceof Kohana_Logger)
```
Yahasana: same probs as above

classes/Kohana/Response.php
enov: @Yahasana kindly note that this `if` test is to insure that the `$log` can `write()`.

classes/Kohana/Log.php
```diff
@@ -225,4 +240,99 @@ public function write()
...
+			$trace = $context['exception']->getTrace();
+		}
+		else
+		{
+			$trace = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS);
+
+			// remove the call that comes from Psr\Log\AbstractLogger
+			// in order to have consistent file and line elements in log
+			$parent_class_file = (new ReflectionClass(get_parent_class()))->getFileName();
+			if (isset($trace[0]['file']) AND  $parent_class_file === $trace[0]['file']) {
```
Yahasana: ill format `{}`

classes/Kohana/Session.php
enov: same comment as above :stuck_out_tongue_winking_eye: To call `Kohana::$log->write();` we need to be sure that it is a `Kohana_Logger`

enov: You see @rjd22, there is not enough comments to explain the things. I think I need to add some explanation here.

classes/Kohana/Log.php
enov: :+1:

classes/Kohana/Log.php
```diff
@@ -225,4 +240,99 @@ public function write()
...
+	 * @uses   Text::interpolate Inserts the replacement values into the message
+	 * @param  mixed  $level    level of message
+	 * @param  string $message  message body
+	 * @param  array  $context  values to replace in the message
+	 * @return Log
+	 */
+	public function log($level, $message, array $context = [])
+	{
+		// check if level is available
+		if (is_int($level) AND array_key_exists($level, $this->_log_levels))
```
Yahasana: `$this->_log_levels[$level] === NULL` shouldn't happen, replace `array_key_exists($level, $this->_log_levels)` with `isset($this->_log_levels[$level])` for better performance

classes/Kohana/Kohana/Exception.php
```diff
@@ -145,10 +145,13 @@ public static function log(Exception $e, $level = Log::EMERGENCY)
 			$error = Kohana_Exception::text($e);
 
 			// Add this exception to the log
-			Kohana::$log->add($level, $error, NULL, array('exception' => $e));
+			Kohana::$log->log($level, $error, array('exception' => $e));
 
-			// Make sure the logs are written
-			Kohana::$log->write();
+			if (Kohana::$log instanceof Kohana_Logger)
```
Yahasana:
```php
if (Kohana::$log instanceof Psr\Log\LoggerInterface)
```

classes/Kohana/Response.php
Yahasana: Oops, you are right, there is no 'write' in `Psr\Log\LoggerInterface`. WTF

classes/Kohana/Log.php
enov: :+1: yes, `$this->_log_levels[$level] === NULL` shouldn't happen because of `is_int` test, and `isset` will be faster.

classes/Kohana/Log.php
```diff
@@ -107,6 +122,29 @@ public function detach(Log_Writer $writer)
...
+		return $this->write_immediately;
+	}
+
+	/**
+	 * Set/unset immediate writing
+	 * 
+	 * @param boolean $write_immediately
+	 * @return Log
+	 */
+	public function set_immediate_write()
```
rjd22: @enov Now you removed the variable completely ;). I think it should be 

``` php
public function set_immediate_write($write_immediately)
```

enov: Yeah, my mistake. Working hastily. :neckbeard:

rjd22:
@enov would it be a good idea to update the kohana/unittest to alway use PHPUnit 4.3 at a minimum? That way we can make your tests pass.

enov:
@rjd22 @Yahasana I took care of the things you asked. Thanks.

I think I still have a couple of things missing:
- Write test for Text::interpolate
- Support for Object `$level` with `__toString()` that translates to one of the PSR levels - but I am not sure if this is necessary, or if it is an overkill

Meanwhile, @acoulton if you could please review.

@rjd22 

> @enov would it be a good idea to update the kohana/unittest to alway use PHPUnit 4.0 at a minimum? That way we can make your tests pass.

You mean PHPUnit 4.3? Sure, I'll issue a PR, we can merge that if there is no objections to the tests done with `@runInSeparateProcess`.

rjd22:
@enov https://github.com/kohana/unittest/pull/59 This should fix the issue right?

enov:
@rjd22 yeah, perfect!

classes/Kohana/Log.php
acoulton: Every protected instead of private variable is a variable that can only in future be renamed/refactored in a breaking version release. For package code, we should probably lean to private by default unless there's a good reason to make it more visible. After all, if someone has a problem we can easily make it `private` => `protected` in a minor version release. We can't go the other way.

(especially since there's a getter/setter for this, so no real need for anyone extending to have direct access to the variable).

enov: Hmmm, I see. Did not think about refactoring in minor versions. That's a +1 for `private`.

acoulton:
@enov I think this is good, nice work! The tests for `Kohana::init` behaviour are a good solution and worth having, I think.

One thing (following to @Yahasana comments) why not have:

``` php
interface WriteableLogger {
  public function write();
}

class Kohana_Log implements Kohana_Logger, WriteableLogger {
//...
}
```

and then everything like this:

``` php
if ($log instanceof Kohana_Logger) {
  $log->write();
}
```

becomes

``` php
if ($log instanceof WriteableLogger) {
  $log->write();
}
```

No need for comments because the reason for the instanceof is very obvious, and much more flexible because someone can define their own PSR logger with a write method without having to also support our attach and detach methods.

enov:
@acoulton `WriteableLogger` is great idea! Probably `Kohana_Writeable_Logger` for the name?

acoulton:
@enov yeah `Kohana_Writeable_Logger` would work given we're not using namespaces yet.

enov:
@acoulton while I have almost implemented `Kohana_Writeable_Logger`, but I am reluctant to push. I see now that `Kohana_Logger` is more of a complete implementation of the observer pattern, and without a `write()`, it would not notify a subscribed writer.

While I see the benefit of having a separate `write()`, `Kohana_Logger` alone or  `Kohana_Log_Writer_Container` alone (I ended up calling it "container", as `attach` and `detach` makes it a mere container) - has little value.

Let me know what you think.

acoulton:
@enov well a logger that didn't implement delayed write - eg that wrote to some stream on every log call would only need `attach` and `detach` - but would not need or want to declare an explicit `write` method?

Likewise a logger that only supported deferred output to a single log destination might want `write` but not `attach` and `detach`.

acoulton:
@enov thinking about it, maybe the problem is the interface/method name. 

`Kohana_Flushable_Logger` and a `flush` method might make more sense - since probably all loggers can write, it's delayed write/flush that makes the default kohana logger different. But that would be an additional BC break.

If we want to keep the method called `write` then maybe the interface should be more like `Kohana_BufferableWrite_Logger` - I don't think a `flushable` interface makes sense with a `write` method, and the logger may or may actually be buffering depending on the immediate_write option.

enov:
Exactly! I was just going to say that... We just need some renaming/refactoring. I was myself thinking also to rename `write` to `flush`. I'll come up with some renamings tomorrow. Thanks!

rjd22:
Ran the tests. It seems that everything is okay now ;)

enov:
Added a test for `Text::interpolate`.

@acoulton I am thinking renaming `Kohana_Logger` interface to  `Kohana_Log_Collector` or  `Kohana_Log_Entry_Collector` , and `write()` to `flush()`. That helps emphasizing the role of the interface. It might go hand in hand if we add later a class `Log_Entry` for the log message and an ArrayAccess `Log_Entry_Collection`, similar to the idea you posted here:

https://github.com/kohana/core/pull/563#issuecomment-61961254

> Might it be worth introducing a class (perhaps with ArrayAccess for compatibility) for the log message itself, rather than the simple array that Log::log pushes onto the _messages array? Would allow better typehinting for that property and also for writer methods etc.

enov:
Renamed `Kohana_Logger` to `Kohana_Log_Buffer`, without extending `Psr\Log\LoggerInterface`.

@acoulton @rjd22 please review 0825761

rjd22:
Looks good :+1: from me

tests/kohana/LogTest.php
```diff
@@ -61,51 +61,360 @@ public function test_attach_attaches_log_writer_and_returns_this()
...
+		$exception = new Exception('dummy exception text');
+		return [
+			// data set #0
+			[
+				\Psr\Log\LogLevel::DEBUG, // level
+				'dummy message', // log text
+				[], // context
+				// expected raw log message array
+				[
+					'time' => time(), // to be reset within the test for HHVM
```
acoulton: Maybe don't set the time at all in the data provider and just add it in the test message, to make it most obvious?

classes/Kohana/Log/Writer.php
```diff
@@ -67,24 +62,202 @@
...
+
+	/**
+	 * Filter messages according to the levels accepted by this writer
+	 *
+	 * @param array $messages
+	 * @return array Filtered messages
+	 */
+	public function filter(array $messages)
+	{
+		if (empty($this->levels))
```
acoulton: This feels a bit backwards - I wouldn't expect $logger->set_levels([]) to result in it logging all messages.

classes/Kohana/Log/Writer.php
```diff
@@ -67,24 +62,202 @@
...
+	}
+
+	/**
+	 * Sets the log levels that this writer accepts to write
+	 *
+	 * @param array $levels
+	 * @throws InvalidArgumentException
+	 * @return Log_Writer
+	 */
+	public function set_levels(array $levels)
```
acoulton: I don't think it's totally clear what type of levels you're accepting here. PSR levels?

classes/Kohana/Log/Writer.php
```diff
@@ -67,24 +62,202 @@
...
+	{
+		$this->levels = array_map('Log::to_psr_level', $levels);
+
+		return $this;
+	}
+
+	/**
+	 * Sets the log levels' range that this writer accepts to write
+	 *
+	 * @param array $levels
```
acoulton: method docs here are wrong.

classes/Kohana/Log/Writer.php
```diff
@@ -67,24 +62,202 @@
...
+			// Write all of the messages
+			return $messages;
+		}
+
+		// Filtered messages
+		$filtered = array();
+
+		foreach ($messages as $message)
+		{
+			if (in_array($message['level'], $this->levels))
```
acoulton: in_array probably isn't the most efficient for multiple messages.

``` php
$include_level = array_fill_keys($this->levels, TRUE);
// Or you could actually store $this->levels as a map of level => include true/false
foreach ($messages as $message) {
  if (isset($include_level[$message['level']])) {

```

classes/Kohana/Log/Writer.php
```diff
@@ -11,38 +11,33 @@
...
-	 * @var  int  Level to use for stack traces
+	 * @var  string  default string format of the log entry
 	 */
-	public static $strace_level = LOG_DEBUG;
+	private $format = "time --- level: body in file:line";
+
+	/**
+	 * @var array log levels that this writer accepts to write
+	 */
+	private $levels;
```
acoulton: I'm not sure the meaning of `levels` is obvious - same for the various setters/getters. `write_levels`, `include_levels`?

classes/Kohana/Log/Writer.php
enov: I'll just test for `NULL`. But it's rather `$writer->set_levels()`.

classes/Kohana/Log/File.php
```diff
@@ -84,7 +84,9 @@ public function write(array $messages)
 			chmod($filename, 0666);
 		}
 
-		foreach ($messages as $message)
+		$filtered_messages = $this->filter($messages);
+
+		foreach ($filtered_messages as $message)
```
acoulton, line 89: I might have missed it, but I don't think the filtering within the log writers (any of them) is covered by tests?

classes/Kohana/Log/Writer.php
enov: Any type, integer levels like `LOG_DEBUG`, `LOG_INFO`, PSR/string levels...

I'll add examples to the docs.

classes/Kohana/Log/File.php
enov, line 89: I did not have finished yet the testing of the Log_Writer.

classes/Kohana/Log/Writer.php
rjd22: @acoulton readability > efficiency. This would make it harder to read imo.

classes/Kohana/Log/Writer.php
acoulton: I wouldn't expect `NULL` to make it log everything either :) I'd prefer if it was explicit - default $levels to contain all levels.

If it's just that you want to skip filtering for performance if it's not required (?) then if $levels was a map of level => true/false you could do `if ($this->levels === array_filter($this->levels))` to know whether any have been excluded.

classes/Kohana/Log/Writer.php
enov: Yep :+1:

classes/Kohana/Log/Writer.php
enov: Hmmm... I copied/pasted the stuff from the old `Log::write()` method. I'll look at this again.

acoulton: @rjd22 imo most readable would be `if ($this->write_level[$message['level']])`

classes/Kohana/Log/Writer.php
acoulton: Isn't that confusing then if `->get_levels` is always returning an array of PSR levels?

``` php
$levels = ['stuff'];
$writer->set_levels($levels);
if ($writer->get_levels() != $levels) print `eh?`
```

Obviously the logger itself has to accept anything for PSR compliance, but for the writer we should pick a single set of levels and stick to them.

acoulton:
@enov thanks for all your work on this, I like where it's going. Will you be able to write some documentation on how to upgrade from a 3.3-style logger/writer setup as well?

And sorry if I confused you by adding some comments as commit comments first, I forgot they show up differently / less usefully in the PR conversation so deleted and moved to PR comments.

classes/Kohana/Log/Writer.php
acoulton: @enov isn't it annoying when you copy/paste code and people who never cared about it before start picking holes in it :) Sorry about that.

classes/Kohana/Log/Writer.php
enov: Well, I see that it's a bit confusing. We can add just a note about it instead of sticking with a single set. Otherwise, it may be more confusing that the logger accepts anything, but the writer sticks with a single set?

enov:
Thanks @acoulton. Glad to know you like where it is going. I'll add some docs once I finish the tests for the writer. We can then skim through everything, review the docs, and happily see this merged :)

classes/Kohana/Log/Writer.php
acoulton: IMO having a getter return different value types to the setter is much more confusing/problematic than this interface (of our own) being stricter than the generic logger interface.

For example if you configured it with `[LOG_DEBUG, LOG_INFO]` etc you might assume elsewhere you could temporarily suppress a level by calling `get_levels`, removing (LOG_DEBUG), and passing back to `->set_levels` but that won't work.

I would make all these methods require PSR LogLevels and throw if they get anything else, so the developer encounters a one-time exception and figures it out rather than getting unexpected behaviour at runtime.

classes/Kohana/Log/Writer.php
enov: Haha! no problem @acoulton !

classes/Kohana/Log/Writer.php
enov: Please note that a bit below, we have `set_levels_range` that needs to accept integer levels for it to convey more its meaning.

acoulton: I don't think it does, I think it can also take PSR levels.

People in the past will /should have been doing `Kohana::$log->attach($writer, Log::CRITICAL, Log::NOTICE)` and in the new implementation they'll do `$writer->set_levels_range(LogLevel::NOTICE, LogLevel::CRITICAL)`.

I don't think the actual value in the constants matters that much. Also forcing them to change their call may help them notice that it's now (correctly) `($min_range, $max_range)` instead of `($max_range, $min_range)`.

enov:
I have taken care of all change requests:
- Set the 'time' in test (dataprovider NULL) https://github.com/kohana/core/pull/563#discussion_r43485352 in https://github.com/kohana/core/commit/df8652c64df059f2d4e0a46887629a2945db704d
- $logger->set_write_levels([]) explicit, default to contain all levels https://github.com/kohana/core/pull/563#discussion_r43486184 in https://github.com/kohana/core/commit/dde274d06cfc0226d899921821757104f2775b2b
- **Still accepting different types of levels** but made distinct getters instead of the ambiguous one. https://github.com/kohana/core/pull/563#discussion_r43485486 in https://github.com/kohana/core/commit/14189a683500bd98569e419c36058fc3020d9802
- Fixed some doc blocs https://github.com/kohana/core/pull/563#discussion_r43485502 in https://github.com/kohana/core/commit/e767c0985c4e6e44d2702112be8f81a73a23836e
- Made filtering more efficient, but maybe less readable, relied on map of levels https://github.com/kohana/core/pull/563#discussion_r43485748 again in https://github.com/kohana/core/commit/dde274d06cfc0226d899921821757104f2775b2b
- Renamed `levels` as `write_levels` as well as respective setters/getters, **but still not 100% convinced, maybe `filter_levels`**? https://github.com/kohana/core/pull/563#discussion_r43485846 in https://github.com/kohana/core/commit/c7302401ae134e4b08795e3a303a9272968bb288
- Added some preliminary tests to filter, more to come. https://github.com/kohana/core/pull/563#discussion_r43485989 in https://github.com/kohana/core/commit/38515d0b201bea4899354237ef14297555029cfd

I welcome all comments, my scrutinizers :)

enov:
...and maybe we should have a :cake: , been a year :)

enov:
It would be nice if you could comment on the last commits, so that I might be able to resume.

@acoulton I am specifically awaiting your comments on https://github.com/kohana/core/commit/14189a683500bd98569e419c36058fc3020d9802 regarding the types of getters/setters of write_level, those are still expecting both integer _and_ psr/string types. Getters and setters are not symmetric anymore, but I left the door open for integer levels for those who prefer to use it, as it seems there is an [open/unresolved issue](https://github.com/php-fig/log/pull/21) on the FIG. Yeah, there might be less than 1% chance for PSR-3 v2, or a follow-up PSR-n, but keeping accepting integer levels in the API might alleviate us from a technical dept in the future. This would also keep consistence with the Log.

Also, cc @rjd22 if you have any reservations.

enov:
Please review https://github.com/kohana/core/commit/7dfcc7bb39349c0380537e0316a79b16b8d1d4ef.
Thanks :)

rjd22:
@enov I really don't like the name `filter_map`. It's kind of unclear what its meaning is or what it should do.

enov:
Thanks for reviewing @rjd22. What do you suggest?

rjd22:
@enov could you explain what it's task is, exactly? So I can understand it and think of a appropriate name.

enov:
`filter_map` is the variable that is actually being used when `filter()`ing. The other `filter` variable only holds the value that is being `set`, so it could `get` back.

`filter_map` could have been computed and used only locally within `filter()`, but that would duplicate some of the code used for validating/normalizing.

rjd22:
@enov what is it filtering? And what is the reason it needs to filter it?

enov:
The original `Log_Writer` had the option to limit what it could actually write, by selecting the levels it wants to filter.

enov:
I should probably go for `filter_levels` instead of just `filter`?

rjd22:
I would name it something like. `enabled_log_levels`. I would remove the booleans from it and enable them if they are in the list and disable them when they are not.

I would remove the $this->filter parameters
I would remove `get_filter_map` or make it private if needed inside
I would rename `get_filter` to `get_enabled_log_levels`
I would rename `set_filter` to `set_enabled_log_levels`

Also make every function that you don't use/need outside the class private. Making everything public will make future refactoring a problem.

I'm sorry for the changes but please remember that naming is extremely important.

enov:
Sorry for the late reply @rjd22. Personally I don't mind, even though I find them a bit verbose :)

I might remove `get_filter_map` altogether :)

acoulton:
@enov apologies for the very slow response.

I prefer `writeable_log_levels` to `enabled_log_levels` but either is OK.

I disagree with @rjd22, I think a hash of `level => boolean` is more useful than an array of just the levels that are active. But I would initialise it using the list of defined levels from `Kohana_Log` rather than have a duplicate hardcoded list of them all. 

I don't care too much about this though if you guys disagree, but we should only provide the values in one format - the user can convert from array => map or vice versa if required.

I still increasingly strongly think you should only accept PSR levels when setting the filter (and validate this). All log messages received by the writer already have a PSR log level regardless of input (they're converted in `Kohana_Log::log`). If you force PSR levels for the filter, the developer will almost certainly just reference the defined PSR constant in their code. This has the least possible confusion and is broadly safe whatever operating system they're on and regardless of what actual value PSR assign to the constants in future.

In the worst case if in future PSR switch to integer log levels per RFC 5424 (`LogLevel::EMERGENCY == 0`) but the user is on a system where `LOG_EMERG == 1` you have no idea which kind of integer the user passed in and whether they want to filter emergency or alert messages.

classes/Kohana/Log.php
```diff
@@ -10,7 +10,7 @@
  * @copyright  (c) 2008-2012 Kohana Team
  * @license    http://kohanaframework.org/license
  */
-class Kohana_Log {
+class Kohana_Log extends Psr\Log\AbstractLogger implements Kohana_Log_Buffer {
 
 	// Log message levels - Windows users see PHP Bug #18090
 	const EMERGENCY = LOG_EMERG;    // 0
```
acoulton: I realise this has always been inconsistent (and always slightly will be for the syslog writer and/or if people are using syslog constants in their own code).

But it makes no sense that for the new logger on some (eg windows) systems `Kohana_Log::EMERGENCY == 1 == \Psr\Log\LogLevel::ALERT`.

In order of preference, we should:
- remove these constants (biggest bc break, best for us)
- make them point directly to the equivalent PSR LogLevel constant
- hardcode the integer values to match the keys of `$log_levels` below (unlikely to break anyone else's code unless they're a windows user relying specifically on `Kohana_Log::EMERGENCY === 1`)

acoulton:
@enov you should upgrade your :cake: to :birthday: (or possibly even :beers: )

enov:
Thanks @acoulton for the review, we should definitely :tada: once this gets merged.

classes/Kohana/Log.php
enov: I would prefer going with the last suggestion, to hardcode the integer values. Yes, it removes the inconsistency. And in `Log_Syslog` we're already reverting back to `Log_` values.

enov:
I have taken away the actual log filtering from the Log_Writer. It can be used maybe elsewhere, not only in the writer. The `Log_Writer` is now `Filter_Aware`, and you can attach as many filters to it as you like, similar to the API you attach writers to the log.

We might only need `Log_Filter_Level` class to be at par, feature-wise, with what already we had. But I have provided some other suggestions, since it now opens new ways of filtering logs. The filters `Log_Filter_Body`, `Log_Filter_File` and `Log_Filter_Exception` could stay if there is consensus about the implementations, or could be removed, and be provided as examples in the docs. `Log_Filter_Union` is there to have an OR union (instead of intersection, AND) when filtering logs with multiple filter requirements.

So much for filtering, I guess. Here's the API:

``` php
$writer->attach_filter(new Log_Filter_Level([LogLevel::EMERGENCY, LogLevel::ALERT]));
```

What we used to do, is to specify the level filtering when attaching the writer to the log:

``` php
$log->attach_writer($writer, Log::EMERGENCY, Log::ALERT);
```

Let me know what you think. Thanks :smile:

rjd22:
@enov I didn't look in depth but as far as I can see it looks good.

enov:
Thanks @rjd22.

enov:
At this moment, `Kohana_Log_Buffer` interface methods accept a `Log_Writer` which is an abstract class. Being a partial implementation, it is better to be renamed to `Log_Writer_Base`, allowing the introduction of an interface `Log_Writer`.

To remove the confusion, the `Log_File` and the `Log_Syslog` should be renamed to `Log_Writer_File` and `Log_Writer_Syslog` respectively. So basically,

| Current class name | New class name |
| --- | --- |
| Log_Writer | Log_Writer_Base |
| Log_File | Log_Writer_File |
| Log_Syslog | Log_Writer_Syslog |

All those three files would nicely sit in the same directory.

Now, I am not able to proceed with those changes as it breaks the tests, because Koharness gets bootstraped with a `Log_File`. So, I wonder, should I only rename the `Log_Writer` and introduce the interface? Or should I proceed with all the changes and we make Koharness compatible? Or should I just forget about all those things and concentrate on finishing this PR as it is now?

rjd22:
@enov I would start wrapping up the PR. You can always make a new one fixing those names before the release.

I would not make a interface for a abstract class since it already enforces some kind of interface. So I would leave the name `Log_Writer` the same but rename the `Log_File` to `Log_Writer_File` and `Log_Syslog` to `Log_Writer_Syslog`.

acoulton:
@enov stop! (in the nicest possible way).

It's getting too hard to follow the changes in this PR already. I like `Log_Filter` but adding the new ones starts to significantly expand the scope of this branch.

I recommend:
- Keep `Log_Filter_Level` - but make it `Log_Filter_PSRLevel` and be strict about the levels you accept (sorry, I'm not going to keep banging on about that - your new design makes it much easier for someone to implement a filter using some other kind of level if they really want to).
- Remove all the other `Log_Filter` implementations - including `Log_Filter_Union` - into a new branch
- Deal with any renaming of `Log_File` etc in a new separate branch
- Change `Log::EMERGENCY` etc to be hardcoded integer values
- Add upgrade notes, documentation etc
- Then we can do a final review of this PR
- And then once this is merged you can make new PRs for your new branches so we can look at them properly.

enov:
I guess you're both right. I need wrap up. I guess I got drifted away at some point.

acoulton:
it's very easily done :grinning:

enov:
This is ready for final review :)

enov:
@acoulton @rjd22 I am hoping to see this reviewed. Is there anything missing? Or should I proceed with squashing commits together?

tests/kohana/ExceptionTest.php
```diff
@@ -96,4 +96,108 @@ public function test_text($exception, $expected)
...
+	 * @param string $expected  expected output
+	 */
+	public function test_log_not_kohana_log($exception, $expected) {
+		$temp = Kohana::$log;
+		Kohana::$log = new Kohana_ExceptionTest_Log_Psr_Mock();
+		Kohana_Exception::log($exception);
+		$this->assertSame($expected, Kohana::$log->logs[0]['message']);
+
+		// clean up
+		Kohana::$log = $temp;
```
acoulton: @enov I would move this to the `setUp` and `tearDown` method - as you have it the cleanup won't run if the assertion fails which may cause issues elsewhere.

tests/kohana/Log/FileTest.php
```diff
@@ -0,0 +1,216 @@
...
+		];
+	}
+
+	/**
+	 * Tests logging with the file writer
+	 *
+	 * @test
+	 * @dataProvider provider_file
+	 */
+	public function test_file(array $log_entries, $expected)
```
acoulton, line 99: Can you add a test that multiple writer instances append properly to the file as expected? (or, that a single instance appends to an existing file with some content which is basically the same).

tests/kohana/Log/FileTest.php
```diff
@@ -0,0 +1,216 @@
...
+	/**
+	 * Sets up the test enviroment
+	 */
+	// @codingStandardsIgnoreStart
+	function setUp()
+	// @codingStandardsIgnoreEnd
+	{
+		parent::setUp();
+
+		$this->log_dir = sys_get_temp_dir()
```
acoulton: I think you should remove this directory in tearDown - or better yet use https://github.com/mikey179/vfsStream to test the file writer against a virtual filesystem.

tests/kohana/Log/WriterTest.php
```diff
@@ -0,0 +1,196 @@
...
+		];
+	}
+	
+	/**
+	 * Tests getters setters
+	 *
+	 * @test
+	 * @dataProvider provider_getters_setters
+	 */
+	public function test_getters_setters($property, $value, $exception)
```
acoulton: I wouldn't usually bother testing setters and getters (apart from testing if they're supposed to throw) - it's enough to test the methods that use the value that's been set IMHO.

tests/kohana/Log/WriterTest.php
```diff
@@ -0,0 +1,196 @@
...
+ * @group kohana.core.logging
+ * @group kohana.core.logging.syslog
+ *
+ * @package    Kohana
+ * @category   Tests
+ * @author     Kohana Team
+ * @copyright  (c) 2008-2012 Kohana Team
+ * @license    http://kohanaframework.org/license
+ */
+class Kohana_Log_WriterTest extends Unittest_TestCase
```
acoulton: It might be better to make this a base class and extend the other writer tests from it? That might reduce some of the duplication of filtering tests, and ensure that they're all doing eg format as expected - I think that's better than explicitly testing the methods of an abstract class.

tests/kohana/LogTest.php
```diff
@@ -61,51 +61,360 @@ public function test_attach_attaches_log_writer_and_returns_this()
...
+
+		// assertions
+		$this->assertSame($flush_immediately, $logger->get_immediate_flush());
+		$this->assertSame($expected_number_of_logs, $actual_number_of_logs);
+
+		// force flush and assert anew
+		$logger->flush();
+		$expected_number_of_logs = 1; // expect logs after forced flush
+		$actual_number_of_logs = count($writer->logs); // recalculate actual
+		$this->assertSame($expected_number_of_logs, $actual_number_of_logs);
```
acoulton: `$this->assertSame(1, count($writer->logs), 'Expect logs written after forced flush');` would be more readable in code and produce a more obvious failure exception message if it fails IMO.

acoulton:
@enov sorry I missed your earlier message. This is looking good I think, there's a few places you could make some small improvements to the tests but it's very close now.

enov:
Thanks @acoulton for the review. I will take care of the notes you left.

tests/kohana/Log/FileTest.php
enov: `mikey179/vfsStream` has been used https://github.com/kohana/core/commit/430a1d2e0cdbec8b197a43132b0ec988acababc5

tests/kohana/Log/WriterTest.php
enov: `Kohana_Log_AbstractWriterTest` is a base class containing filtering tests
https://github.com/kohana/core/commit/304678db4944edfa9811e4f4a37339f8449c697e

tests/kohana/LogTest.php
enov: done https://github.com/kohana/core/commit/a279f7e6ea7a7e6aa383a5ed86e1f4b1f4849f55

tests/kohana/Log/WriterTest.php
enov: Removed `Kohana_Log_WriterTest` https://github.com/kohana/core/commit/304678db4944edfa9811e4f4a37339f8449c697e

tests/kohana/Log/FileTest.php
enov, line 99: Tested that a single writer instance appends to an existing file https://github.com/kohana/core/commit/df30f2fb7b95f31b4bddc182cdb6408c56c4eee3

enov:
@acoulton I took care of the notes you left. Some issues though:
- When extending the base for filter tests, PHPUnit's autoload fails. It turned out that the `kohana` folder just under the `tests` folder is all small caps. This needs to be addressed in a future PR so that a PSR-0 autoloader would work. It's working now, because it's named `AbstractWriterTest.php` and that would make it come before the extending test classes, when PHPUnit traverses the filesystem.
- I have completely removed `Kohana_Log_WriterTest`. In case you think any test is needed, I'll just add it back.
- Composer is not downloading from source when Travis is testing on HHVM. I still did not fully investigate why. Is it related to https://github.com/kohana/kohana/pull/92?

Thanks :)

acoulton:
@enov this all looks good to me now. I'm not sure about the issue with travis/composer, the build output for 304678d looks normal to me?

enov:
Thanks for reviewing again, @acoulton. Yeah, it built normally after I restarted it, most probably Travis/HHVM issue.

acoulton:
@enov :beers: :cake: :balloon: :fireworks: and general :tada:

Thanks very much for this!

enov:
Wow :flushed: ... merged!

Thank you @acoulton and thank you all who participated: @cyrrill, @arteymix @Yahasana and @rjd22!

rjd22:
@enov Thanks a log for the hard work!!

enov:
> @enov Thanks a log for the hard work!!

Haha! Good one, @rjd22, thanks!

PR #564, opened 2014-11-04
Adds PHPDoc to Logger::log

cyrrill:


classes/Kohana/Log.php
```diff
@@ -219,7 +219,15 @@ public function write()
 			}
 		}
 	}
-
+	
+	/**
+	 * Logs with an arbitrary level.
+	 *
+	 * @param  mixed  $level    level of message
```
enov, line 226: Thanks for this @cyrgit, the `$level` should be `string`. 

Kindly ammend the commit message to

> Adds PHPDoc to Log::log

Also add a usage example, you can take one from the `Log::add`. Don't forget the braces in the message.

Thanks again!

cyrrill, line 226: @enov I was thinking about that, and wondering why in the PSR spec they define it as mixed?
https://github.com/php-fig/log/blob/master/Psr/Log/LoggerInterface.php#L108

cyrrill, line 226: @enov Really sorry to be OCD about this, don't mean to waste time on petty things, it's just that it's implementing a standard so just to make sure were thorough on this. For example monolog implements it as mixed as well?
https://github.com/Seldaek/monolog/blob/master/src/Monolog/Logger.php#L432

enov, line 226: Frankly, I don't know. Reading `Monolog\Logger`, it will throw an exception [at some point](https://github.com/Seldaek/monolog/blob/master/src/Monolog/Logger.php#L384), if `$level` is not `string`.

But, thanks for the thorough research @cyrgit. Leave it as `mixed` since PSR spec defines it that way.

cyrrill, line 226: @enov Yeah, it's a bit confusing the definition they provide for the log levels as string, but then mixed in the interface:

> A ninth method, log, accepts a log level as first argument. Calling this method with one of the log level constants MUST have the same result as calling the level-specific method. Calling this method with a level not defined by this specification MUST throw a Psr\Log\InvalidArgumentException if the implementation does not know about the level. Users SHOULD NOT use a custom level without knowing for sure the current implementation supports it.

It looks like that would open the possibility for an implementation to add more levels other than the ones provided, which aren't necessarily required to be strings as per the spec. The 8 standard levels must be implemented as strings to satisfy the contract, but others are allowed so as long as they are defined?

> Don't forget the braces in the message.

Not sure I understand so to clarify, do you mean in the commit message to write as Logger::log()  rather Logger::log, or something else?

enov, line 226: > The 8 standard levels must be implemented as strings to satisfy the contract, but others are allowed so as long as they are defined

Yeah. I think you're right.

> MUST throw a Psr\Log\InvalidArgumentException

Hmm, I see I did not comply with that. Did I miss anything else?

> Not sure I understand so to clarify

I meant if you could kindly add a usage example in the PHPDoc, similar to the way we have in the PHPDoc of the `Log::add` method. But that usage example should differ in that `$message` string variables need to have the braces around.

enov, line 226: Will merge this and add the usage example. Thanks @cyrgit!

cyrrill:
@enov Got caught up in other things, among them, trying to see if we could implement this interface:

```
namespace Psr\Log;
/**
 * Describes a logger-aware instance
 */
interface LoggerAwareInterface
{
    /**
     * Sets a logger instance on the object
     *
     * @param LoggerInterface $logger
     * @return null
     */
    public function setLogger(LoggerInterface $logger);
}
```

However, couldn't see a way how, since the Kohana class consumes the $logger as a static property rather than an instance one.  Kohana::$log  vs.  $kohana->log. I don't think we could do this on a static context?

On that note, it would be nice at some point to convert the Kohana kernel into an instance variable created at the bootstrap level, and then use Setter Injection with the prescribed setLogger() method, instead of the current ::$->attach() way. But I know that's not really feasible at least in this 3.4 version...

Thanks for merging this PR. I'm looking for small things I can do to join the Kohana dev team on a more regular basis. I've been using it since '09 on version 2.x, and finally I thought it would be nice to invest time in its core, rather than just as a user. :)

PR #565, opened 2014-11-06
Don't allow insecure, predictable random numbers for IVs

enov:
Silently use MCRYPT_DEV_URANDOM when
the chosen random number generator
is not one of those that are considered secure.

### DIFF
--- a/classes/Kohana/Encrypt.php
+++ b/classes/Kohana/Encrypt.php
@@ -153,41 +153,6 @@ class Kohana_Encrypt {
 	 */
 	public function encode($data)
 	{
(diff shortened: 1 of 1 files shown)


enov:
This is related to #540 and #545

acoulton:
@enov this looks great, but should we document this behaviour on the $_rand property up at the top so it doesn't take people by surprise? Also, is there any reason we can't default that property to MCRYPT_DEV_URANDOM rather than NULL to make the standard behaviour even more obvious?

enov:
:+1:

enov:
done

PR #566, opened 2014-11-13
add file driver config can be save like the database

ljyf5593:
add file driver config can be save like the database

### DIFF
--- a/classes/Kohana/Config/File/Writer.php
+++ b/classes/Kohana/Config/File/Writer.php
@@ -1,0 +1,56 @@
+<?php
+/**
+ * File-based configuration writer. Multiple configuration directories can be
+ * used by attaching multiple instances of this class to [Kohana_Config].
+ *
+ * @package    Kohana
+ * @category   Configuration
+ * @author     Kohana Team
+ * @copyright  (c) 2014 Kohana Team
+ * @license    http://kohanaframework.org/license
+ */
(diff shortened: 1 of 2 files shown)


classes/Kohana/Config/File/Writer.php
```diff
@@ -0,0 +1,56 @@
...
+     * @param array       $config The configuration to write
+     * @return boolean
+     */
+    public function write($group, $key, $config)
+    {
+        $this->_loaded_keys[$group][$key] = $config;
+
+        $config_file = APPPATH.DIRECTORY_SEPARATOR.'config'.DIRECTORY_SEPARATOR.$group.'.php';
+
+        $result = file_put_contents($config_file, "<?php\nreturn ".var_export($this->_loaded_keys[$group], TRUE).';');
```
shadowhand, line 52: this should probably check for `is_writable` first.

shadowhand:
I think this should be updated to use the same pattern as https://github.com/kohana/database/pull/77

classes/Kohana/Config/File/Writer.php
enov, line 52: @ljyf5593 very sorry to come back to you this late. Could you kindly note @shadowhand 's comment here, as I think the sooner we check for `is_writable` the better. And maybe should we add a `try... catch` block here for the same reasons?

Thanks.

enov:
This needs unittesting too. Thank you for your time.

neo22s:
needs to be reviewed tested and considered for 4.x

PR #348, opened 2013-04-16
refs #4701: Prevent hash length extension attack.

birkir:
http://blog.whitehatsec.com/hash-length-extension-attacks/

PR #569, opened 2014-11-18
Prevent hash length extension attack

enov:
refs Redmine R4701
refs Github #348
see http://blog.whitehatsec.com/hash-length-extension-attacks/

Thanks @birkir


shadowhand:
:+1: x :100:

PR #570, opened 2014-11-18
Arr::get overhaul

enov:
- Relax constraint to make it work with instanceof ArrayAccess
- using `isset` at the beginning for performance reasons
- Use `array_key_exists` because `isset` will return FALSE for
  values explicitly defined as NULL
- Remove _workaround_ text as HHVM is compatible with PHP

Thanks @hkdobrev 

Related to #299, #301, #330

### DIFF
--- a/classes/Kohana/Arr.php
+++ b/classes/Kohana/Arr.php
@@ -267,14 +267,32 @@ class Kohana_Arr {
 	 *     // Get the value "sorting" from $_GET, if it exists
 	 *     $sorting = Arr::get($_GET, 'sorting');
 	 *
-	 * @param   array   $array      array to extract from
-	 * @param   string  $key        key name
-	 * @param   mixed   $default    default value
+	 * @param   array|ArrayAccess|NULL   $array      array to extract from
+	 * @param   int|string               $key        key name
+	 * @param   mixed                    $default    default value
 	 * @return  mixed
+	 * @throws Kohana_Exception If $array is not of expected types
 	 */
 	public static function get($array, $key, $default = NULL)
 	{
-		return isset($array[$key]) ? $array[$key] : $default;
+		// using isset for performance reasons
+		if (isset($array[$key]))
+			return $array[$key];
+
+		if ($array === NULL)
+			return $default;
+
+		if (is_array($array))
+		{
+			return array_key_exists($key, $array) ? $array[$key] : $default;
+		}
+
+		if ($array instanceof ArrayAccess)
+		{
+			return $array->offsetExists($key) ? $array[$key]: $default;
+		}
+
+		throw new Kohana_Exception('Argument 1 to Arr::get() must be an array or an instance of ArrayAccess');
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


classes/Kohana/Arr.php
```diff
@@ -275,19 +275,25 @@ public static function range($step = 10, $max = 100)
...
-			return $array->offsetExists($key) ? $array->offsetGet($key) : $default;
-		} else {
-			return isset($array[$key]) ? $array[$key] : $default;
+		// using isset for performance reasons
+		if (isset($array[$key]))
+			return $array[$key];
+
+		if ($array instanceof ArrayAccess)
+		{
+			return ($array->offsetExists($key)) ? $array[$key] : $default;
```
kemo: Remove the surrounding braces, this is a terminating condition/body (as on the previous line)

classes/Kohana/Arr.php
```diff
@@ -275,19 +275,25 @@ public static function range($step = 10, $max = 100)
...
+		}
+
+		if (is_array($array))
+		{
+			return (array_key_exists($key, $array)) ? $array[$key] : $default;
```
kemo: same here

classes/Kohana/Arr.php
enov: Thanks @kemo

kemo:
:+1:

hkdobrev:
:+1:

classes/Kohana/Arr.php
```diff
@@ -275,20 +275,22 @@ public static function range($step = 10, $max = 100)
...
-		if ($array instanceof ArrayObject) {
-			// This is a workaround for inconsistent implementation of isset between PHP and HHVM
-			// See https://github.com/facebook/hhvm/issues/3437
-			return $array->offsetExists($key) ? $array->offsetGet($key) : $default;
-		} else {
-			return isset($array[$key]) ? $array[$key] : $default;
-		}
+		// using isset for performance reasons
+		if (isset($array[$key]))
+			return $array[$key];
```
shadowhand, line 287: i don't like the lack of `{ }` enclosure around these statements.

classes/Kohana/Arr.php
```diff
@@ -275,20 +275,22 @@ public static function range($step = 10, $max = 100)
...
+
+		if ($array instanceof ArrayAccess)
+			return ($array->offsetExists($key)) ? $array[$key] : $default;
```
shadowhand: and yet there is extra `( )` wrapping around a ternary condition that doesn't need it? :-1:

tests/kohana/ArrTest.php
```diff
@@ -188,12 +211,12 @@ public function provider_get()
 	 *
 	 * @test
 	 * @dataProvider provider_get()
-	 * @param array          $array      Array to look in
-	 * @param string|integer $key        Key to look for
-	 * @param mixed          $default    What to return if $key isn't set
-	 * @param mixed          $expected   The expected value returned
+	 * @param array|ArrayAccess|NULL $array      Array to look in
```
shadowhand: what's `NULL` doing here?

enov:
I have removed the extra parenthesis, as well as the extra `NULL`.

PR #572, opened 2014-11-22
Add UTF-8 support for Text::ucfirst

enov:
By using UTF8::substr, UTF8::strtoupper, UTF8::strlen

Thanks to @Come2Daddy PR #521

### DIFF
--- a/classes/Kohana/Text.php
+++ b/classes/Kohana/Text.php
@@ -238,14 +238,24 @@ class Kohana_Text {
 	 *
 	 *      $str = Text::ucfirst('content-type'); // returns "Content-Type"
 	 *
-	 * @param   string  $string     string to transform
-	 * @param   string  $delimiter  delimiter to use
-	 * @return  string
+	 * @param string $string string to transform
+	 * @param string $delimiter delimiter to use
+	 * @return string
+	 * @uses UTF8::substr
+	 * @uses UTF8::strtoupper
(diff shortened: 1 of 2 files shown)


classes/Kohana/Text.php
```diff
@@ -238,14 +238,24 @@ public static function random($type = NULL, $length = 8)
...
+	 * @uses UTF8::strlen
 	 */
 	public static function ucfirst($string, $delimiter = '-')
 	{
+		// ucfirst closure with UTF-8 support
+		$utf8_ucfirst = function($word)
+		{
+			$first_char = UTF8::substr($word, 0, 1);
+			$remaining = UTF8::substr($word, 1, UTF8::strlen($word));
+			return UTF8::strtoupper($first_char) . UTF8::strtolower($remaining);
```
cs278, line 255: This is a change in behaviour as `ucfirst()` only changes the case of the first character.

Come2Daddy, line 255: You are right, it's more like ucword() behaviour, except we specify delimiter.

Just get the code optimized, to use UTF8::ucfirst() instead of a closure.

enov, line 255: @cs278 you are right, the `Text::ucfirst` should have been named `Text::ucwords`, in the first place.

Note that I am not changing the behaviour of the existing functionality.

Issue #571, opened 2014-11-19
Namespaces for controllers modules and helpers


rjd22:
We should start thinking about implementing namespaces in Kohana and it's modules but also decide on a vendor / prefix.

Like `\Kohana\<module>\<class>`

Also enable users to namespace their own application. And point the routes to the namespace.

What do you guy's think? (Typed this on my mobile so keeping it short)

rjd22:
@enov @lenton @acoulton @shadowhand  pinging you for some opinions :)

kemo:
There was discussion on this topic before;
http://forum.kohanaframework.org/discussion/9740/kohana-and-namespaces/p1

Anyway, this definitely gets a :+1: from me - right after we make some things clear;
- Which version should we aim this for (maybe this should be Kohana4)? 
- Should we convert to PSR-1/2 at the same time?
- Therefore, should we use Ohanzee components where possible / strip the framework completely
- Would CFS / transparent extensions stay?

enov:
I am all for using namespaces @rjd22 :smile:

But the question remains: at what stage?

Limiting the scope of features for 3.4 would help us be able to release it at some point.

A namespaced version of Kohana could go out as a separate version on it own. I mean no new features in that release, except that it is a new Kohana based on the older version, but namespaced.

Another option would be to namespace while we work on some classes. But in that case we would have partly namespaced and partly non-namespaced classes. We can give preferences to the classes that are internally instantiated first. I mean, usually, a user does not do a `new Log()` as it is done internally somewhere in `Kohana::init`. That way, the new version is upgradeable by the user. Upgradeable releases is something I care about, as opposed to all new full-rewrite releases.

Keep in mind we always support 2 dev versions. Refactoring those class names means refactoring everywhere those are instantiated. It means going farther from the previous version. It means more merge conflicts when we try to merge up bug fixes from below.

> Also enable users to namespace their own application

Users can already namespace their classes in 3.3 as it is PSR-0 compliant - except for the Controller classes, I guess.

enov:
> Also enable users to namespace their own application. And point the routes to the namespace.

Yeah, I see now that you are concerned about the Controllers to be able to be namespaced.

acoulton:
I think this would be really good - for starters it would make working with tools like PHPSpec, and integrating with other components and projects, much easier.

It does feel like a big change to get in for 3.4, and I agree with @enov that we should avoid 3.4 becoming the "fix absolutely everything to the point we never actually release it" version.

[...]

PR #578, opened 2014-11-30
Load controllers with FQCN.

rjd22:
Fixes part of #571 

It's BC and works by the following syntax:

``` php
Route::set('default', 'controller-name/<action>')
    ->defaults(array(
        'controller' => '\Acme\Module\Controller\Demo'
    ));
```

The <controller> syntax won't work but I think that when using FQCN you should not use variable controller parameters.

@acoulton or @enov can you please review if this is okay?

### DIFF
--- a/classes/Kohana/Request/Client/Internal.php
+++ b/classes/Kohana/Request/Client/Internal.php
@@ -45,6 +45,12 @@ class Kohana_Request_Client_Internal extends Request_Client {
 			$prefix .= str_replace(array('\\', '/'), '_', trim($directory, '/')).'_';
 		}
 
+		// Use the FQCN to load the Controller
+		if (substr($controller, 0, 1) === '\\')
+		{
+			$prefix = '';
+		}
+
 		if (Kohana::$profiling)
 		{
 			// Set the benchmark name
--- a/classes/Kohana/Route.php
+++ b/classes/Kohana/Route.php
@@ -351,6 +351,14 @@ class Kohana_Route {
 			return $this->_defaults;
 		}
 
+		if (isset($defaults['controller']) AND substr($defaults['controller'], 0, 1) === '\\')
+		{
+			if (isset($defaults['directory']))
+			{
+				throw new Kohana_Exception('Route directory should not be set when the controller is a FQCN.');
+			}
+		}
+
 		$this->_defaults = $defaults;
 
 		return $this;
--- a/guide/kohana/routing.md
+++ b/guide/kohana/routing.md
@@ -68,6 +68,8 @@ If a key in a route is optional (or not present in the route), you can provide a
 
 [!!] The `controller` and `action` key must always have a value, so they either need to be required in your route (not inside of parentheses) or have a default value provided.
 
+[!!] When using the `Fully Qualified Class Name` or `FQCN` for the `controller` you aren't allowed to set `directory`. The reason is that when using the `FQCN` the autoloader will look for the right location of the file.
+
 [!!] Kohana automatically converts controllers to follow the standard naming convention. For example /blog/view/123 would look for the controller Controller_Blog in classes/Controller/Blog.php and trigger the action_view() method on it.
 
 In the default route, all the keys are optional, and the controller and action are given a default.   If we called an empty url, the defaults would fill in and `Controller_Welcome::action_index()` would be called.  If we called `foobar` then only the default for action would be used, so it would call `Controller_Foobar::action_index()` and finally, if we called `foobar/baz` then neither default would be used and `Controller_Foobar::action_baz()` would be called.
@@ -79,6 +81,24 @@ You can also use defaults to set a key that isn't in the route at all.
 TODO: example of either using directory or controller where it isn't in the route, but set by defaults
 
 ### Directory
+
+### Selecting controllers by their FQCN
+
+You can also use the `Fully Qualified Class Name` to find the controller. It will enable you to store your controllers at the location of your choosing as long as the autoloader is able to find it.
+
+When using `FQCN` controllers the controllers will not use any prefixed or affixes so `\Acme\Module\Controller\Demo` will point to `class Demo {}` and `\Acme\Module\Controller\DemoController` will point to `class DemoController {}`. You will have full flexibility on how you name you classes.
+
+Because of the autoloading the `directory` default value will not work and setting it will result in a `Exception` being thrown.
+
+An example of a `FQCN` route:
+
+    /*
+     * Routes using FQCN
+     */
+    Route::set('default', 'controller-name/<action>')
+        ->defaults(array(
+            'controller' => '\Acme\Module\Controller\DemoController'
+        ));
 
 ## Route Filters
 
--- a/tests/kohana/RouteTest.php
+++ b/tests/kohana/RouteTest.php
@@ -522,6 +522,24 @@ class Kohana_RouteTest extends Unittest_TestCase
 	}
 
 	/**
+	 * When using a FQCN test if the directory is not set else throw an exception
+	 *
+	 * @test
+	 * @covers Route::defaults
+	 */
+	public function test_defaults_throws_exception_when_setting_fqcn_and_directory()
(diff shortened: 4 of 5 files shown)


enov:
I think it is ok for full BC with the old routing system.

For consistancy between the two methods, however, I would have prefered the notion of the prefix as well as the directory to go away. Not sure though how it can be done.

kemo:
Maybe it should rather check for existance of `\` anywhere else except the first char because that can also be referring to a "namespaceless" class? 

So `strpos($controller, '\\')` might work better, this also goes for checking anywhere else. Opinions?

rjd22:
@kemo a namespaceless class should also begin with a \ like \Kohana. Because it's still in a name space only the global one.

kemo:
Yes but as things stand right now, those won't support directories (unless written without the initial `\`)

acoulton:
@rjd22 thanks for this, my thoughts:
- What you have is right for FQCN - it makes no sense for them to support directories.
- Ideally I'd like to see support for `<controller>` params too - with a separate `namespace` param
- I don't think we should drop `directory` and the `Controller_` prefix for legacy routes in 3.4 - people will be using modules that define routes as well as their own, and it's too big a change to ask them to update them all.
- The throw should be in Route::set, not in the request client - so that it's guaranteed to trigger with a sensible trace as soon as the developer first defines an invalid route, not later when (if) they happen to visit a matching URL.
- Obviously this would need tests and also an update to the [routing docs](https://github.com/kohana/core/blob/3.3/master/guide/kohana/routing.md) before merge.

acoulton:
Actually, just a thought - we could satisfy several of these cases and keep BC with a `controller_format` parameter that defaults to `\Controller_:directory:controller`.

If `<controller>` begins with a backslash, treat it as a FQCN with no further processing.

Otherwise, build a class name by passing the format through strtr:

``` php
# /welcome => Controller_Welcome as now
Route::set('default', '<controller>/<action>')
    ->defaults(array());

# /welcome => Controller_Other_Welcome as now
Route::set('default', '<controller>/<action>')
    ->defaults(array(
        'directory' => 'Other'
    ));

# /welcome => \My\Controller\Welcome
Route::set('default', '<controller>/<action>')
    ->defaults(array(
        'controller_format' => '\My\Controller\:controller'
    ));

# /welcome => \My\Whatever\WelcomeController
Route::set('default', '<controller>/<action>')
    ->defaults(array(
        'controller_format' => '\My\Whatever\:controllerController'
    ));

# /directory/welcome => \My\Whatever\Directory_WelcomeController - yuk, but supported
Route::set('default', '<directory>/<controller>/<action>')
    ->defaults(array(
        'controller_format' => '\My\Whatever\:directory:controllerController'
    ));
```

That would be fully BC, but would also support url-based controllers in namespaces with any (or no) prefix/suffix/whatever and directories if you really need them.

Thoughts?

loonies:
@acoulton This will add another layer of complexity in the route
definition. Lets keep it simple.

I would like to see the `Controller_` prefix and `directory` parameter
go away. The `Controller_` prefix forces naming your controller classes
in a certain way, which is not flexible.

What if we introduce the `namespace` parameter having `NULL` value by
default and we just glue `namespace` and `controller` together?

Basically we use `namespace` instead of `directory` when using FQCN.

If we want to drop `directory` parameter and the `Controller_` prefix we
should raise `E_DEPRECATED` in 3.4 and remove them later (e.g. 4.0).

Not tested though, but should be fully BC.

``` php
# /welcome => Controller_Welcome (BC)
Route::set('default', '<controller>/<action>')
    ->defaults(array())

# /welcome => Foo_Controller_Welcome (BC)
Route::set('default', '<controller>/<action>')
    ->defaults(array(
        'directory' => 'Foo'
    ))

# /welcome => Controller_MyWelcome (BC)
Route::set('default', 'welcome/<action>')
    ->defaults(array(
        'controller' => 'MyWelcome'
    ))

# /welcome => \Acme\Module\Ctnl\Welcome
Route::set('default', '<controller>/<action>')
    ->defaults(array(
        'namespace' =>'\Acme\Module\Ctnl'
    ))

# /welcome => \Acme\Module\Ctnl\MyWelcome
# $namespace !== NULL
Route::set('default', 'welcome/<action>')
    ->defaults(array(
        'namespace'  => '\Acme\Module\Ctnl',
        'controller' => 'MyWelcome
    ))

# /welcome => \Acme\Module\Ctnl\MyWelcome
# strpos($controller, '\\')
Route::set('default', 'welcome/<action>')
    ->defaults(array(
        'controller' => '\Acme\Module\Ctnl\MyWelcome'
    ))
```

rjd22:
Shouldn't we just keep things simple for now? More complex implementations can always be implemented if there is a need for it.

I will move the check to the router and add documentation for it.

acoulton:
@rjd22 sorry, you're right - let's just start by supporting explicit FQCN in the definition. As a side-effect that might encourage more people to use at least a route per controller, which would be no bad thing...

And we can add more complex implementations later if we need them. Obviously there's also scope for people to do that with filters now.

I think let's merge this once you've done the test/docs/moved the exception.

@loonies that would at least satisfy your "get rid of the prefix", "drop directory" and "keep it simple"?

rjd22:
@acoulton sorry I haven't made the time to split this up. I will try to do so next weekend. Just giving a heads up.

acoulton:
@rjd22 no worries, I hadn't made the time to even reply :)

loonies:
Keep it simple as it is currently, without `controller_format` parameter
proposed by you.

We need the `directory`, and the `Controller` parameter only for BC, but
they are not needed with FQCN.

rjd22:
@acoulton After looking I think the check is on the right spot. The router does not care about the directory or the FQCN the Kohana internals do because it has to locate the file. Maybe we should change the message the Route class just doesn't have a appropriate spot for it.

acoulton:
@rjd22 I think objects should be responsible for validating their own state. If it's not valid for a Route to have both a FQCN controller param and a directory param, then it should refuse to accept them.

Absolutely Route shouldn't care about file location, but `directory` here doesn't really mean that, it means `(part of the) controller name prefix`. The fact that Kohana happens to look for a `Controller_Directory_` class in a `Controller/Directory/Controller.php` path is incidental.  It's not valid to specify a complete class name and a class name prefix for the same route.

If anything, the misplaced logic is where Kohana request client assembles the class name, after the Route object [separately handles `ucwords`ing the controller and directory params](https://github.com/kohana/core/blob/3.3/master/classes/Kohana/Route.php#L448). If I was building from scratch, the Route class would have complete responsibility for getting from URL + definition to class name, method name and parameters. The request client would then just be responsible for checking that class/method combination exists and handling dispatch.

From a pragmatic point of view, what's more useful if you define a route with an FQCN and a directory (classname prefix):
- an exception thrown on every request, with a trace back to the Route::set that created the invalid state?
- an exception thrown by the request client when the request happens to match, with a trace that goes nowhere near the offending code?

rjd22:
The problem is that we're talking about default values. So we won't be able to check in route set. The only way is the check if both directory and a FQCN is set when calling defaults(). The a exception will be thrown when setting the routes. Maybe that is better?

rjd22:
@enov & @acoulton I've moved the checks and tests. Please see if this can be merged

acoulton:
@rjd22 looks great, but shouldn't there also be a test that `Request_Client_Internal` is loading the correct controller? Also would you be able to squash this to a single commit now please?

rjd22:
@enov @acoulton it will be quite hard to test `Request_Client_Internal` as it is now. I swuashed the commits. Can someone please check and merge?

acoulton:
@rjd22 I've added tests, what do you think?

acoulton:
@rjd22 PS, very happy with your commit, all looks good to me. If we're happy with the new test style we can in future add coverage of more of the untested Request_Client_Internal behaviour.

If you or @enov are happy with the test go ahead and merge.

rjd22:
Loons great.

PR #581, opened 2014-12-05
Validate multi dimensional array

evtaylor:
See http://dev.kohanaframework.org/issues/4035

``` php
$array = array(
    'user' => array(
        'name' => 'Evan'
    )
);

$validation = Validation::factory($array)
    ->rule('user.name', 'not_empty'); 

```

### DIFF
--- a/classes/Kohana/Arr.php
+++ b/classes/Kohana/Arr.php
@@ -659,5 +659,43 @@ abstract class Kohana_Arr {
 			array(":type" => Debug::type($array))
 		);
 	}
+	
+	/**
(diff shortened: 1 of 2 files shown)


rjd22:
@evtaylor this needs tests for both the array path as the Validator using the array path.

arteymix:
If we go multidimensional, what about a `*` notation to specify all keys within an array?

``` php
->rule(`users.*.name`, `not_empty`)
```

neo22s:
for 4.0.0 ?

PR #582, opened 2014-12-07
Update class Route method filter

WinterSilence:
Update description. Works as getter\setter. Protection from adding duplicate filters.

### DIFF
--- a/classes/Kohana/Route.php
+++ b/classes/Kohana/Route.php
@@ -362,13 +362,13 @@ class Kohana_Route {
 	 *     $route->filter(
 	 *         function(Route $route, $params, Request $request)
 	 *         {
-	 *             if ($request->method() !== HTTP_Request::POST)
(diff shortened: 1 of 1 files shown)


acoulton:
@WinterSilence thanks - but I think we shouldn't be adding more combined setter/getter methods. They are bad for readability, static code analysis and IDE autocompleting. They also add a lot of conditional logic and optional params that aren't really required, and are harder to configure mocks for.

Could you add the getter as a separate method?

What's the usecase for silently ignoring adding an existing filter (rather than throwing, surely that's unexpected)?

Also this will require unit tests before we can merge it.

WinterSilence:
> Could you add the getter as a separate method?

Ok, today fix it. How to call methods? `getFilters()` \ `setFilter()`?

acoulton:
Thanks yes I think `get_filters` is good for consistency with current coding style.

PR #583, opened 2014-12-07
Update Route matches

WinterSilence:
Reduced code. Fixed uppercase for params (controller, category).

### DIFF
--- a/classes/Kohana/Route.php
+++ b/classes/Kohana/Route.php
@@ -441,7 +441,7 @@ class Kohana_Route {
 
 		if ( ! empty($params['controller']))
 		{
-			if (strpos($params['controller'], '_') !== false)
+			if (strpos($params['controller'], '_') !== FALSE)
 			{
 				// PSR-0: Replace underscores with spaces, run ucwords, then replace underscore
 				$params['controller'] = str_replace(' ', '_', ucwords(str_replace('_', ' ', $params['controller'])));
--- a/classes/Kohana/Route.php
+++ b/classes/Kohana/Route.php
@@ -397,9 +397,8 @@ class Kohana_Route {
 	}
 
 	/**
-	 * Tests if the route matches a given Request. A successful match will return
-	 * all of the routed parameters as an array. A failed match will return
-	 * boolean FALSE.
+	 * Tests if the route matches a given [Request]. A successful match will return
+	 * all of the routed parameters as an array, a failed match will return FALSE.
 	 *
 	 *     // Params: controller = users, action = edit, id = 10
 	 *     $params = $route->matches(Request::factory('users/edit/10'));
@@ -411,34 +410,29 @@ class Kohana_Route {
 	 *         // Parse the parameters
 	 *     }
 	 *
-	 * @param   Request $request  Request object to match
-	 * @return  array             on success
-	 * @return  FALSE             on failure
+	 * @param   Request  $request  Request object to match
+	 * @return  mixed
 	 */
 	public function matches(Request $request)
 	{
 		// Get the URI from the Request
 		$uri = trim($request->uri(), '/');
 
-		if ( ! preg_match($this->_route_regex, $uri, $matches))
+		if ( ! preg_match($this->_route_regex, $uri, $params))
 			return FALSE;
 
-		$params = array();
-		foreach ($matches as $key => $value)
+		foreach ($params as $key => $value)
 		{
+			// Delete all unnamed keys
 			if (is_int($key))
 			{
-				// Skip all unnamed keys
-				continue;
+				unset($params[$key]);
 			}
-
-			// Set the value for all matched keys
-			$params[$key] = $value;
 		}
 
 		foreach ($this->_defaults as $key => $value)
 		{
-			if ( ! isset($params[$key]) OR $params[$key] === '')
+			if ( ! isset($params[$key]) OR empty($params[$key]))
 			{
 				// Set default values for any key that was not matched
 				$params[$key] = $value;
@@ -447,14 +441,27 @@ class Kohana_Route {
 
 		if ( ! empty($params['controller']))
 		{
-			// PSR-0: Replace underscores with spaces, run ucwords, then replace underscore
-			$params['controller'] = str_replace(' ', '_', ucwords(str_replace('_', ' ', $params['controller'])));
+			if (strpos($params['controller'], '_') !== false)
+			{
+				// PSR-0: Replace underscores with spaces, run ucwords, then replace underscore
+				$params['controller'] = str_replace(' ', '_', ucwords(str_replace('_', ' ', $params['controller'])));
+			}
+			else
+			{
+				$params['controller'] = ucfirst($params['controller']);
+			}
 		}
 
 		if ( ! empty($params['directory']))
 		{
-			// PSR-0: Replace underscores with spaces, run ucwords, then replace underscore
-			$params['directory'] = str_replace(' ', '_', ucwords(str_replace('_', ' ', $params['directory'])));
+			$params['directory'] = ucwords(str_replace(array('\\', '/'), ' ', $params['directory']));
+			$params['directory'] = str_replace(' ', DIRECTORY_SEPARATOR, $params['directory']);
+
+			if (strpos($params['directory'], '_') !== false)
+			{
+				// PSR-0: Replace underscores with spaces, run ucwords, then replace underscore
+				$params['directory'] = str_replace(' ', '_', ucwords(str_replace('_', ' ', $params['directory'])));
+			}
 		}
 
 		if ($this->_filters)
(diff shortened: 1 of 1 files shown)


classes/Kohana/Route.php
```diff
@@ -411,34 +410,29 @@ public function filter($callback)
...
-		$params = array();
-		foreach ($matches as $key => $value)
+		foreach ($params as $key => $value)
 		{
+			// Delete all unnamed keys
 			if (is_int($key))
 			{
-				// Skip all unnamed keys
-				continue;
+				unset($params[$key]);
```
acoulton, line 429: It might be clearer and faster to use `array_filter` here?

classes/Kohana/Route.php
```diff
@@ -447,14 +441,27 @@ public function matches(Request $request)
...
 		}
 
 		if ( ! empty($params['directory']))
 		{
-			// PSR-0: Replace underscores with spaces, run ucwords, then replace underscore
-			$params['directory'] = str_replace(' ', '_', ucwords(str_replace('_', ' ', $params['directory'])));
+			$params['directory'] = ucwords(str_replace(array('\\', '/'), ' ', $params['directory']));
+			$params['directory'] = str_replace(' ', DIRECTORY_SEPARATOR, $params['directory']);
+
+			if (strpos($params['directory'], '_') !== false)
```
acoulton, line 460: Doesn't this also require an else block with `ucfirst`? This replacement seems to be duplicate for directory and controller, could you extract a method?

I think this doesn't actually change behaviour, just saves some operations if there's no `_` in the param value? Might be even clearer/more efficient to do something like:

``` php
$param = preg_replace_callback(
    '/(^|_)\w/', 
    function($matches){return strtoupper($matches[0]);},
   $param
);
```

Perhaps worth benchmarking?

classes/Kohana/Route.php
```diff
@@ -473,10 +480,12 @@ public function matches(Request $request)
 				{
 					// Filter has modified the parameters
 					$params = $return;
+					// @todo if $return != $params, then uppercase values (directory, controller)
 				}
 			}
 		}
 
+		// @todo throw exception then some params not exists (controller, action)
```
acoulton, line 488: Isn't that covered already in the request client?

classes/Kohana/Route.php
```diff
@@ -473,10 +480,12 @@ public function matches(Request $request)
 				{
 					// Filter has modified the parameters
 					$params = $return;
+					// @todo if $return != $params, then uppercase values (directory, controller)
```
acoulton, line 483: I think if filters return controller/directory etc they should be responsible for returning in proper case. The only reason we do the case transformation is so that URLs can be lowercase. Application code should be case-sensitive just like everywhere else.

In fact one use for a filter would be to transform a URL to some different casing convention eg `mysql` to `mySQL` rather than `Mysql` so I think we shouldn't mess about with the return.

Also if you are introducing `@todos` in a PR please could you flag in the description so it's clear it's not ready to merge?

acoulton:
Thanks for this, it needs unit tests particularly to show what functionally changes (I think just that it now supports directory params including `\` or `/`?).

Also the commit message is not very clear, it's not enough to say "Fixed xxx", we need to know what about them wasn't right before and is now. Did you fix a bug, fix performance, improve implementation without changing behaviour? If it's a reported issue you should reference it, if not please include a full bug/issue description in the commit or PR.

classes/Kohana/Route.php
WinterSilence, line 429: To do this need create closure or there is basic function?

classes/Kohana/Route.php
WinterSilence, line 460: `preg_*` functions are slower, function "faster" objects (closure).
Need to fix code https://github.com/kohana/core/issues/571 ?

WinterSilence:
> I think just that it now supports directory params including `\` or `/`?

```
$params['directory'] = ucwords(str_replace(array('\\', '/'), ' ', $params['directory']));
$params['directory'] = str_replace(' ', DIRECTORY_SEPARATOR, $params['directory']);
```

> If it's a reported issue you should reference it, if not please include a full bug/issue description in the commit or PR.

My main problem is English translation (( It takes more time than writing code.

At first, we need to discuss whether changes are needed? After this, make adjustments and 
test final version.

classes/Kohana/Route.php
WinterSilence, line 488: I did not find it in request

classes/Kohana/Route.php
WinterSilence, line 483: > Also if you are introducing @todos in a PR please could you flag in the description so it's clear it's not ready to merge?

How do it?

classes/Kohana/Route.php
```diff
@@ -447,14 +441,27 @@ public function matches(Request $request)
 
 		if ( ! empty($params['controller']))
 		{
-			// PSR-0: Replace underscores with spaces, run ucwords, then replace underscore
-			$params['controller'] = str_replace(' ', '_', ucwords(str_replace('_', ' ', $params['controller'])));
+			if (strpos($params['controller'], '_') !== false)
```
kemo: Allcaps `FALSE`

classes/Kohana/Route.php
acoulton, line 460: Sure `preg_*` and callback has a penalty, but multiple `strpos`, `ucwords`, `str_replace` will also add up - they all have to scan the full string and switch between native and PHP repeatedly. Worth checking I think because it would definitely make the code much more readable to get it in a single operation.

classes/Kohana/Route.php
acoulton, line 429: Well `is_string` is probably the opposite of `is_int` for this case? 

But a closure with `! is_numeric` might be safer to be sure it filters `"1"` etc:

``` php
$params = array_filter(
    $matches, 
    function($key) { return ! is_numeric($key); },
    ARRAY_FILTER_USE_KEY
);
```

classes/Kohana/Route.php
acoulton, line 483: For example name the PR "[WIP] Update Route matches", and if it's `@todo` because you're not sure then in the description or as a line comment write something like "Question: should we uppercase controller/directory again after filters have run?"

classes/Kohana/Route.php
acoulton, line 488: My mistake, assumed it was but you're right so yes we should throw here if required params are missing.

classes/Kohana/Route.php
WinterSilence, line 460: code for #571

```
if (!empty($params['namespace'])) {
    if (strpos('\\', $params['namespace']) !== false) {
        $params['namespace'] = str_replace(' ', '\\', ucwords(str_replace('\\', ' ', $params['namespace'])));
    } else {
        $params['namespace'] = ucfirst($params['namespace']);
    }
}
```

WinterSilence, line 460: @acoulton also, your code create closure object at each call `Route::match()`, its wrong - need create closure only once and contain in protected property of route. Now `Route::match()` use something like this, but need fix it.

classes/Kohana/Route.php
WinterSilence, line 483: ok, thanks add it in new PRs.

classes/Kohana/Route.php
WinterSilence, line 429: your code call function -> function create object -> function call object in cycle, 
how it can work faster?

classes/Kohana/Route.php
acoulton, line 460: @WinterSilence #571 won't be merged, #578 is the implementation for that which is going to be much simpler.

My snippet with closure was just for example, yes obviously you'd probably optimise/tidy up for implementation.

classes/Kohana/Route.php
acoulton, line 429: `array_filter` is often a lot faster than anything with foreach, because the loop and the data management all happens in native code - there are many fewer userland operations and stack changes. You're right I'd not allowed for the overhead of the closure though which is higher than I'd expected.

We can't use my example above because `ARRAY_FILTER_USE_KEY` is PHP >= 5.6. For lower versions the best I've found is:

``` php
$int_keys = array_filter(array_keys($matches), 'is_numeric');
$params   = $int_keys ? array_diff_key($matches, array_fill_keys($int_keys, NULL)) : $matches;
```

I just did a quick benchmark of:
- the original code (foreach and copy valid elements to new array)
- your code (foreach and unset invalid)
- array_filter with native function (can't use)
- array_filter with closure (can't use)
- possible solution

|  | 4 string | 3 string + 1 int | 13 string | 10 string + 3 int |
| --- | --- | --- | --- | --- |
| copy valid | 24 μs | 22 μs | 69 μs | 68 μs |
| unset invalid | 20 μs | 19 μs | 58 μs | 58 μs |
| array_filter native | 6 μs | 6 μs | 7 μs | 8 μs |
| array_filter closure | 37 μs | 36 μs | 106 μs | 102 μs |
| array_filter + diff_key | 11 μs | 19 μs | 14 μs | 22 μs |

So it's basically always worth doing `array_filter` if there's a native function that does what you need, or if you can use a native function to find the values to reject and then `array_diff_key` to strip them out of the array.

Note also that it's fastest in the common (pure associative array) case, with only a small hit if it finds integer keys to remove.

PR #585, opened 2014-12-07
Update View.php

sergeyklay:
Fixed View::set


acoulton:
@sergeyklay thanks for this, a few issues:
- We can't merge to 3.3/master so I'll close this - please resubmit targeted to 3.3/develop
- Please use `OR` rather than `||`
- Can you provide a unit test that verifies the expected behaviour?

sergeyklay:
Done https://github.com/kohana/core/pull/586

PR #586, opened 2014-12-08
Update View::set

sergeyklay:


tests/kohana/ViewTest.php
```diff
@@ -80,4 +80,32 @@ public function test_instantiate($path, $expects_exception)
 			$this->assertSame(TRUE, $expects_exception);
 		}
 	}
+
```
acoulton, line 97: Please use tabs for indentation

tests/kohana/ViewTest.php
```diff
@@ -80,4 +80,32 @@ public function test_instantiate($path, $expects_exception)
...
+    /**
+     * Tests that we can set using array or traversable object
+     *
+     * @test
+     *
+     * @return null
+     */
+    public function test_set()
```
acoulton: This covers all the right things, but our convention would be to use a data provider rather than testing all three data types in a single test method. This test could give false positive if somehow they all set the value to NULL, or if it failed it wouldn't be immediately obvious which setter was failing.

tests/kohana/ViewTest.php
```diff
@@ -80,4 +80,32 @@ public function test_instantiate($path, $expects_exception)
...
+        $view1 = View::factory()->set('foo', 'bar');
+        $view2 = View::factory()->set(array('foo' => 'bar'));
+        $view3 = View::factory()->set($traversable);
+
+        $this->assertSame($view1->foo, $view2->foo);
+        $this->assertSame($view2->foo, $view3->foo);
+    }
+}
+
+class myData implements IteratorAggregate
```
acoulton: Could you just pass in ArrayIterator directly rather than create an extra class in the global namespace?

sergeyklay:
Done

enov:
> This covers all the right things, but our convention would be to use a data provider rather than testing all three data types in a single test method. 

@sergeyklay, I think @acoulton is referring to a data provider method (in addition) to the test method, something along these lines:

``` php
/**
 * Data provider to test_set
 */
public function provider_set()
{
    return array(
        array('foo', 'bar', 'foo', 'bar'),
        array(array('foo' => 'bar'), NULL, 'foo', 'bar'),
        array(new ArrayIterator(array('foo' => 'bar')), NULL, 'foo', 'bar'),
    );
}
/**
 * @dataProvider  provider_set
 */
public function test_set($data_key, $value, $test_key, $expected)
{
    $view = View::factory()->set($data_key, $value);
    $this->assertSame($expected, $view->$test_key);
}
```

enov:
Also, now that a Traversable is acceptable as data to be set to the view, there should be a note that we're not attaching the whole object to the view, i.e. the object's standard properties will not be available in the view context.

``` php
$object = new ArrayIterator(array('foo' => 'bar'));
// suppose $object had some property attached
$object->prop = 'Hello, I am a property';
// set it to the view
$view = View::factory()->set($object);
$view->prop; // won't work
// set it to the view, this time through the second param
$view = View::factory()->set('obj', $object);
$view->obj->prop; //works
```

classes/Kohana/View.php
```diff
@@ -272,18 +272,21 @@ public function set_filename($file)
...
 	 *     $view->set('foo', 'my value');
 	 *
-	 * You can also use an array to set several values at once:
+	 * You can also use an array or Traversable object to set several values at once:
 	 *
 	 *     // Create the values $food and $beverage in the view
 	 *     $view->set(array('food' => 'bread', 'beverage' => 'water'));
+     *
+     * [!!] Note: When setting with using Traversable object we're not attaching the whole object to the view,
+     * i.e. the object's standard properties will not be available in the view context.
```
enov: I am sorry to be picky here @sergeyklay could you fix the indentation?

enov:
I guess that's good to go for `View::set` :+1:
But now maybe we want to do the same for `View::set_global` :smile:
@acoulton ?

acoulton:
@sergeyklay thanks and sorry I wasn't clear about the test. Now that this is all good, could you rebase and squash it into a single commit to keep the history clean? Let me know if you need instructions.

@enov - argh, it's so long since I used `View::set_global` I forgot it even existed. We probably should do that one too - @sergeyklay do you have time to do that too in this PR, or shall I do it once we merge?

sergeyklay:
@acoulton `View::set_global` already updated

acoulton:
@sergeyklay but not tested... sorry to be picky :smiley:

sergeyklay:
Squashed commits

acoulton:
@sergeyklay thanks very much, nice one :smile:

sergeyklay:
:+1:

cyrrill:
What about deprecating set_global in 3.4?

acoulton:
@cyrgit +1 from me, can you make a PR/issue for that?

cyrrill:
@acoulton I've added the issue as you suggested:  https://github.com/kohana/core/issues/587

PR #590, opened 2014-12-16
Fixed Security::token

sergeyklay:
Fixed `Security::token`
- Improved hash algo
- Added random length and padding params
- Removed all unneeded chars

Now `Security::token` result may contain A-Z, a-z, 0-9, `-` and `_`. `=` is also used if `$padding` is true.

See [RFC 3548](http://tools.ietf.org/html/rfc3548) for the definition of URL-safe base64.

### DIFF
--- a/classes/Kohana/Security.php
+++ b/classes/Kohana/Security.php
@@ -10,7 +10,7 @@
  */
 class Kohana_Security {
 
-	// The length of the random length
+	// The length of the random string
 	const RANDOM_LENGTH = 16;
 
 	/**
@@ -41,7 +41,7 @@ class Kohana_Security {
 	 * If $n is not specified, Secure::RANDOM_LENGTH is assumed. It may be larger in future.
 	 *
 	 * @param   boolean $new     force a new token to be generated?
-	 * @param   integer $n       specifies the length of the random length
+	 * @param   integer $n       specifies the length of the random string
 	 * @param   bool    $padding
 	 * @return  string  The result may contain A-Z, a-z, 0-9, "-" and "_". "=" is also used if $padding is true.
 	 * @uses    Session::instance
--- a/classes/Kohana/Security.php
+++ b/classes/Kohana/Security.php
@@ -10,6 +10,9 @@
  */
 class Kohana_Security {
 
+	// The length of the random length
+	const RANDOM_LENGTH = 16;
+
 	/**
 	 * @var  string  key name used for token storage
 	 */
@@ -18,6 +21,8 @@ class Kohana_Security {
 	/**
 	 * Generate and store a unique token which can be used to help prevent
 	 * [CSRF](http://wikipedia.org/wiki/Cross_Site_Request_Forgery) attacks.
+	 *
+	 * See [RFC 3548](http://tools.ietf.org/html/rfc3548) for the definition of URL-safe base64.
 	 *
 	 *     $token = Security::token();
 	 *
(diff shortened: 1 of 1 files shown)


classes/Kohana/Security.php
```diff
@@ -33,12 +38,15 @@ class Kohana_Security {
 	 *     ));
 	 *
 	 * This provides a basic, but effective, method of preventing CSRF attacks.
+	 * If $n is not specified, Secure::RANDOM_LENGTH is assumed. It may be larger in future.
 	 *
-	 * @param   boolean $new    force a new token to be generated?
-	 * @return  string
+	 * @param   boolean $new     force a new token to be generated?
+	 * @param   integer $n       specifies the length of the random string
+	 * @param   bool    $padding
```
enov, line 45: I am not sure about the importance of this parameter.

classes/Kohana/Security.php
```diff
@@ -33,12 +38,15 @@ class Kohana_Security {
...
+	 * @return  string  The result may contain A-Z, a-z, 0-9, "-" and "_". "=" is also used if $padding is true.
 	 * @uses    Session::instance
 	 */
-	public static function token($new = FALSE)
+	public static function token($new = FALSE, $n = NULL, $padding = FALSE)
```
enov, line 49: Could you please set the default `$n = 16` here? Why did you choose 16 (it was openssl_random_pseudo_bytes(32))?

enov, line 49: No need for Security::RANDOM_LENGTH constant

classes/Kohana/Security.php
```diff
@@ -52,14 +60,17 @@ public static function token($new = FALSE)
 			{
 				// Generate a random pseudo bytes token if openssl_random_pseudo_bytes is available
 				// This is more secure than uniqid, because uniqid relies on microtime, which is predictable
-				$token = base64_encode(openssl_random_pseudo_bytes(32));
+				$token = pack('a*', openssl_random_pseudo_bytes($n ?: Security::RANDOM_LENGTH));
+				$token = str_replace(array("\n", "\r", "\n\r"), '', $token);
```
enov, line 64: Why are you removing those newline characters? Wouldn't `base64_encode` take care of them, make them URL safe?

classes/Kohana/Security.php
```diff
@@ -52,14 +60,17 @@ public static function token($new = FALSE)
...
 			}
 			else
 			{
 				// Otherwise, fall back to a hashed uniqid
-				$token = sha1(uniqid(NULL, TRUE));
+				$token = substr(hash('sha256', uniqid(null, true)), 0, $n ?: Security::RANDOM_LENGTH);
 			}
 
+			$token = $padding ? strtr(base64_encode($token), '+/', '-_') : rtrim(strtr(base64_encode($token), '+/', '-_'), '=');
```
enov, line 72: I don't think giving the choice whether to keep the '=' padding or not is important here.

From the RFC you linked:

>   In some circumstances, the use of padding ("=") in base encoded data
>    is not required nor used.  In the general case, when assumptions on
>    size of transported data cannot be made, padding is required to yield
>    correct decoded data.

I suggest you either keep it or trim it.

classes/Kohana/Security.php
```diff
@@ -83,25 +94,25 @@ public static function check($token)
 	{
 		return Security::slow_equals(Security::token(), $token);
 	}
-
```
enov, line 86: Kindly do not submit formatting changes with this PR. Make them separate PR if those changes are important.

enov:
@sergeyklay Thanks for this. I reviewed and added notes. Thank you for your time.

If you have more time to work on this, I suggest you enhance it further to make token generation decoupled from session and PR anew to 3.4.

Thanks again.

WinterSilence:
**Also, token must be longer than 2 characters**

enov:
bump @sergeyklay . Are you still willing to work on this?

sergeyklay:
Okay. Give me some time

sergeyklay:
Unfortunately I do not have time to do this

PR #593, opened 2014-12-22
Add reference to CONTRIBUTING, reduce chances for out of sync copies

randomecho:
Referring to the main CONTRIBUTING instead of adding copies allows
us to track one and not have to deal with the other versions running
out of sync.

Can't seem to link to the default current branch without
specifying exactly the version, so going to refer to where
it can be found instead of adding in permalink.


lenton:
Excellent, this was much needed.

:+1: for this to be added to every kohana module.

PR #599, opened 2015-02-17
Update Cookie.php :: fix error on quotes

Fanamurov:
``` php
$array['test']['0'] = 'value2';
$array['test']['1'] = 'value2';
Cookie::set('test', json_encode($array));
```

after reload

``` php
Cookie::get('test', NULL); //NULL
```

== or ==

``` php
Cookie::set('test', 'Data with "quotes"');
```

after reload 

``` php
Cookie::get('test', NULL); //NULL
```

Just try write this date on cookie. The data will be recorded in set(), but removed when reading Cookie::get() after the Security::slow_equals() - hashes does not match

//HARDCODE MODE OFF

### DIFF
--- a/classes/Kohana/Cookie.php
+++ b/classes/Kohana/Cookie.php
@@ -74,6 +74,7 @@ class Kohana_Cookie {
 			if (Security::slow_equals(Cookie::salt($key, $value), $hash))
 			{
 				// Cookie signature is valid
+				$value = str_replace('\\', '', $value); //deleting slashes (json, etc)
 				return $value;
 			}
 
@@ -117,7 +118,7 @@ class Kohana_Cookie {
 		}
 
 		// Add the salt to the cookie value
-		$value = Cookie::salt($name, $value).'~'.$value;
+		$value = Cookie::salt($name, addslashes($value)).'~'.$value; //add slashes(json support, etc)
 
 		return static::_setcookie($name, $value, $lifetime, Cookie::$path, Cookie::$domain, Cookie::$secure, Cookie::$httponly);
 	}


enov:
Hey @Fanamurov, I can not seem to be able to reproduce the bug. Both examples are working on my side. I am well receiving the json encoded array as well as the data with quotes on reload.

> Just try write this date on cookie

What do you mean by `date` here?

Fanamurov:
Hey!
Data, not date.

``` php
Cookie::set('test', '{"102750":{"count":"1","cost":"2750","modify":[],"group":"10"}}');
var_dump(Cookie::get('test'));  //NULL
```

Value in cookie before static::delete: 
hash~'{\"102750\":{\"count\":\"1\",\"cost\":\"2750\",\"modify\":[],\"group\":\"10\"}}

This after upgrading to 3.3.3.1. In the cookie data is written to the screening, and to verify the hash is taken without it. Reproduced anywhere. Maybe I somehow crookedly updated ... check it out

Fanamurov:
Cookie::set:121

``` php
$value = Cookie::salt($name, $value).'~'.$value; //Salt counted as a pass a value without slashes
```

After that, the browser save data with slashes (_setcookie), after generate salt

enov:
I wonder if this has to do anything with your `Cookie::$salt`... Could you kindly give us details about your environment (OS, PHP version, etc...)? Also, please test this on a vanilla Kohana installation and see if it is reproducible.

Fanamurov:
Hi! MacOs, MAMP PRO php 5.3 

On php 5.4 all ok! 

Sorry, this is problem on my env.

enov:
Glad that you have solved your problem, but PHP 5.3 is officially supported. Wildly judging from the way you have submitted your patch - as well as the way you solved the problem - it seems to me:
- `magic_quotes_gpc` was `on` on your system
- and, for some reason, Kohana could not [sanitize](https://github.com/kohana/core/blob/3.3/master/classes/Kohana/Core.php#L440) your `$_COOKIE`

Please note that you need to point your PRs to the `3.3/develop` or `3.4/develop` branch for us to be able to merge it.

Thanks.

enov:
Now, it is obvious why Kohana is not sanitizing `$_COOKIE`: #601

PR #605, opened 2015-04-09
New mime for Firefox webapp

saruman:
New mime for those applications that generate manifest.webapp with kohana 3.3 in controllers.


randomecho:
Please target this against the **3.3/develop** branch instead

PR #608, opened 2015-05-14
support for PECL HTTP extension v2

enov:
PECL HTTP extension v2 introduced a new namespaced API that is not compatible with v1 API

### DIFF
--- a/classes/Kohana/HTTP.php
+++ b/classes/Kohana/HTTP.php
@@ -95,7 +95,10 @@ abstract class Kohana_HTTP {
 		if (extension_loaded('http'))
 		{
 			// Use the fast method to parse header string
-			return new HTTP_Header(http_parse_headers($header_string));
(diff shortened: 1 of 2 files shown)


enov:
I understand that this feels patchy. If it is necessary, we can either create a shim class `PECL_HTTP`, or instead transfer the logic into the respective classes/methods (eg: HTTP_Header::parse_from_string, HTTP_Header::create_from_environment) in v.3.4.

enov:
I prefer not to merge my own PRs. But if you're not going to merge, at least review then +1.

Thanks.

acoulton:
@enov sorry that I didn't get a chance to look at this. Given it's a bugfix, I think the way you've done it is fine. Perhaps create an issue for 3.4 to look at whether a cleaner refactoring (a shim class, potentially, or even just a polyfill of some kind?) is possible, but I'm OK with this as-is.

enov:
Thanks for merging @acoulton.

PR #609, opened 2015-05-15
Length

saruman:
Get the length of a number like strlen does for a string.


enov:
Thanks for this @saruman. Note that `strlen` already works with numbers, it also converts them to string beforehand.

``` php
echo strlen(10); // outputs 2
echo PHP_EOL;
echo strlen(50000); // outputs 5
```

http://3v4l.org/YcmXn

PR #610, opened 2015-05-18
Use a dummy controller in RequestTest::test_param

enov:
A cleaner way to test parameters passed by `Route` to `Request`:
Use a dummy controller instead of the empty `try...catch` that
was trying to silence "controller not found" 404 exceptions.

### DIFF
--- a/tests/kohana/RequestTest.php
+++ b/tests/kohana/RequestTest.php
@@ -130,8 +130,10 @@ class Kohana_RequestTest extends Unittest_TestCase
 
 		// We need to execute the request before it has matched a route
 		$response = $request->execute();
+		$controller = new Controller_Kohana_RequestTest_Dummy($request, $response);
 
 		$this->assertSame(200, $response->status());
+		$this->assertSame($controller->get_expected_response(), $response->body());
 		$this->assertArrayHasKey('id', $request->param());
 		$this->assertArrayNotHasKey('foo', $request->param());
 		$this->assertEquals($request->uri(), $uri);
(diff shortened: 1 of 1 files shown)


enov:
@acoulton could you please review this one too?

tests/kohana/RequestTest.php
```diff
@@ -730,9 +722,9 @@ public function test_passes_client_params()
 
 } // End Kohana_RequestTest
 
-class Controller_Kohana_RequestTest_Dummy extends Controller
+class Controller_Kohana_RequestdTest_Dummy extends Controller
```
enov: This is really strange.

Now that I am reviewing this myself, I wonder how the tests passed with the "d" in the middle of `RequestdTest`. I remember I was playing around when I added a letter, in order to check if the routes were picking the controller and the tests were passing/failing properly.

Not sure how it got committed and why the tests are green.

acoulton:
@enov sorry I won't have time to look at this till Thursday. Nothing immediately obvious with the `RequestdTest` issue but I can take a look.

enov:
No worries, @acoulton, don't look at this yet. I think I know what's going
on. I let you know when it is ready for review.
Thanks.

tests/kohana/RequestTest.php
rjd22: @enov I suggest making them red again and than make them green. ;)

enov:
Added failing tests, as per request of @rjd22. It tests if there is a success response but gets `404`.

`Request::execute` never fails with an exception when a controller is not found. Rather, in compliance with the very HMVC nature of Kohana, and because of [HMVC isolation feature](https://kohanaframework.org/3.3/guide/kohana/upgrading#hvmc-isolation)  made available in version 3.3.0, it converts the exception into a response with the specific `404` code.

enov:
@acoulton this is ready for review.

rjd22:
:+1:

acoulton:
@enov great, nice one :)

PR #611, opened 2015-05-19
RouteTest: Request mock needs mocking of `method` method

enov:
In order to avoid errors in newer versions of PHPUnit,
a mock of the Request class should mock its `method` method.
This is to avoid fatals such as:

```
Fatal error: Declaration of Mock_Request_cb18252f::method()
must be compatible with that of Kohana_HTTP_Request::method() in
phpunit/phpunit-mock-objects/src/Framework/MockObject/Generator.php(335)
eval()'d code on line 1
```


JackEllis:
:+1:

enov:
Thanks for reviewing @JackEllis

PR #612, opened 2015-05-21
RequestClientInternalTest: Request mock needs mocking of `method` method

enov:
In order to avoid errors in newer versions of PHPUnit,
a mock of the Request class should mock its `method` method.
This is to avoid fatals such as:

```
Fatal error: Declaration of Mock_Request_cb18252f::method()
must be compatible with that of Kohana_HTTP_Request::method() in
phpunit/phpunit-mock-objects/src/Framework/MockObject/Generator.php(335)
eval()'d code on line 1
```

Related to #611

### DIFF
--- a/tests/kohana/request/client/InternalTest.php
+++ b/tests/kohana/request/client/InternalTest.php
@@ -59,6 +59,11 @@ class Kohana_Request_Client_InternalTest extends Unittest_TestCase
 			->method('response')
 			->will($this->returnValue($this->getMock('Response')));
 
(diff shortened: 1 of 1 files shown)


rjd22:
:+1:

enov:
Thanks @rjd22 for the review, merging.

enov:
This is still not working. I forgot to add 'method' to the second array parameter of `$this->getMock` on L39.

enov:
Fixed [here](https://github.com/kohana/core/commit/01f7cc363c5283297dad018f8717ef9e29cd81e1).

PR #618, opened 2015-06-11
PHPUnit_MAIN_METHOD not available in PHPUnit 4.0+

enov:
Instead `PHPUNIT_COMPOSER_INSTALL` is available when PHPUnit is
installed through composer, and `__PHPUNIT_PHAR__` is available
when PHPUnit is installed by downloading the phar


enov:
@acoulton this is needed in order for PHPUnit not to hang up forever, as it is explained by the few lines of comments above the ones I have edited.

PR #621, opened 2015-06-24
Error, delete method called by get method

zogame:
ErrorException [ 2 ]: Cookie names cannot contain any of the following '=,; \t\r\n\013\014' ~ SYSPATH/classes/Kohana/Cookie.php

### DIFF
--- a/classes/Kohana/Cookie.php
+++ b/classes/Kohana/Cookie.php
@@ -181,7 +181,22 @@ class Kohana_Cookie {
 	 */
 	protected static function _setcookie($name, $value, $expire, $path, $domain, $secure, $httponly)
 	{
-		return setcookie($name, $value, $expire, $path, $domain, $secure, $httponly);
+		$deleted = FALSE;
+		try
+		{
+			// Remove the cookie
+			unset($_COOKIE[$name]);
+			// Nullify the cookie and make it expire
+			setcookie($name, NULL, -86400, Cookie::$path, Cookie::$domain, Cookie::$secure, Cookie::$httponly);
+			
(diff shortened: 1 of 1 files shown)


enov:
Hey @Zogame,

Thanks for this. Could you give us a code to reproduce this? Perhaps you are using invalid characters while getting a cookie? Maybe [this](https://github.com/slimphp/Slim/issues/115) is related? Silencing an exception might not be the right way to solve things. Kindly dig deeper and provide us with better insight/patch.

I will leave this open for the time being. Thanks.

zogame:
Hi @enov ,

A reproduction cord is this.

``` PHP
Cookie::set("foo=,; \t\r\n", 'value');
```

but, I write a right code again.　Thanks.

enov:
@Zogame I see that an exception is legitimate in that case. [RFC 6265](https://tools.ietf.org/html/rfc6265#section-4.1) states that cookie name should be tokens as described in [RFC 2616](https://tools.ietf.org/html/rfc2616#section-2.2):

> ```
>    token          = 1*<any CHAR except CTLs or separators>
>    separators     = "(" | ")" | "<" | ">" | "@"
>                   | "," | ";" | ":" | "\" | <">
>                   | "/" | "[" | "]" | "?" | "="
>                   | "{" | "}" | SP | HT
> ```

`CTL` stands for any US-ASCII control character (octets 0 - 31) and DEL (127).

If you really need to set Cookies with exceptional names, I suggest you base64 encode/decode them at the application level.

zogame:
I get it.
The exception is correct.
Close this. Thanks.

Issue #623, opened 2015-07-03
Arr::map doesn't work with multidimensional array and keys 


acidtv:
The current implementation of Arr::map() does not pass the $keys array on to itself when doing the recursive call. The result is that $callbacks is applied to every element in sub-arrays, not just to the fields specified in $keys.

For example:

``` php
Arr::map('strip_tags', array(array('foo' => '<p>bar</p>', 'bar' => '<p>baz</p>')), 'foo')
```

Returns

```
array
  0 => 
    array
      'foo' => string 'bar' (length=3)
      'bar' => string 'baz' (length=3)
```

The 'bar' element here should have been untouched.

I can issue a pull request and add unit tests for this if someone tells me in which branch this can best be fixed :)

enov:
Hey @acidtv, thanks for this report.

Bug fixes should be targeted to `3.3/develop` whenever possible. Gets usually merged up to `3.4/develop` later.

acidtv:
Ok, I've issued a pull request.

PR #624, opened 2015-07-10
Fixing #623: Arr::map doesn't work with multidimensional array and keys

acidtv:
Arr::map didn't pass the $keys array on to itself when doing the
recursive call. This resulted at $callbacks being run on every element
in multidimensional arrays.


enov:
Excellent @acidtv. Thanks!

PR #625, opened 2015-07-17
Mod: now URL::is_trusted_host uses preg_quote()

Asenar:


enov:
@Asenar thanks for this. But we left the ability to write regex pattern, similar to the way it's done in Symfony.

https://github.com/kohana/core/blob/3.3/master/guide/kohana/upgrading-from-3-3-3-1.md#trusted-hosts

> [!!] Do not forget to escape your dots (.) as these are regex patterns. These patterns should always fully match, as they are prepended with ^ and appended with $.

:open_mouth: just realized that I still have to update the main kohanaframework.org site with the latest release.

Asenar:
indeed, I didn't checked the upgrade notes. Maybe this should also be added in the phpdoc in classes/Kohana/URL.php ?

enov:
The "empty" `url.php` config file has [a reminder](https://github.com/kohana/core/blob/3.3/master/config/url.php#L13) as well as the [List your trusted hosts](https://github.com/kohana/core/blob/3.3/master/guide/kohana/install.md#configure) point in the Configure section of the installation help.

Any doc addition would be welcome :+1:

PR #626, opened 2015-07-17
Mod: now URL::is_trusted_host uses preg_quote()

Asenar:


Asenar:
Cancelled - https://github.com/kohana/core/pull/625

PR #629, opened 2015-08-01
when creating directories, use 0755 permissions

timkuijsten:
It is considered bad practice and often unneeded to have world
writable directories on a production server. Therefore don't create
cache and log directories with mode 0777 or with the setgid bit set.

### DIFF
--- a/classes/Kohana/Core.php
+++ b/classes/Kohana/Core.php
@@ -905,10 +905,10 @@ class Kohana_Core {
 		if ( ! is_dir($dir))
 		{
 			// Create the cache directory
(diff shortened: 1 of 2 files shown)


enov:
Hey @timkuijsten, thanks for this.

While it might be considered a bad practice and often unneeded, it is designed to work on all setups/environments.

Kohana will not create directories and set up the permissions in case those directories exist. So if they are created manually, or through a deploy tool (capistrano, rocketeer, etc...), your permissions can be set according to your needs.

Politely closing this as it should also be targeted to a develop branch `3.3/develop`.

PR #633, opened 2015-08-31
Change detecting urls starting with // in HTML::script, HTML::style and HTML::image + support data-scheme in HTML::image

arkagod:
Support urls starting with // in HTML::script, HTML::style and HTML::image + support data-scheme in HTML::image


rjd22:
@arkagod can you please point this PR to `3.3/develop`? And could you also add tests to https://github.com/kohana/core/blob/3.3/develop/tests/kohana/HTMLTest.php

That way we know what cases you're trying to cover.

PR #637, opened 2015-09-21
Fix Content-Type detection

emirb:


arkagod:
$headers = [];

broken dependency

.travis.yml
language: php
php:
- 5.3

PR #639, opened 2015-09-29
Merge 3.3/develop into 3.4/develop

enov:
I believe there's a circular dependency between the two PRs. This PR needs the changes done in kohana/unittest/3.3/develop, and the related https://github.com/kohana/unittest/pull/50 needs the changes done in kohana/core/3.3/develop, as well.

Could someone have a look at both?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ php:
   - 5.4
   - 5.5
   - 5.6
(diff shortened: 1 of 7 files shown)


rjd22:
@enov @acoulton something is still wrong. The build is still failing.

PR #649, opened 2015-12-16
Call `mkdir` and `file_put_contents` once to improve performance

enov:
- Returns early if there's nothing to write
- Call `mkdir` once with recursive flag on
- Prepare string to write and call `file_put_contents` once

Thanks.


enov:
On Travis CI, composer can't finish updating packages for HHVM, as well as in https://github.com/kohana/image/pull/28 for all runtimes.

Is this related to https://github.com/composer/composer/issues/1314?

acoulton:
@enov when I flushed the travis cache and re-ran the build it was OK. I've started #650 which may help.

PR #650, opened 2015-12-16
Attempt to resolve composer caching issues on travis

acoulton:
As reported in #649, some travis builds are failing to install composer
packages - this appears to be since we added package caching.

https://github.com/travis-ci/travis-ci/issues/4579#issuecomment-138548147
suggests only caching composer's cached package archives, not metadata
and the checked out vendors, which may themselves be out of date.

Also try running with the --no-interaction flag, which should trigger it to fall
back to a git checkout when the travis API token has been rate limited.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,7 @@ matrix:
       env: 'COMPOSER_PHPUNIT="lowest"'
 
 before_script:
+  - composer self-update
   - COMPOSER_ROOT_VERSION=3.3.x-dev composer install --prefer-dist --no-interaction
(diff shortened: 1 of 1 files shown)


acoulton:
Also I notice travis are using a very old composer version (Feb 2015) which may also be an issue - if this still isn't reliable then potentially we should also update composer at the start of the build.

rjd22:
@acoulton It's always a good idea to run composer self-update at the start of a build.

acoulton:
@rjd22 yeah, I normally do - it's a shame that we can't easily cache the updated version though so it has to do it on each build.

I think this PR would fix the broken builds, but mostly because it's not really caching anything now - travis is basically never able to download archive versions (which would be cached) because of the github rate limit so it falls back to cloning (which doesn't seem to be cached).

PR #652, opened 2015-12-17
Add ability to specify mode for created files and directories

enov:
I did not address the problem explained in #629 well.

This PR will allow us to bootstrap the Log_File writer with custom permissions.

cc @timkuijsten

### DIFF
--- a/tests/kohana/Log/FileTest.php
+++ b/tests/kohana/Log/FileTest.php
@@ -168,6 +168,86 @@ EXPECTED
 		$this->assertSame($expected_file_contents, $writer->get_written_logs());
 	}
 
+	/**
+	 * Provider for test_created_files_folders_permissions
+	 */
+	public function provider_created_files_folders_permissions()
+	{
+		return [
+			[0777, 0666, TRUE],
+			[0700, 0600, TRUE],
+			[0000, 0666, FALSE],
+			[0777, 0000, FALSE],
+			[0000, 0000, FALSE],
+		];
+	}
+
+	/**
+	 * Tests that different file/folder modes affect the writers ability to
+	 * write
+	 *
+	 * @test
+	 * @dataProvider provider_created_files_folders_permissions
+	 */
+	public function test_created_files_folders_permissions($dir_mode, $file_mode, $should_write)
+	{
+		if ( ! $should_write)
+		{
+			$this->setExpectedException('PHPUnit_Framework_Error_Warning');
+		}
+
+		$writer = new Kohana_Log_FileTest_Testable_Log_File($this->vfs_root->url(), $dir_mode, $file_mode);
+		$writer->write([$this->dummy_log]);
+
+		$this->assertSame($dir_mode, fileperms($writer->get_directory()) & $dir_mode);
+		$this->assertSame($file_mode, fileperms($writer->get_directory() . $writer->get_filename()) & $file_mode);
+	}
+
+	/**
+	 * Provider for test_writer_does_not_create_or_change_mode_existing_file_folder
+	 */
+	public function provider_writer_does_not_create_or_change_mode_existing_file_folder()
+	{
+		return [
+			[
+				0700, // mode of existing folder
+				0600, // mode of existing file
+				0777, // mode for newly created folders
+				0666, // mode for newly created files
+			],
+		];
+	}
+
+	/**
+	 * Tests that writer does not create a folder when it exists and does not
+	 * change its mode
+	 *
+	 * @test
+	 * @dataProvider provider_writer_does_not_create_or_change_mode_existing_file_folder
+	 */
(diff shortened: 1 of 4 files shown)


timkuijsten:
cool! I see you dropped the default setgid bit as well.

rjd22:
:+1: you can merge this.

acoulton:
I don't mean to be a nag, but it would be great to add tests for this before it's merged...

enov:
@acoulton alright. Do you think I should add anything to the docs?

acoulton:
It never hurts :) especially since we're probably for security recommending
people make the permissions stricter than the defaults?

On 18 December 2015 at 10:08, Samuel Demirdjian notifications@github.com
wrote:

> @acoulton https://github.com/acoulton alright. Do you think I should
> add anything to the docs?
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/kohana/core/pull/652#issuecomment-165737109.

## 

Andrew Coulton
Founder
inGenerator Ltd

Follow us on Twitter @inGenerator
Phone us on 0131 510 0271

inGenerator Ltd is a company registered in Scotland (no SC435161) with its
registered office at 8 Craigleith Hill Row, Edinburgh EH4 2JX.

enov:
Tests and docs added.

tests/kohana/Log/FileTest.php
```diff
@@ -168,6 +168,39 @@ public function test_writer_does_not_create_existing_file_but_only_appends(array
...
 	}
 
+	/**
+	 * Provider for test_created_files_folders_permissions
+	 */
+	public function provider_created_files_folders_permissions()
+	{
+		return [
+			[0777, 0666, TRUE],
+			[0700, 0600, TRUE],
```
enov, line 178: I am not sure about this test case. I wonder if it fails when PHPUnit is run with different user in different CI environments?

acoulton, line 178: I think maybe it's better to explicitly assert the mode of the created directory/file after the logs have been written? Possibly also assert that it doesn't modify the attributes of an already-existing directory even if the dir_mode is different?

I think there shouldn't be issues between CI environments/users/etc because the virtual filesystem should behave the same everywhere, it's not tied to the OS environment in any way.

timkuijsten:
ps. I would still vote for a default permission setting that makes the log file not world readable and (or at least) not world writable.

Otherwise the recommendation to change the default should maybe stress a bit more that the default settings are for development only and not for production.

timkuijsten:
Aside from my suggestions I think the most prominent case is that if a site gets hacked, the attacker can completely overwrite each log since the code has full access. That won't be fixed with a strict 700 or 600 mode either.

enov:
@timkuijsten

I went with `0750` and `0640` as defaults. I think `0700` and `0600` might be too restrictive.

@acoulton @rjd22 please review again and let me know what do you think.

enov:
Please note that [Monolog creates directories with `0777`](https://github.com/Seldaek/monolog/blob/master/src/Monolog/Handler/StreamHandler.php#L138)

Also, [`mkdir` has its default `0777`](http://php.net/manual/en/function.mkdir.php).

acoulton:
I'm OK with those defaults, I don't think there's a need to lock them down to the individual user level. 

The biggest issues from a security point of view I think would be a compromised web process either reading sensitive data from the logs, or overwriting them to mask details of an attack. There's no effective way to avoid either of those with a simple file-based logger, since by definition as @timkuijsten  says the web process has to be able to write (and therefore read) that file. The only really secure way to handle it is an inter-process / remote logging solution eg syslog / a log monitoring agent / etc so that it can appear to the application as an append-only black hole.

So these defaults are better but shouldn't give too much of a sense of security.

timkuijsten:
Op 22-12-15 om 18:35 schreef Andrew Coulton:

> There's no effective way to avoid either of those with a simple
> file-based logger

Besides an external log server there is the sappnd flag, see chflags(1). 
But this doesn't invalidate the point you make :)

timkuijsten:
Op 22-12-15 om 12:04 schreef Samuel Demirdjian:

> Also, |mkdir| has its default |0777|.

It should be noted that these permissions are susceptible to the 
processes umask so in practice most directories are created with 0022 
cleared.

enov:
> > Also, |mkdir| has its default |0777|.
> 
> It should be noted that these permissions are susceptible to the 
> processes umask so in practice most directories are created with 0022 
> cleared.

:+1: I should have noted that. So `umask` affects Monolog's `createDir` as well.

enov:
PR needs some love :heart:

acoulton:
@enov sorry about the delay, thanks

enov:
No problem @acoulton. Thanks! Thank you @timkuijsten for the observations and suggestions.

timkuijsten:
nice one!

PR #657, opened 2016-01-08
More log writer filters

enov:
Hi,

These are the additional filters to log writers coming from #563.

Do you think it's better to have them as examples in the docs?

Thanks :smile:


neo22s:
sae needs to be moved to 4.0.0

PR #653, opened 2015-12-22
Set the content-length in case request body be changed after $request->render().

sunchuo:
$request = Request::factory('http://www.google.com/')->method(Request::POST)->post(['aaa' => 'bbb']);
echo $request; // request->render() set the content-length as 0.
$response = $request->execute(); //request error

PR #661, opened 2016-01-22
Bugfix when method Request::render() break instance of request object

pbardov:
Fix of this bug:
$request = Request::factory('http://httpbin.org/headers')->method(Request::GET)->post(array('foo' => 'bar', 'bar' => 'baz'));
Log::instance()->add(Log::DEBUG, **METHOD** . " request = \n" . $request->render() . "\n");
//in log file Content-Length: 0
$resp = $request->execute();
$this->response->body($resp->body());
// http://httpbin.org/headers say that Content-Length: 0, $request object is broken after render() call

P.S. Sorry for my english

### DIFF
--- a/classes/Kohana/Request.php
+++ b/classes/Kohana/Request.php
@@ -709,6 +709,16 @@ class Kohana_Request implements HTTP_Request {
 		}
 	}
 
+    /**
+	 * Sets some parameters of cloning object
+	 */
+	public function __clone()
+	{
+		$this->_route  = isset($this->_route)  ? clone $this->_route  : null;
+		$this->_header = isset($this->_header) ? clone $this->_header : null;
+		$this->_client = isset($this->_client) ? clone $this->_client : null;
+	}
+
 	/**
 	 * Returns the response as the string representation of a request.
 	 *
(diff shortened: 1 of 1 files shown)


enov:
Thanks @pbardov! I think this was fixed in #653 and #655. It should be available in the next maintenance release. Could you confirm if the #655 fix works for you?

pbardov:
This bug was not fixed in #653 and #655 . Bug stil present. Example (Request.php with fix #653 and #655):

```
public function action_index()
{
    $answer              = array('render' => '', 'firstRun' => '', 'secondRun' => '');
    $request             = Request::factory('http://httpbin.org/headers')->method(Request::GET)->post(array('foo' => 'bar', 'bar' => 'baz'));
    $answer['render']    = $request->render();
    $resp1               = $request->execute();
    $answer['firstRun']  = $resp1->body();
    $request->post(array('one' => 'one', 'two' => 'two', 'three' => 'three'));
    $resp2               = $request->execute();
    $answer['secondRun'] = $resp2->body();
    $this->response->headers('Content-Type', 'text/plain');
    $this->response->body(print_r($answer, true));
}
```

output:

Array
(
    [render] => GET http://httpbin.org/headers HTTP/1.1
Content-Type: application/x-www-form-urlencoded; charset=utf-8
Content-Length: 15 // OK, all right

foo=bar&bar=baz
    [firstRun] => {
  "headers": {
    "Accept": "_/_", 
    "Content-Length": "15", // OK, all right
    "Content-Type": "application/x-www-form-urlencoded; charset=utf-8", 
    "Host": "httpbin.org"
  }
}

```
[secondRun] => {
```

  "headers": {
    "Accept": "_/_", 
    "Content-Length": "15", // WTF? bug here!!!
    "Content-Type": "application/x-www-form-urlencoded; charset=utf-8", 
    "Host": "httpbin.org"
  }
}

)

This is because $request-render() modify object and set content-length header. Render method must be constant method!!!

enov:
@pbardov ah, I'll check out again very soon. Note that I had to close the PR as it was towards `3.3/master`. I can not merge directly into `3.3/master` as per the [contributing guidelines](https://github.com/kohana/kohana/blob/3.3/master/CONTRIBUTING.md). The branch `3.3/develop` should have probably be the right branch.

Cheers!

PR #662, opened 2016-01-22
Bugfix Request::render() break instance of request object

### LINKED ISSUE #651: Use of mikey179/vfsStream for Log tests breaks module builds
#563 introduced a composer `require-dev` for mikey179/vfsStream to allow virtual filesystem tests of the log writers.

Unfortunately this breaks the builds for the other modules (eg https://travis-ci.org/kohana/image/jobs/97191751) because those builds include all the kohana/core tests (to check the module hasn't broken anything in core, I guess) but don't pick up the `require-dev` dependency from this package. Similarly, the kohana/kohana build is also going to fail for the same reason.

I think the best solution is to `require` mikey179/vfsStream in kohana/unittest - the same way we handle phpunit, so it's always present in development environments for any module that has a `require-dev` on kohana/unittest.

Alternatively we could modify the test suite to check if vfsStream is present and skip those tests if not.

Thoughts?

pbardov:
Bug fix #653 #655 #661 . Create new pull request with right branch. See discussion in #661 pull request.

### DIFF
--- a/classes/Kohana/Request.php
+++ b/classes/Kohana/Request.php
@@ -709,7 +709,17 @@ class Kohana_Request implements HTTP_Request {
 		}
 	}
 
-	/**
+    /**
+     * Sets some parameters of cloning object
+     */
+    public function __clone()
+    {
+        $this->_route  = isset($this->_route) ? clone $this->_route : null;
+        $this->_header = isset($this->_header) ? clone $this->_header : null;
+        $this->_client = isset($this->_client) ? clone $this->_client : null;
+    }
+
+    /**
 	 * Returns the response as the string representation of a request.
 	 *
 	 *     echo $request;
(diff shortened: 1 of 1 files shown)


enov:
@pbardov I like the idea to clone the request before rendering. I think that the original solution submitted by @sunchuo in #653 might fix this. Note that [we still use tabs for indentation](http://kohanaframework.org/3.3/guide/kohana/conventions#indentation), so we might kindly ask you to fix the indentation if we want to merge your proposed solution. Thanks again for your time, and sorry for being slow to respond.

cc @acoulton

enov:
@pbardov I had some time to look at this again, and tested the [original solution](https://github.com/kohana/core/commit/e1664aaae027728276460900cddd36fa33796ec0) provided by @sunchuo in #653, that is, to add the following code in order to reset the `content-length` in `Request_Client_External` on line 129:

``` php
$request->headers('content-length', (string) $request->content_length());
```

Let me know if this fix works for you. Thank you for your patience.

acoulton:
I'm not sure I completely understand the issue, sorry.

But I think for a bugfix to the 3.3/develop series, explicitly fixing the content-length to resolve this specific issue (if possible) is better than cloning the request which might introduce other behavioural changes especially for people who've extended/mocked the request in some way.

PR #663, opened 2016-01-29
Adding namespaces to Kohana

rjd22:
Problems:
- Could not namespace `Kohana_HTTP_Exception_300` and all other numbers because classnames can't only contain numbers


birkir:
Good stuff here.

classes/Kohana/Arr.php
```diff
@@ -8,7 +16,7 @@
  * @copyright  (c) 2007-2014 Kohana Team
  * @license    http://kohanaframework.org/license
  */
-abstract class Kohana_Arr {
+class Arr {
```
acoulton, line 19: Did you intentionally drop `abstract` for the `Arr` class?

acoulton:
@rjd22 thanks for this, makes sense to me. Are you intentionally leaving the extension classes (eg `\Cookie`, `\Arr` etc) in the root namespace for compatibility, or were you planning to also rename those once you'd done the core base classes?

classes/Kohana/Arr.php
rjd22, line 19: Yes else you won't be able to use the name spaced version directly.

rjd22:
@acoulton I'm still unsure. Maybe I will change them to the namespaced versions but at this moment leaving them there keep the tests working and making the refactoring easier.

classes/Kohana/Arr.php
acoulton, line 19: There are no instance methods in `\Kohana\Core\Arr` though, so it doesn't make any difference?

acoulton, line 19: @rjd22 although I guess it feels neater without it, and in keeping with the other static-method-helper classes which I now see were already not defined as abstract, so :+1: from me. Sorry.

acoulton:
Maybe we should leave the extension classes in the root namespace until we decide exactly what's happening about transparent extension and how people should migrate away from that...

neo22s:
ok, this should go to 4.0.0, no?

acoulton:
Depends on @rjd22 I guess. If it's close to completion then yes, if it needs more time we should get 4.0 out and put this in 5.0.

rjd22:
@acoulton It depends on when you want to release 4.0. It's almost done but I have some trouble with class collision's that I need to figure out. I'm also thinking of removing the extension classes because they make it really confusing for the user.

neo22s:
If this is going to make migration from 3.3.x to 4.x harder I vote for 5.x
;)

acoulton:
:+1: for 5.x, both to split the migration effort for users and because I think we need time to think properly about the implications of dropping extension classes and how best to replace transparent extension (which we should do, but carefully)

rjd22:
@acoulton @neo22s I agree with releasing it with 5.x also. I will give me some breathing room to fix this properly.

rjd22:
@acoulton @enov I could only remove the Cascading file system and leave the classes. This way we can include it into 4.0.

I'm having troubles with the HTTP exceptions. Would it be an idea to rename this to exception similar to symfony?

`Kohana_HTTP_Exception_400` becomes `HttpBadRequestException` based on https://en.wikipedia.org/wiki/List_of_HTTP_status_codes

Another problem I'm having is that I have collisions between `class Request` and `interface Request` shall I rename those to `interface RequestInterface` for now?

This problem is the same for `interface Response` and more.

neo22s:
@rjd22 maybe is better we create an issue instead of talking on the PR? so namespaces wont be  in 4.0.0

acoulton:
@rjd22 

> I could only remove the Cascading file system and leave the classes.

I'm not quite sure what you're suggesting?

> `Kohana_HTTP_Exception_400` becomes `HttpBadRequestException` 

that makes sense to me

> collisions between class `Request` and interface `Request` shall I rename those to `interface RequestInterface`

personally I prefer the generic name (`Request`) for the interface (as the most generic case) and then naming the class to say give some information about the implementation. class `KohanaRequest`, `AbstractRequest`, `WebRequest`, `InternalRequest` or whatever.

This also means existing typehints for `Request` don't have to change and automatically become typehints for the interface rather than any particular implementation.

rjd22:
@acoulton The cascading file system won't work anymore because we will only use the namespaced instances in the framework, but you can still use the extending classes to get an instance of the namespaced classes. This way we can deprecate the extending classes in version 4.0 without completely breaking people their applications, and help them move over to the namespaced classes for 5.0.

PSR uses the `RequestInterface` and `ResponseInterface` route. My idea was that it would be smart to follow how they do things to ease transition later.

acoulton:
@rjd22 

> The cascading file system won't work anymore because we will only use the namespaced instances in the framework,

That has the potential to be a big breaking change for some end-users, and I think needs to be a release of its own. People may be relying on Kohana picking up their custom `Request`, `Route` or whatever class - I don't think we can remove that until there is a way (facade, DI container, whatever) to inject custom implementations into the framework.

> PSR uses the `RequestInterface` and `ResponseInterface` route

Fair enough, makes sense. I think they're wrong, but better to be consistent.

rjd22:
@acoulton @enov @neo22s  I've done a lot of work on this PR but am starting to feel burned out. Maybe someone else would be willing to help out or take over?

There are some parts that are hard to namespace:
- The Http Error Code classes.
- The UTF8 class function combi

Some parts in the internal classes still use the extending classes. These need to be replaced with internal classes.

neo22s:
hello @rjd, I think we should stop this if we have planned it for 5.0.0 and then retake it once we want to release 5.0.0 ?

rjd22:
@neo22s I think it's better to merger this into the 5.0 branch and reitterate on it. Else the work will go stale.

neo22s:
@rjd22 I understand what you mean, but we base it on 4.0.0 then?

rjd22:
Yeah, since we want people to be able to move over.

rjd22:
@acoulton @neo22s @enov I quit working on Kohana. So if someone wants to take over this branch feel free to do so but I'm going to close this PR

neo22s:
@rjd22 I am not starting any new project on kohana either... I have moved on. thanks for your time. Seems everytime less of us are interested on kohana. thanks to all

PR #664, opened 2016-02-01
Fix uri with falsey strings eg '/0' [fixes kohana/kohana#94]

acoulton:
Calling `(new Request('/0'))->uri()` should return '0' but
was previously returning '/' because the '0' was incorrectly
treated as empty.

Note that this is in part because `Request::_construct`
always trims the leading '/' from internal requests - the
previous comment implies the original intention was just to
remove trailing '/' so this may not actually be desired but
has been standard Kohana behaviour for a long time.

Thanks to @sergiozia for the report and proposed patch in kohana/kohana#94

Amended the comment to clarify.


enov:
Thanks @acoulton, @sergiozia.

PR #665, opened 2016-02-05
Reset request `content-length` header during request execution

enov:
Related to #653 #655 #661 #662

`Request::render` sets `content-length` header. This creates problems on subsequent changes to the request object.

cc @sunchuo, @pbardov, @acoulton

### DIFF
--- a/tests/kohana/RequestTest.php
+++ b/tests/kohana/RequestTest.php
@@ -736,9 +736,47 @@ class Kohana_RequestTest extends Unittest_TestCase
 		$this->assertEquals($client->strict_redirect(), FALSE);
 	}
 
+	/**
+	 * Tests correctness request content-length header after calling render
+	 */
+	public function test_content_length_after_render()
+	{
+		$request = Request::factory('https://example.org/post')
+			->client(new Kohana_RequestTest_Header_Spying_Request_Client_External)
+			->method(Request::POST)
+			->post(array('aaa' => 'bbb'));
 
+		$request->render();
+
+		$request->execute();
+
+		$headers = $request->client()->get_received_request_headers();
+
+		$this->assertEquals(strlen($request->body()), $headers['content-length']);
+	}
 
 } // End Kohana_RequestTest
+
+/**
+ * A dummy Request_Client_External implementation, that spies on the headers
+ * of the request
+ */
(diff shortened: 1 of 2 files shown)


rjd22:
@enov how do this exactly solve the problem?

enov:
@rjd22 sorry for replying late.

`content-length` header is set when `Request::render` is called. When changing post data, the old header is sent. Note that the test in https://github.com/kohana/core/commit/c9a75faf0f3a3f70a16269cc004293e02bf71ef6 fails if the fix in https://github.com/kohana/core/commit/e1589612dae760e278ae1a9ecf557d0bd9c44bba is not applied. The first test fails when #655 is not applied.

Thanks.

rjd22:
:+1:

tests/kohana/RequestTest.php
```diff
@@ -736,10 +736,60 @@ public function test_passes_client_params()
...
+		$request = Request::factory('https://example.org/post')
+			->client(new Request_Client_Content_Length_Kohana_RequestTest)
+			->method(Request::POST)
+			->post(array('aaa' => 'bbb'));
+
+		$request->render();
+
+		$response = $request->execute();
+
+		$this->assertEquals(strlen($request->body()), $response->body());
```
acoulton: @enov I found this assertion a bit hard to understand at first, without going to look at the implementation of `Request_Client_Content_Length_Kohana_RequestTest`. It might become hard to maintain once the context of the commit/PR is not available.

It might be better to use a `Kohana_RequestTest_Headers_Echoing_Request_Client` and an assertion like `$this->assertEquals(strlen($request->body()), json_decode($response->body(), TRUE)['content-length'])`.

Or better yet a `Kohana_RequestTest_Header_Spying_Request_Client_External` and an assertion like `$this->assertEquals(strlen($request->body()), $client->getReceivedRequestHeaders('content-length'))` ?

Would also make that mock more reusable for other tests in future.

enov: @acoulton I was reviewing this PR and I just found your note here! It had slipped away from my attention all this time, I am sorry.

enov: So basically, you're suggesting to rename that class and slightly tweak it?

enov:
@acoulton PR updated as suggested. Thanks.

acoulton:
@enov looks good, thanks

PR #667, opened 2016-02-13
Change detecting urls starting with //

arkagod:
Change detecting urls starting with // in Form::open, HTML::script, HTML::style and HTML::image + support data-scheme in HTML::image

### DIFF
--- a/classes/Kohana/Form.php
+++ b/classes/Kohana/Form.php
@@ -45,7 +45,7 @@ class Kohana_Form {
 			// Allow empty form actions (submits back to the current url).
 			$action = '';
 		}
-		elseif (strpos($action, '://') === FALSE)
+		elseif (strpos($action, '://') === FALSE AND strncmp($action, '//', 2))
 		{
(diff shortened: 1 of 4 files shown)


enov:
Sorry for the late reply @arkagod.

I think this is good and can go to `3.3/develop`.

Regarding `array()` => `[]` changes, you need to do that in separate PR, targeting `3.4/develop`.

cc @acoulton @rjd22

rjd22:
@enov I think we can merge this. It doesn't have to be in 3.3 :)

enov:
Yeah, it doesn't have to be in 3.3... merging!

Thanks @arkagod, @rjd22.

WinterSilence:
it's not all (=  https://github.com/kohana/core/blob/3.4/develop/classes/Kohana/Request.php#L682
`if ( ! $allow_external OR strpos($uri, '://') === FALSE)`, need run search in files for finds all strings with `://`

PR #675, opened 2016-04-13
MIME type "manifest" added

saruman:


rjd22:
@saruman could you please point this patch to `3.3/develop` since we don't merge in master. We only use master for releases.

Closing this so you can make a new PR.

PR #679, opened 2016-04-20
MIME type "manifest" added.

saruman:


rjd22:
:+1: @enov do you also approve?

enov:
Thank you for your time @saruman.

@rjd22 the spec seems to prefer `.appcache` instead of `.manifest`:
https://html.spec.whatwg.org/multipage/browsers.html#offline

See also:
https://developer.mozilla.org/en-US/docs/Web/HTML/Using_the_application_cache
http://stackoverflow.com/questions/7118389/load-cache-manifest-file
http://www.html5rocks.com/en/tutorials/appcache/beginner/

rjd22:
@enov I think in this case @saruman is trying to fix a problem that he is having. Adding the spec MIME types would be okay outside this PR when someone else needs those, no?

enov:
> Adding the spec MIME types would be okay outside this PR when someone else needs those, no?

You mean if the user adds it to `config/mimes.php` in the `application` folder? Yea, if I understood your question, that would be OK.

saruman:
This will be available in 3.4?

enov:
> This will be available in 3.4?

This will be available in 3.3 if #682 is merged, but using `.appcache` file extension. If you need to use `.manifest` extension, you need to extend the config list of MIME types in your application directory as @rjd22 suggested.

Thank you @saruman @rjd22

rjd22:
@enov would it be wrong to add both? Maybe just add both for the sake of being complete.

enov:
@rjd22 not sure if that's a good idea. They have specifically changed the suggested extension for reasons related to Microsoft's use of the extension, see below:

https://github.com/whatwg/html/commit/b3e6c2404739a6850afe389c487175f7e09a605c

The only trusted source that I could find, that still suggest `.manifest` extension is Safari's documentation:

https://developer.apple.com/library/iad/documentation/iPhone/Conceptual/SafariJSDatabaseGuide/OfflineApplicationCache/OfflineApplicationCache.html#//apple_ref/doc/uid/TP40007256-CH7-SW2

It says:

> For most web servers, the filename must end in .manifest.

But that does not mean Safari, the browser itself, requires `.manifest` extension.

PR #681, opened 2016-04-21
Correct comments of the "action" method.

muffetlab:
Is the comment "Action to execute the controller from" correct?


classes/Kohana/Request.php
```diff
@@ -866,7 +866,7 @@ public function controller($controller = NULL)
 	/**
 	 * Sets and gets the action for the controller.
 	 *
-	 * @param   string   $action  Action to execute the controller from
+	 * @param   string   $action  Action to execute the controller of
```
acoulton, line 869: should probably actually be `Action of the controller to execute`

PR #685, opened 2016-04-22
Current Protocol (Helper)

saruman:
Hello.
I have this method to get the current protocol in every project, some with http,other with https.
I made this method to help me identify better which protocol is present in my project so I can use properly http or https in my links.  I don't know if there exists something that do the same thing, but I searched and found nothing.

Hope this helps or maybe if you don't want it in main class you can put it in a helper class like I did.  I extended Kohana_HTTP.


WinterSilence:
Why not use `Request::current()` ?

PR #689, opened 2016-05-28
Make tests independent from OS and application locale

enov:
Some tests are written with specific locale in mind.

In https://github.com/kohana/core/commit/0eb2898b234d2201c7df0d6cb5dbbb719cf990ba, most-probably by mistake, we have removed a line that backups the original locale, before setting the locale used in the tests. Not being to be able to restore the locale, and instead setting the OS default locale, the `NumTest` is affecting other subsequent tests.

Similarly in `DateTest`, tests are written with specific locale in mind. When changing the application default locale in `bootstrap.php` to a different locale, `DateTest` tests fail.

Related to https://github.com/kohana/kohana/issues/106

cc @piotrgolasz, @acoulton, @rjd22

### DIFF
--- a/tests/kohana/DateTest.php
+++ b/tests/kohana/DateTest.php
@@ -17,6 +17,7 @@
 class Kohana_DateTest extends Unittest_TestCase
 {
 	protected $_original_timezone = NULL;
+	protected $default_locale;
 
 	/**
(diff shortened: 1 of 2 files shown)


acoulton:
Great, thanks @enov

PR #691, opened 2016-07-07
Don't use short notatnion arg_separator

bato3:
When url is `?foo=&not_foo=` browsers interpret it as one value `?foo=&not;_foo=` --> `?foo=¬_foo=`

### DIFF
--- a/classes/Kohana/URL.php
+++ b/classes/Kohana/URL.php
@@ -195,7 +195,7 @@ class Kohana_URL {
 		}
 
 		// Note: http_build_query returns an empty string for a params array with only NULL values
-		$query = http_build_query($params, '', '&');
+		$query = http_build_query($params, '', '&amp;');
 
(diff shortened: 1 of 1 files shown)


acoulton:
Hi @bato3, thanks for the pull request. Unfortunately this isn't correct - the URL should originally be returned with the unescaped `&` separator.

Escaping the `&` to `&amp;` is only sometimes appropriate, depending on where you're using the generated URL. For example, you wouldn't want to escape it in a JSON response, a plain-text email, or a `Location:` redirect header.

You do need to encode URLs in `href` (or any other) HTML attributes - but in that case there are other characters that may need to be escaped, not just the `&`. You should use `HTML::chars`, `HTML::anchor` or `HTML::attributes` (or another suitable method) to escape your link attributes for display.

If I've misunderstood the problem you're having, feel free to reopen with more details.

PR #693, opened 2016-07-21
Always parse query string, on the main request or on sub-requests

enov:
Initial request will have the global $_GET re-applied in the `factory()`
method.

Related to https://github.com/kohana/core/pull/627#issuecomment-234012570

cc @hkdobrev 

@acoulton could you please advise how to add a test case for this.

### DIFF
--- a/classes/Kohana/Request.php
+++ b/classes/Kohana/Request.php
@@ -667,13 +667,9 @@ class Kohana_Request implements HTTP_Request {
 		$split_uri = explode('?', $uri);
 		$uri = array_shift($split_uri);
 
-		// Initial request has global $_GET already applied
-		if (Request::$initial === NULL)
+		if ($split_uri)
 		{
-			if ($split_uri)
-			{
-				parse_str($split_uri[0], $this->_get);
-			}
+			parse_str($split_uri[0], $this->_get);
 		}
 
 		// Detect protocol (if present)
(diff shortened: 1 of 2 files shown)


hkdobrev:
Thanks for the swift fix! It looks good to me! I could probably try to work on a test for that a bit later today and if I succeed will PR your branch.

hkdobrev:
@enov I've added a test for your changes in https://github.com/kohana/core/pull/694

acoulton:
@hkdobrev @enov thanks for this, looks good to me. @enov I'll let you double-check you're happy with the test case, if so go ahead and merge this.

enov:
Thank you @hkdobrev, @acoulton. I reviewed the test case and it's good. I also run the test without the patch and it's failing as expected. Keep up the good work @hkdobrev!

hkdobrev:
@enov @acoulton Do you think this could be released as 3.3.6 or something like that? I'm not sure why you need to keep versions of core modules synchronised with core.

enov:
@acoulton @hkdobrev 3.3.6 is released, please test and let me know if I missed anything.

cc @rjd22 @neo22s

neo22s:
hello,

Thanks to all, what are the main differences with 3.3.5? EDIT here it is sorry xD https://github.com/kohana/kohana/releases

I will test later, regards

acoulton:
Thanks @enov

neo22s:
@enov the release is missing, the packed version with the core in it I mean.

Also the modules, I have used the ones packed for 3.3.5.

I have updated the release and added the ZIP file. I hope I have done this correctly.

neo22s:
ok,have tested and all seems ok ;)

enov:
> I have updated the release and added the ZIP file.

Thanks @neo22s

PR #694, opened 2016-07-23
Test query params are parsed correctly in both initial and subrequests

hkdobrev:
A test for #693 failing on `v3.3.5` and passing on `3.3/bugfix/always-parse-get`.


acoulton:
@hkdobrev thanks for this, I've merged into #693 so we can see the build.

Issue #686, opened 2016-04-26
[Security] Encrypt


piotrbaczek:
According to: https://paragonie.com/blog/2015/05/if-you-re-typing-word-mcrypt-into-your-code-you-re-doing-it-wrong mcrypt is no longer a secure encryption mechanism.

Therefore I would suggest swapping from mcrypt to openssl, which is now considered a Core Internet Infastructure.

I would therefore suggest applying this patch: https://github.com/piotrgolasz/core/commit/b6769b4a1c49ac554ef7ffcfde48e715b22eedfb

This uses openssl in two modes: aes-128-cbc and aes-256-cbc, using the Encrypt-then-MAC scheme. It also allows using RSA encryption using Encrypt-then-Sign scheme. This exact method is used in Laravel framework.

Before, the encryption was not hmac-ed, meaning that crypto originating from it could be susceptible to padding oracle attacks. With this patch, first the check hash($ciphertext,$key,$hash) == $hmac is made, so that if the hashes don't match, we already know that this message is not valid, and therefore we don't have to decrypt.

rjd22:
@piotrgolasz why was this issue closed? Was it fixed?

piotrbaczek:
I was going to make changes on a separate branch including changing phpunit tests.

piotrbaczek:
Ok, so the main point of the change is to drop mcrypt in favour of openssl which is now well-tested and globally maintained software. The second reason for change is to use Encrypt-then-mac encryption, so that before we even try to decrypt anything we can validate the hmac, and therefore we won't accept anything that doesn't come from us.

Let me know if you have any suggestions.

piotrbaczek:
Ok, so I have a test example to prove this.

``
public function action_test()
    {
        $this->auto_render = false;

```
    $encrypt = Encrypt::instance();

    $crypto = $encrypt->instance()->encode(serialize(['user' => 'asdf', 'admin' => true]));

    $list = array(0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 'a', 'b', 'c', 'd', 'e', 'f');

    $get = unpack('H*', $crypto);

    for ($i = 0; $i < count($list); $i++)
    {
        for ($j = 0; $j < count($list); $j++)
        {
            //get the number
            $g2 = $get['1'];
            //subsitute next-to last position of IV
            $g2['31'] = $list[$i];
            //substitute last position of IV
            $g2['32'] = $list[$j];

            //echo the combination of bits that we're flipping
[...]

PR #698, opened 2016-08-17
[WIP] Use external libraries for encrypt

enov:
This is a work-in-progress PR to remove crypto ops from the Kohana `Encrypt` class and instead interface it with external libraries.

Related to https://github.com/kohana/core/issues/686

Changes so far:
- Suggest `defuse/php-encryption`
- Kept a simple API with almost full BC, ex: `Encrypt::instance()->encrypt($plaintext);`
- Methods `encode/decode` _became_ `encrypt/decrypt` but this can be reverted back
- Uses config based initialization.
- Drivers should implement `Kohana_Crypto` interface and should name their classes in the format `Encrypt_Drivername`
- Old mcrypt based `Encrypt` is still there, marked as deprecated, named `Encrypt_Legacy` to facilitate migration of data

This is still work in progress, but I wish I'd have a feedback sooner to adjust accordingly. I'd rebase/squash some commits when done.

Thank you for reviewing.

Cheers!


acoulton:
I'll have a proper look tomorrow/friday but from a quick skim this looks reasonable, thanks for the work.

I'd keep the method names the same - we're already pushing a chunk of breaking change and I don't think there's a huge benefit to renaming.

I would suggest removing the ability to encrypt from Encrypt_Legacy altogether. It could either not have the `Kohana_Crypto` interface, or could stay as is but throw from encrypt. We really just want to provide a way for the user to decrypt their existing secrets and re-encrypt them with the new driver.

End users who really don't care about security can always reimplement an `Encrypt_Legacy` pretty easily with the code from the 3.3 branch so I don't think we should make it easy for them.

enov:
Thanks @acoulton for the quick review. I will resume work once you have a deeper look into this.

> I'd keep the method names the same - we're already pushing a chunk of breaking change and I don't think there's a huge benefit to renaming.

Alright.

> I would suggest removing the ability to encrypt from Encrypt_Legacy altogether. It could either not have the Kohana_Crypto interface, or could stay as is but throw from encrypt. We really just want to provide a way for the user to decrypt their existing secrets and re-encrypt them with the new driver.

I suggest that we take a softer approach by raising an error within `Encrypt_Legacy::encode()`:

``` php
trigger_error('Legacy driver is not considered secure anymore. Use Defuse driver instead.', E_USER_WARNING);
```

Users need to silence it manually with `error_reporting(E_ALL & ~E_USER_WARNING);` in case they don't care about security.

Thanks.

PR #699, opened 2016-08-29
Patch Valid::ip to work with PHP v7.0.10+ AND v5.6.25+

enov:
In a breaking change, https://github.com/php/php-src/pull/1954 added `FILTER_FLAG_NO_PRIV_RANGE` to `FILTER_FLAG_NO_RES_RANGE`.

The patch looks awful. Let me know if there is any better way to fix this.

Cheers!

### DIFF
--- a/classes/Kohana/Valid.php
+++ b/classes/Kohana/Valid.php
@@ -224,16 +224,27 @@ class Kohana_Valid {
 	 */
 	public static function ip($ip, $allow_private = TRUE)
 	{
-		// Do not allow reserved addresses
-		$flags = FILTER_FLAG_NO_RES_RANGE;
 
-		if ($allow_private === FALSE)
+		// In PHP v7.0.10+ AND v5.6.25+ FILTER_FLAG_NO_PRIV_RANGE ∈ FILTER_FLAG_NO_RES_RANGE
+		// However we have to combine both flags to support older versions
+		$flags = FILTER_FLAG_NO_RES_RANGE | FILTER_FLAG_NO_PRIV_RANGE;
+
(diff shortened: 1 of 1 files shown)


classes/Kohana/Valid.php
```diff
@@ -224,16 +224,27 @@ public static function url($url)
...
-		$flags = FILTER_FLAG_NO_RES_RANGE;
 
-		if ($allow_private === FALSE)
+		// In PHP v7.0.10+ AND v5.6.25+ FILTER_FLAG_NO_PRIV_RANGE ∈ FILTER_FLAG_NO_RES_RANGE
+		// However we have to combine both flags to support older versions
+		$flags = FILTER_FLAG_NO_RES_RANGE | FILTER_FLAG_NO_PRIV_RANGE;
+
+		$is_valid_public_ip = (bool) filter_var($ip, FILTER_VALIDATE_IP, $flags);
+
+		if ( ! $allow_private)
```
rjd22: Please keep this `if ($allow_private === FALSE)` for backwards compatibility. What if someone does:

``` php
Valid::ip('ip', null);
```

rjd22:
@enov I'm missing tests if possible. :)

classes/Kohana/Valid.php
enov: Hmmm... I understand.

enov:
Thanks for reviewing @rjd22. Kindly note that the tests are failing already, as PHP v7.0.10+ AND v5.6.25+ introduced the breaking bug fixes in those minor versions.

see for example https://github.com/kohana/core/pull/698

Thanks again.

enov:
Please do not merge this, as it might be solved in the next minor versions of PHP https://github.com/php/php-src/pull/2113

enov:
Fixed in PHP.

PR #707, opened 2018-02-20
Mastercard new numbers

antrouss:
Added Expansion to 2-series Bank Identification Numbers (BINs) for Mastercard. This applies to credit card validation.


farazmunj:
unfortunately this project is not supported any more, please switch to koseven/koseven

